PR #83, opened 2020-09-30
Update README.md

shubhhhhh:


lerno:
I don't know if that it can be said to be easier. It's similar to C after all and I would not make such a claim.

PR #107, opened 2021-03-06
Fix build on x86_64-linux-musl

zweimach:
Hi, I fixed a typo caught by the compiler and also fixed some errors when I tried to compile on x86_64-linux-musl.

As to why I changed `sys/errno.h` to `errno.h` on musl ([source](https://git.musl-libc.org/cgit/musl/tree/include/sys/errno.h)):
```c
#warning redirecting incorrect #include <sys/errno.h> to <errno.h>
#include <errno.h>
```

And why I changed the order of `sys/stat.h` on musl ([source](https://git.musl-libc.org/cgit/musl/tree/arch/x86_64/bits/stat.h)):
```c
struct stat {
	/* .... omitted */
	long __unused[3];
	/* omitted ... */
};
```
which conflicts with:
https://github.com/c3lang/c3c/blob/1d50beb3303715a6e2df2478f58919e0716f2e8b/src/utils/common.h#L19-L21

### DIFF
--- a/src/compiler/bigint.c
+++ b/src/compiler/bigint.c
@@ -193,7 +193,7 @@ static void to_twos_complement(BigInt *dest, const BigInt *source, size_t bit_co
 		if (dest->digit == 0) dest->digit_count = 0;
 		return;
 	}
-	dest->digits = malloc_arena(dest->digit_count * sizeof(u_int64_t));
+	dest->digits = malloc_arena(dest->digit_count * sizeof(uint64_t));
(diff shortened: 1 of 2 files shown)


zweimach:
I think this might not be the most elegant solution (except for the typo). Maybe the maintainers could close this PR and address this issue with the best solution.

lerno:
@zweimach I picked __unused as it is defined on Clang as an alias for  __attribute__((unused)). I'd work on this if it wasn't for me in the middle of rewriting casts and implicit conversions. So I'll

Issue #112, opened 2021-04-05
Exceeding bit width in shifts should not be UB


lerno:
Exceeding bit width should not be UB, which means that in LLVM we should freeze the result.

lerno:
This is fixed now.

PR #114, opened 2021-04-07
UB / trapping updates.

lerno:


lerno:
This fixes issue #112

PR #118, opened 2021-04-14
Don't use -Werror

data-man:
gcc version 10.2.0
>file_utils.c:131:2: error: ‘strncpy’ specified bound depends on the length of the source argument [-Werror=stringop-overflow=]
  131 |  strncpy(path, full_path, path_len + 1);

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -143,7 +143,7 @@ add_executable(c3c
         src/compiler/llvm_codegen_c_abi_riscv.c
         src/compiler/llvm_codegen_c_abi_wasm.c)
 
-target_compile_options(c3c PRIVATE -Wsign-compare -Wimplicit-int -Werror -Wall -Wno-unknown-pragmas -Wextra -Wno-unused-function -Wno-unused-variable -Wno-unused-parameter)
+target_compile_options(c3c PRIVATE -Wsign-compare -Wimplicit-int -Wall -Wno-unknown-pragmas -Wextra -Wno-unused-function -Wno-unused-variable -Wno-unused-parameter)
 
 target_include_directories(c3c PRIVATE
         "${CMAKE_SOURCE_DIR}/src/")


lerno:
Thank you, but I've fixed the usage instead. Can you verify that it works on 10.2.0?

data-man:
Thanks!

> Can you verify that it works on 10.2.0?

`c3c/src/compiler/llvm_codegen.c:13:14: error: variable ‘severerity_name’ set but not used [-Werror=unused-but-set-variable]
   13 |  const char *severerity_name = "unknown";
      |              ^~~~~~~~~~~~~~~
cc1: all warnings being treated as errors`

lerno:
@data-man I think that one is fixed now as well. If you have a list of warnings, paste it here and I'll fix them all.

data-man:
> I think that one is fixed now as well.

No. The same error for 7f3963cbc5b2f25b108865dc37630234f0837037 commit.

data-man:
> If you have a list of warnings, paste it here and I'll fix them all.

> c3c/src/compiler/llvm_codegen_expr.c:1257:15: warning: ‘len’ may be used uninitialized in this function [-Wmaybe-uninitialized]

> c3c/src/compiler/llvm_codegen_expr.c:1266:11: warning: ‘parent_load_value’ may be used uninitialized in this function [-Wmaybe-uninitialized]

> c3c/src/compiler/llvm_codegen_expr.c:1851:2: warning: ‘val’ may be used uninitialized in this function [-Wmaybe-uninitialized]

data-man:
@lerno BTW, can you enable Discussions for this repo (or for c3docs)?
For non-Discord users? :)

lerno:
@data-man Those errors all seem like hideously poor flow analysis by GCC, or GCC does not pick up __attribute__((noreturn)) for some weird reason. I'll have to try to see if I can replicate this.

I'm not sure about the discussions, but there is a forum here: https://c3.handmade.network/forums

data-man:
@lerno 
If you dislike my patch as is I propose to add CMake option for it.
Somethink like `-DC3_IGNORE_COMPILING_WARNINGS`.

lerno:
It's more like there should not even be these warnings and I've now tested with GCC on Ubuntu at least and I can't reproduce, even when explicitly adding those flags. What's your environment?

lerno:
This is the version of gcc I've used to test: gcc (Ubuntu 10.2.0-13ubuntu1) 10.2.0

data-man:
>  What's your environment?

Artix Linux, gcc 10.2.0-6
$ gcc --version -v

>Configured with: /build/gcc/src/gcc/configure --prefix=/usr --libdir=/usr/lib --libexecdir=/usr/lib --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=https://bugs.archlinux.org/ --enable-languages=c,c++,ada,fortran,go,lto,objc,obj-c++,d --with-isl --with-linker-hash-style=gnu --with-system-zlib --enable-__cxa_atexit --enable-cet=auto --enable-checking=release --enable-clocale=gnu --enable-default-pie --enable-default-ssp --enable-gnu-indirect-function --enable-gnu-unique-object --enable-install-libiberty --enable-linker-build-id --enable-lto --enable-multilib --enable-plugin --enable-shared --enable-threads=posix --disable-libssp --disable-libstdcxx-pch --disable-libunwind-exceptions --disable-werror gdc_include_dir=/usr/include/dlang/gdc
Thread model: posix

lerno:
I'm now at Artix Linux with exactly that environment and I can´t reproduce those errors. Do you use CMake to build? What version of cmake?

data-man:
> Do you use CMake to build? What version of cmake?

Yes. 3.20.1 version.

> I can´t reproduce those errors.

With Release build type?

```sh
mkdir -p build-rel
cd build-rel
cmake -DCMAKE_BUILD_TYPE=Release ..
```

data-man:
Yes, Debug build hasn't errors.

lerno:
@data-man Thanks, I tracked down one issue which was use disappearing in non debug builds, which was ok. But the others were simply false positives thanks to GCCs poor data flow analysis, so I removed that particular warning (maybe-uninitialized). It should now work even in release for you, correct?

data-man:
> It should now work even in release for you, correct?

Yes, thank you!

PR #121, opened 2021-04-16
minor change in checking for unclear precedence

mathis2003:
Previously the function which is now named "unclear_op_precedence()" used to have an extra not operator when an unclear precedence was found, and then there was also a not operator in the if statement checking for unclear precedence. This was harder to read and having two not operators after each other is totally redundant.


lerno:
That's a fine improvement. I will be moving this into the semantic checker later on as it's not strictly a parsing problem.

PR #134, opened 2021-05-14
Make the build more deterministic

chritchens:
Clang-11 is expected to be installed and to be the clang version used to build the compiler. Forcing the use of clang-11 avoid miscompilations in two cases:

1. in the case other clang versions are present and the clang link does not link to clang-11.
2. in the case cc does not link to a suitable compiler.

### DIFF
--- a/src/compiler/linker.c
+++ b/src/compiler/linker.c
@@ -173,7 +173,7 @@ const char *concat_string_parts(const char **args)
 void platform_linker(const char *output_file, const char **files, unsigned file_count)
 {
 	const char **parts = NULL;
-	vec_add(parts, "cc");
+	vec_add(parts, "clang-11");
 	vec_add(parts, "-lm");
 	vec_add(parts, "-o");
 	vec_add(parts, output_file);
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,9 @@ message(STATUS "Found LLVM ${LLVM_PACKAGE_VERSION}")
 message(STATUS "Using LLVMConfig.cmake in: ${LLVM_DIR}")
 message(STATUS "Libraries located in: ${LLVM_LIBRARY_DIRS}")
 
+message(STATUS, "Using C compiler in: {CMAKE_C_COMPILER}")
+message(STATUS, "Using CXX compiler in: {CMAKE_CXX_COMPILER}")
+
 include_directories(${LLVM_INCLUDE_DIRS})
 add_definitions(${LLVM_DEFINITIONS})
 
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ work on Windows. Also, parts of the code is still rough and needs to be made sol
 5. Enter the C3C directory `cd c3c`.
 6. Create a build directory `mkdir build`
 7. Change directory to the build directory `cd build`
-8. Set up CMake build for debug: `cmake -DLLVM_DIR=/usr/lib/llvm-11/cmake -DCMAKE_BUILD_TYPE=Debug ..`
+8. Set up CMake build for debug: `cmake -DLLVM_DIR=/usr/lib/llvm-11/cmake -DCMAKE_C_COMPILER=clang-11 -DCMAKE_CXX_COMPILER=clang++-11 -DCMAKE_BUILD_TYPE=Debug ..`
 9. Build: `cmake --build .`
 
 You should now have a `c3c` executable.


lerno:
Ideally it should work to use any compiler here. I'm typically using GCC for Linux. What issues do you see @chritchens ?

chritchens:
@lerno, cc links to gcc on my laptop, but it didn't compile with it. It did with clang (from version 11 on). And cc links generally to gcc, but it's not the case on every system, so it would be a stricter option.

gcc version: gcc (Debian 10.2.1-6) 10.2.1 20210110

The build command (switching to Debug doesn't change the behavior): `mkdir build && cd build && cmake -DLLVM_DIR=/usr/lib/llvm-11/cmake -DCMAKE_C_COMPILER=/usr/bin/clang-11 -DCMAKE_CXX_COMPILER=/usr/bin/clang++-11 -DCMAKE_BUILD_TYPE=Release .. && cmake --build .`

The test command (using -fpic or -fpie doesn't change the behavior): `.c3c compile ../resources/example/hash`

The error log (up to commit 2f8f620f9fa80f4ba5a671ab755b97af729b8081):
```
/usr/bin/ld: std.math.o: in function `std::math.log10':
std::math:(.text+0x1): undefined reference to `log10'
collect2: error: ld returned 1 exit status
Failed to link executable 'foo.out' using command 'cc -lm -o foo.out hash.o std.runtime.o std.builtin.o std.io.o std.mem.o std.array.o std.math.o'.
```

For compiling "hash.c3", the problem was fixed with commit 7ee015c84f12fee15ba12212c04817e2bab856d3.

I updated the branch and using -fpie or -fpic fixed it. The problem  was with the log10 function. There are still problems compiling `../resources/examples/gameoflife.c3` or `../test/test_suite/examples/gameoflife.c3`.

Error log:
```
/usr/bin/ld: game_of_life.o: in function `game_of_life.show':
game_of_life:(.text+0x84): undefined reference to `__stdoutp'
/usr/bin/ld: game_of_life.o: in function `main':
game_of_life:(.text+0x343): undefined reference to `__stdoutp'
/usr/bin/ld: game_of_life:(.text+0x443): undefined reference to `__stdoutp'
collect2: error: ld returned 1 exit status
Failed to link executable 'foo.out' using command 'cc -lm -o foo.out game_of_life.o std.runtime.o std.builtin.o std.io.o std.mem.o std.array.o std.math.o'.
```

In any case, I'd personally prefer a stricter option with the platform linker, so that the build is more deterministic. And I think there is a need for wider CI testing and a channel outside github for questions and discussing issues and proposals. And I'm going to delete the CI file changes (that's long-term stuff paranoias or obsessive-compulsive behavior). Cool language btw.

src/compiler/linker.c
```diff
@@ -173,7 +173,7 @@ const char *concat_string_parts(const char **args)
 void platform_linker(const char *output_file, const char **files, unsigned file_count)
 {
 	const char **parts = NULL;
-	vec_add(parts, "cc");
+	vec_add(parts, "clang-11");
```
data-man, line 176: IMO, it's incorrect.
User can to have clang-12, clang-13, etc.

chritchens: Yes, but the compiler targets llvm-11 and the user is asked to install clang-11. A user may not even had clang before installing it. It works with `clang-12` and `clang-13`, but it's a less deterministic option. Will it work with `clang-14` or on? Who can say.

data-man: > Who can say.
I can. :)
I tested c3c with clang-12 and clang-git.

lerno:
The idea here was to make it independent of LLVM and Clang. The .o files should be possible to use with a project that is building using GCC for example. LLVM-11 should be the lowest version supported. In CI that's what's available, but I develop with LLVM-12 right now.

In regards to gameoflife.c3 that was just using a quick fix to get flush etc. It works on MacOS since those external variables exists there, but on Linux there is another set. This should all be encapsulated in the std::io module, but until recently I did not have a way to export an aliased variable. The correct constants to determine the arch at compile time is likewise not added yet, so it can't quite be platform correct yet. What one should do is probably to add a command line switch to set the compiler.

chritchens:
@lerno, ok, got it. One could switch compiler also with the CC environment variable. I don't know if it can be useful, but maybe using cosmopolitan could be of help (https://github.com/jart/cosmopolitan). The pros I see is greater portability, the cons is one more dependency, and a very specific one.

lerno:
I'm actually looking at musl, which is basically doing the same thing, so as a long term goal c3c will be cross platform, including having linker built in.

data-man:
Personally, I'm waiting for asm and prefer libc-free stdlib.

lerno:
In terms of asm, I would prefer *not* to just wrap it in strings, but to actually lex it and then reassemble.

So let's say this is passed to LLVM:
`i32 (i32) asm "bswap $0", "=r,r"`

Then maybe that one can look like:
```c
int y = asm(x) {
   bswap $0 @(=r, r)
}
```
That is some minimal lexing that then is reassembled into a string for LLVM.

I am open to suggestions.

data-man:
> I am open to suggestions.

Where can we discuss this? As I proposed earlier, GitHub's Discussions is very convenient.

lerno:
@data-man I've opened an issue for it here: https://github.com/c3lang/c3c/issues/142

lerno:
Also, I've enabled discussions now. I finally found the setting for it. But the issue is probably better to keep things on subject.

src/compiler/linker.c
gdm85: In my opinion you are both right: if users wants more determinism, they should stick to a known supported version. If they want to try things out, they should be able to do so :)

lerno:
I'm going to close this issue now. The ambition is to make everything compile under both GCC and Clang, so error reports on those are fine and should be treated as bug reports. You're bringing up valuable issues @chritchens, so let's just track those as errors instead.

PR #136, opened 2021-05-15
Fixed potential bug

jasmcaus:
I haven't had much time to read through the entire source code, but I think there may be a possible bug here:  `lexer->current[-1`] points to illegal memory. 

Or perhaps yet, we could return `*(lexer->current--)`, but this then defeats the purpose of 'looking' at the previous character.

### DIFF
--- a/src/compiler/lexer.c
+++ b/src/compiler/lexer.c
@@ -31,7 +31,7 @@ static inline char peek(Lexer *lexer)
 // Look at the prev character in the buffer.
 static inline char prev(Lexer *lexer)
 {
-	return lexer->current[-1];
+	return lexer->current[lexer->index--];
 }
 
 // Backtrack the buffer read one step.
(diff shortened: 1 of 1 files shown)


lerno:
I think this comes from a misconception of what the code does. It's not super obvious because of how it went from lexing-on-demand to lexing-up-front. So using lexer_advance actually moves the current token forward in the already lexed stream, whereas peek/prev are used duing lexing. So consequently "lexer_index" here is something completely different.

jasmcaus:
Right, makes sense - but don't you think `lexer->current[-1]` and `lexer->current[1]` are incorrect for prev() and peek_next() respectively? I mean, sure you may have used lexer_index in a different way, but logically those two lines don't make sense to me.

lerno:
This is how they work.
prev => peek at previous char (without changing the current pointer)
peek => return current char (without changing the current pointer)
peek_next => return next char (without changing the current pointer)
next => return the current char and step forward

lerno:
lexer_index should more properly be moved into Context though.

jasmcaus:
Okay yes, my bad - I was operating under the assumption that `lexer->current` is the immutable char buffer for a single lexer state. Your code makes sense :)

PR #139, opened 2021-05-17
Fix casts in math module

data-man:


lerno:
Thank you, I made a mistake checking those in. I'll be working on an alternative cast syntax that looks like that though. Feel free to have opinions about that. I've considered many types of cast syntax, many can be found here: https://dev.to/lerno/a-zoo-of-casts-4bob

PR #154, opened 2021-05-28
Fix broken hyperlink

SarahIsWeird:
There was a space between the brackets, and as a result, the hyperlink didn't display properly.


lerno:
Thank you

PR #169, opened 2021-06-13
Add support of builds for Ubuntu 18/20 through Docker

gdm85:
I was trying to build `c3c` on Ubuntu 18 (LTS) and found a few issues; in order to help others build it with ease I thought it would be a good idea to dockerize the build.

This PR allows to build `c3c` for either Ubuntu 18 or 20 through a Docker container, making the process very easy.

Unfortunately `c3c` binaries are not static so one still needs some library on host system, but the whole builds becomes more reproducible.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,4 @@ dkms.conf
 /resources/lex.yy.c
 /resources/y.tab.c
 /resources/y.tab.h
+/bin/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.13)
+cmake_minimum_required(VERSION 3.10)
 project(c3c)
 
 SET(CMAKE_FIND_PACKAGE_SORT_ORDER NATURAL)
--- a/README.md
+++ b/README.md
@@ -130,9 +130,7 @@ work on Windows. Also, parts of the code is still rough and needs to be made sol
 - Are you a Windows dev? Please help make the compiler work on Windows!
 - Install instructions for other Linux and Unix variants are appreciated.
 
-#### Installing on Ubuntu
-
-(This installation has been tested on 20.10)
+#### Installing on Ubuntu 20.10
 
 1. Make sure you have a C compiler that handles C11 and a C++ compiler, such as GCC or Clang. Git also needs to be installed.
 2. Install CMake: `sudo apt install cmake`
@@ -147,6 +145,23 @@ work on Windows. Also, parts of the code is still rough and needs to be made sol
 You should now have a `c3c` executable.
 
 You can try it out by running some sample code: `./c3c compile ../resources/examples/hash.c3`
+
+#### Building via Docker
+
+You can build `c3c` using either an Ubuntu 18.04 or 20.04 container:
+
+```
+./build-with-docker.sh 18
+```
+
+Replace `18` with `20` to build through Ubuntu 20.04.
+
+For a release build specify:
+```
+./build-with-docker.sh 20 Release
+```
+
+A `c3c` executable will be found under `bin/`.
 
 #### Installing on OS X using Homebrew
 
--- a/build-with-docker.sh
+++ b/build-with-docker.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+## build-with-docker.sh
+## @author gdm85
+##
+## Script to build c3c for either Ubuntu 18 or Ubuntu 20.
+##
+#
+
+if [ $# -ne 1 -a $# -ne 2 ]; then
+	echo "Usage: build-with-docker.sh (18|20) [Debug|Release]" 1>&2
+	exit 1
+fi
+
+set -e
+
+if [ -z "$2" ]; then
+	CMAKE_BUILD_TYPE=Debug
+else
+	CMAKE_BUILD_TYPE="$2"
+fi
+
+if [ "$1" = 18 ]; then
+	UBUNTU_VERSION="18.04"
+	DEPS="llvm-10-dev liblld-10-dev libclang-10-dev"
+elif [ "$1" = 20 ]; then
+	UBUNTU_VERSION="20.04"
+	DEPS="llvm-11-dev liblld-11-dev clang-11 libllvm11 llvm-11-runtime"
+else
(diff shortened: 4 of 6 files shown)


lerno:
1. This looks very nice, but in terms of LLVM we're going to need changes that are in the LLVM versions after 12, so the removal of that old line for LLVM 10 support was deliberate. Can we make 20.18 use LLVM 12 instead?

2. As far as c3c dependencies are concerned it only uses libc. Any libc++ dependency should be due to LLVM and LLVM is statically linked into the c3c binary. The fewer dependencies the better, so if you find something to improve, let me know.

gdm85:
> This looks very nice, but in terms of LLVM we're going to need changes that are in the LLVM versions after 12, so the removal of that old line for LLVM 10 support was deliberate.

I did not notice any old line removed :thinking: 

> Can we make 20.18 use LLVM 12 instead?

Sure; then the GitHub action should also do that.

> 2\. As far as c3c dependencies are concerned it only uses libc. Any libc++ dependency should be due to LLVM and LLVM is statically linked into the c3c binary. The fewer dependencies the better, so if you find something to improve, let me know.

Not sure how to improve it (not an LLVM expert here), but these are the dependencies:
```
$ ldd bin/c3c 
	linux-vdso.so.1 (0x00007ffe627ff000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f8047a69000)
	libtinfo.so.5 => /lib/x86_64-linux-gnu/libtinfo.so.5 (0x00007f804783f000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f80474a1000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f8047282000)
	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f8046ef9000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f8046ce1000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f80468f0000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f804c2c6000)
```
`libtinfo` and `libz` both require some package installed.

lerno:
Libtinfo and libz are both LLVM dependencies the compiler don't use them. I wonder if LLVM has made some progress here. Is this for both LLVM 12 and LLVM 10?

The reason the GitHub action doesn't do that is that LLVM-10 was originally all I could pull in easily. Then LLVM-11 was available, which is why that one is pulled right now. The preferred solution would actually to build LLVM from scratch, but in order to make that fast one needs to make the LLVM compilation cached. I know some projects that have already done that, but I haven't had time.

But I was just wondering if at least LLVM 11 was available for 20.18 so we didn't need to have that `#if`

gdm85:
> But I was just wondering if at least LLVM 11 was available for 20.18 so we didn't need to have that `#if`

There is a misunderstanding here: that `#if` is there to allow building on Ubuntu 18 LTS, which supports LLVM max version 10.

> Is this for both LLVM 12 and LLVM 10?

The version built with Ubuntu 20 has these library deps:
```
	linux-vdso.so.1 (0x00007ffc06de2000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f591c368000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f591c164000)
	libtinfo.so.6 => not found
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f591bdc6000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f591bba7000)
	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f591b81e000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f591b606000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f591b215000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f5921134000)
```

Ubuntu 20.04 (LTS) has maximum LLVM 11, not 12; there is no Ubuntu 20.18, as far as I know.

The only LTS versions are 18.04 and 20.04 and they respectively support LLVM 10 and LLVM 11.

lerno:
No I mean that it will probably not be sustainable to support LLVM 10. In this case it's just this small change, but as the language supports increasingly newer versions of LLVM, it will be unsustainable. So the question is whether it's instead possible to let the docker images build LLVM from source. It was more of a question.

In terms of linked libraries libtinfo and libz are LLVM dependencies as far as I know.

But I'll add this pull request and we could perhaps have a look at the LLVM version issue later on.

lerno:
Thank you for the contribution!

gdm85:
> No I mean that it will probably not be sustainable to support LLVM 10. In this case it's just this small change, but as the language supports increasingly newer versions of LLVM, it will be unsustainable.

Ok, my intent was to support LTS versions of Ubuntu because that is what one uses for stability.

> So the question is whether it's instead possible to let the docker images build LLVM from source. It was more of a question.

I understand; yes, it should be possible, perhaps LLVM has official images to facilitate this.

> In terms of linked libraries libtinfo and libz are LLVM dependencies as far as I know.

It looks like so; it is a bit puzzling to include these even for a simple hello world using C's `printf`.

lerno:
> I understand; yes, it should be possible, perhaps LLVM has official images to facilitate this.

I am not sure, I haven't investigated this in depth.

> It looks like so; it is a bit puzzling to include these even for a simple hello world using C's printf.

Wait, you mean for c3c right? Not a binary generated by c3c?

gdm85:
> > It looks like so; it is a bit puzzling to include these even for a simple hello world using C's printf.
> 
> Wait, you mean for c3c right? Not a binary generated by c3c?

My bad, the produced binaries do not have such dependency. The `c3c` binary on Ubuntu 20 has that extra dependency `libdl.so.2` which is not present on Ubuntu 18 (this is due to the different LLVM versions).

lerno:
Ah, yes unfortunately the LLVM dependencies is out of our hands as long as LLVM is the backend.

PR #177, opened 2021-06-27
Make testproject buildable & misc.

gdm85:
This PR does 3 things, but each change is in its own commit:

1. make `testproject` buildable again and build it as part of the CI
2. add more instructions to the example in `README.md` and change the default from `foo.out` to `a.out` (not sure if you are emotionally attached to `foo.out` there)
3. tag the Docker images with `:18` and `:20` so that caching works better, plus copy the `lib` folder after build with Docker is done (it is needed anyways)


resources/testproject/project.toml
```diff
@@ -5,7 +5,7 @@ authors = ["John Doe <john.doe@example.com>"]
 langrev = "1"
 warnings = ["no-unused"]
 # sources compiled
-sources = ["./**"]
+sources = ["hello_world.c3", "bar.c3", "zab.c3", "foo.c3"]
```
lerno, line 8: Is this necessary? ./** should compile all sources, or there is something wrong. I don't want to make people believe that they need to point out the sources.

README.md
```diff
@@ -27,6 +28,15 @@ func void main()
 }
 ```
 
+Make sure you have the standard libraries at either `../lib/std/` or `/lib/std/`.
+
+Then run
+```sh
+c3c compile main.c3 -fPIC
```
lerno, line 35: Better if we change so that Linux binaries are built by default with -fPIC. What do you think?

README.md
```diff
@@ -27,6 +28,15 @@ func void main()
...
+```
+
+The generated binary will be called `a.out`.
```
lerno, line 38: If we want a.out by default we can do so. Is there a reason why we don't want to differentiate from GCC / Clang?

lerno:
I have changed the linking somewhat. Can you see if it works without the -fPIC flag now?

data-man:
> Can you see if it works without the -fPIC flag now?

`$ c3c compile helloworld.c3`

```
/usr/bin/ld: helloworld.o: relocation R_X86_64_32 against `.rodata' can not be used when making a PIE object; recompile with -fPIE
collect2: error: ld returned 1 exit status
Failed to link executable 'foo.out' using command 'cc -lm -fno-PIE -fno-pie -o foo.out helloworld.o std.env.o std.cinterop.o std.runtime.o std.builtin.o std.io.o std.mem.o std.array.o std.math.o'.
```

`$ c3c -fPIC compile helloworld.c3 helloworld.c3`
```
/usr/bin/ld: helloworld.o: relocation R_X86_64_32 against `.rodata' can not be used when making a PIE object; recompile with -fPIE
collect2: error: ld returned 1 exit status
Failed to link executable 'foo.out' using command 'cc -lm -fno-PIE -fno-pie -o foo.out helloworld.o std.env.o std.cinterop.o std.runtime.o std.builtin.o std.io.o std.mem.o std.array.o std.math.o'.
```

resources/testproject/project.toml
gdm85: It does not work here on Linux, I have bash 4.4

README.md
gdm85: Sure

README.md
gdm85: `foo` is something used in examples, so I think it is not a great choice for the actual default output binary name. You could go literally with anything else, but it feels a bit dissonant to read in the example that the actual binary output is `foo.a` (readers might think that it is so because the filename was `foo.c3`).

gdm85:
> Can you see if it works without the -fPIC flag now?

Yes, it is broken here as well as @data-man reported.

README.md
gdm85: Since it is not yet working, you could merge it as-is and then adjust it once the defaults change.

lerno:
I made a new change. Please retry without PIC?

resources/testproject/project.toml
lerno: It's not bash. See the code file_add_wildcard_files

build-with-docker.sh
```diff
@@ -30,12 +30,12 @@ else
 	exit 2
 fi
 
-cd docker && docker build -t c3c-builder --build-arg UID=$(id -u) --build-arg GID=$(id -g) \
+cd docker && docker build -t c3c-builder:$UBUNTU_VERSION --build-arg UID=$(id -u) --build-arg GID=$(id -g) \
 	--build-arg DEPS="$DEPS" --build-arg UBUNTU_VERSION="$UBUNTU_VERSION" .
 cd ..
 
-rm -rf build bin
+rm -rf build bin lib
```
lerno, line 37: The problem here is actually in the build script. Using "install" it should copy "lib" to a good location. Which it doesn't.

resources/testproject/project.toml
gdm85: It errors with `Failed to resolve ./**`; may I suggest to add this project's compilation to the GitHub workflow?

gdm85:
> I made a new change. Please retry without PIC?

Same; you can test it with:
```sh
./build-with-docker.sh 18
docker run -ti --rm -v "$PWD":/home/c3c/source -w /home/c3c/source c3c-builder:18.04 bash -c \
        "cd resources/testproject/ && ../../bin/c3c build"

build-with-docker.sh
gdm85: You mean `CMakeLists.txt`'s `install(TARGETS c3c DESTINATION bin)`? This docker build script is doing everything inside the container, so it aims at keeping a local built copy outside of `build/` (which needs to be removed before building with either Ubuntu 18 or 20 because it would conflict otherwise).

resources/testproject/project.toml
lerno: I'd be happy to accept a pull request for that. In addition, the tests should be run and checked. That's also on my todo. As for the file includes, I'm working on a fix right now.

build-with-docker.sh
lerno: Ah, still, the install is probably not doing the right thing.

resources/testproject/project.toml
lerno: ./** is fixed now, but any improvements to the CI is greatly appreciated.

lerno:
I think the PIC/PIE issue is resolved now.

lerno:
@gdm85 could I get a pull req with only the changes that aren't addressed?

data-man:
> I think the PIC/PIE issue is resolved now.

Works for me, thank you! :rocket:

README.md
gdm85: `-fPIC` removed.

gdm85:
> @gdm85 could I get a pull req with only the changes that aren't addressed?

Can you please be more specific? :thinking: Which commits shall I remove from this PR and move to a new one?

resources/testproject/project.toml
gdm85: How do you run the tests?
Could not figure it out.

lerno:
Thank you!

PR #207, opened 2021-07-11
Windows compatibility

kvk1920:
Add MSYS2 support

### DIFF
--- a/src/compiler/linker.c
+++ b/src/compiler/linker.c
@@ -1,5 +1,9 @@
 #include "compiler_internal.h"
 
+#if PLATFORM_WINDOWS
+#include <llvm/Config/llvm-config.h>
+#endif
+
 extern bool llvm_link_elf(const char **args, int arg_count, const char** error_string);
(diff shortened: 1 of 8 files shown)


src/compiler/linker.c
```diff
@@ -14,6 +18,60 @@ static void add_files(const char ***args, const char **files_to_link, unsigned f
 	}
 }
 
+static const char *join_strings(const char **args, unsigned count)
```
lerno, line 21: If you want to do a strcat on the arena allocator, then you can use `strcat_arena`

src/compiler/linker.c
```diff
@@ -1,5 +1,9 @@
 #include "compiler_internal.h"
 
+#if PLATFORM_WINDOWS
+#include <llvm/Config/llvm-config.h>
```
lerno, line 4: Remove the ifdef here, or it won't compile on other platforms.

kvk1920: fixed

lerno (approved):

PR #223, opened 2021-07-13
Windows compatibility

kvk1920:
try to use winpthreads under MSYS2

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -27,8 +27,8 @@ jobs:
       - name: CMake
         run: |
           mkdir build && cd build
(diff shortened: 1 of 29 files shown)


Ccode-lang:
Try turning the slashes around

Ccode-lang:
D:\a\_temp\145c3ef1-bbea-443b-bed7-c3b052f7caec: line 2: D:a_tempmsysclang64bincmake.exe: command not found




It looks like the slashes are the wrong way.

Ccode-lang:
It is removing the slash because that is what works on Linux.

PR #233, opened 2021-07-15
Add inf and nan.

lerno:


lerno:
This partly addresses #233

PR #234, opened 2021-07-17
CI testing

kvk1920:
establish automatic test running, add MSYS2 MinGW64 environment into CI, fix some issues with path handling on Windows


.github/workflows/main.yml
```diff
@@ -2,11 +2,44 @@ name: CI
 
 on:
   push:
-    branches: [ master, dev, windows_compatibility ]
+    branches: [ master, dev, ci_testing ]
```
data-man, line 5: This repository hasn't these branches.

lerno: I think it is fine, this is just adding some branches where you can get CI automatically.

Issue #244, opened 2021-07-23
Mismatch between build instructions and code


lerno:
From the error report:
- c3c -new doesn't seem to work (i assume this just got renamed to c3c init, since that did work. no biggie)
- when I try to do c3c run it just says Can't open the directory 'hello_world'. Please check the paths. No such file or directory
- If I make another directory called hello_world and put main.c3 in that, I still just get Failed to resolve main.c3.
Is there some step I'm missing here?

C34A:
I managed to get it working by editing  `file_add_wildcard_files() `in `file_utils.c` to maintain a full path when it calls it self recursively, like so:
```c
void file_add_wildcard_files(const char ***files, const char *path, bool recursive)
{
	DIR *dir = opendir(path);
	if (!dir)
	{
		error_exit("Can't open the directory '%s'. Please check the paths. %s", path, strerror(errno));
	}
	struct dirent *ent;
	while ((ent = readdir(dir)))
	{
		size_t namelen = strlen(ent->d_name);
		if (namelen < 4) continue;

		// Doesn't end with .c3
		if (strncmp(&ent->d_name[namelen - 3], ".c3", 3) != 0)
		{
			bool is_directory;
#if PLATFORM_WINDOWS
			struct stat st;
			is_directory = stat(ent->d_name, &st) == 0 && S_ISDIR(st.st_mode);
#else
			is_directory = ent->d_type == DT_DIR;  // is it POSIX-compliant? As
#endif
			if (is_directory && ent->d_name[0] != '.' && recursive)
			{
				char *path_dup = strndup(ent->d_name, namelen); // old (renamed)
				/**
				 * new: this part allocates a longer string to fit path plus this directory's
				 * name, and concatenates them before recursing.
				 */
				char *new_path = malloc_arena(namelen + strlen(path) + 1);
				strcpy(new_path, path);
				if(path[strlen(path) - 1] != '/')
				{
					// make sure directories are separated by slashes
					strcat(new_path, "/");
				}
				strcat(new_path, path_dup);
				file_add_wildcard_files(files, new_path, recursive);
				free(path_dup);
			}
			continue;
		}

		/**
		 * new: This part is also modified, to concatenate the filename to path before calling
		 * vec_add.
		 */
		char *name = malloc_arena(namelen + 1 + strlen(path));
		strcpy(name, path);
		if(path[strlen(path) - 1] != '/')
		{
			strcat(name, "/");
		}
		strcat(name, ent->d_name);
[...]

PR #245, opened 2021-07-24
Recursive entry into sub paths fixed.

lerno:


lerno:
This should resolve #244 as the docs are updated as well.

PR #251, opened 2021-07-24
fixed stat() call in file_add_wildcard_files()

C34A:
This just moves the code building the full path so that it can be used in the call to `stat()`


C34A:
This seems to fix #244 for me.

lerno:
Looks ok! Thank you.

PR #268, opened 2021-08-12
Make the targets more consistent 

pitust:
The targets used to be a bit inconsistent:
 - The targets used to use either underscores or dashes, and which one they used wasn't consistent.
 - The x86_64 targets were reffered to as "x64", which is the microsoft name for the architecutre, and not the name that any CPU manufacturer uses.
 - GCC/LLVM both refer to the target as "x86_64", which makes the new name more consistent.

This patch contains three changes:
 - Two new targets are added: x86-elf and x86_64-elf. These are intended to be baremetal targets meant for things such as development of operating systems.
 - All targets are now consistent: `<arch>-<os>`
 - All "x64-<os>" targets have been renamed to "x86_64-<os>" to match intel's (and GCC/LLVM's) arch name.


lerno:
I can't accept this patch. x86_64 is rejected for several reasons, the most important reason being that it doesn't code complete well and the code conforms strictly to unified naming, since not doing so makes bugs easier to occur and harder to debug. So basically I'm rejecting this because basically it would ruin searchability for all `x86_` functions.

Eventually the compiler will need a clean way to specify not just architecture and OS, but also CPU details and sub architectures. I welcome filing an issue for that in which we can discuss how this future feature should look where everything is consistent and correctly named, but for now let's keep it as is.

pitust:
Okay, i reverted the rename from "x64" to "x86_64".

pitust:
I'm closing this, and i'll submit another PR later taking the comments into account!

PR #288, opened 2021-09-08
Create .gitattributes

BenStigsen:
Add C language syntax highlighting to C3


BenStigsen:
Tested and it does work. Of course the syntax highlighting isn't perfect since the languages differ in some ways. But it does make code reading quite a bit easier.

lerno:
Thank you!

PR #293, opened 2021-09-15
Support LLVM 14+

data-man:
`$ c3c -V`
>C3 Compiler Version:   Pre-alpha A232
Installed directory:   /home/dataman/Projects/C3/c3c/
LLVM version:          14.0.0git
LLVM default target:   x86_64-pc-linux-gnu


lerno:
They moved the old macho to a separate library?

data-man:
According to https://reviews.llvm.org/D107516 now LLVM has new Darwin backend.
This PR just my dirty solution. :)

lerno:
Thanks!

data-man:
What do you think about adding various versions of the LLVM from https://apt.llvm.org to CI?

lerno:
If someone would do that it would be great :)

data-man:
Ok.
What is the minimal version of the LLVM that must be supported by c3c?

lerno:
I think at this point LLVM 11

PR #296, opened 2021-09-27
Improve CI

data-man:
~~- added MacOS to matrix (with system's Clang and LLVM)~~
- test LLVM{11, 12, 13, 14} in Ubuntu

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -77,17 +77,23 @@ jobs:
     strategy:
       matrix:
         build_type: [Release, Debug]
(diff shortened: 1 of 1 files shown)


data-man:
I think CMakeLists.txt should be refactored first.
I'll come back to this later after experimenting.

lerno:
Looking forward to it!

data-man:
I think https://github.com/halide/Halide/blob/master/dependencies/llvm/CMakeLists.txt will very helpful.

PR #338, opened 2021-11-22
Add support for visual studio 2022 preview

C34A:
This seems to work for now, but likely will need to be updated once VS2022 comes out of preview and the library paths change.


lerno:
Thank you

PR #339, opened 2021-11-26
Smarter windows library path detection

C34A:
This switches to using a smarter system to detect windows library paths. It should be more forwards and backwards compatible.
The specific implementation is using this code:
https://gist.github.com/ActuallyaDeviloper/cd25b190743234d58079d6b08a8631e3


lerno:
Thank you, I'll add this. Now all we need is a good CI :)

PR #349, opened 2021-12-17
Update actions/checkout to v2

data-man:


data-man:
Thanks!
The time for testing several LLVM versions is getting closer. :smile:

PR #352, opened 2021-12-18
Add `-o` option to allow setting output binary name

xpyxel:
Add a simple `-o` flag to the cli to allow setting the name of the binary at runtime


lerno:
Thank you.

PR #359, opened 2021-12-21
add vscode syntax plugin

xpyxel:
This syntax highlighting is fairly basic, and could be improved, but for the most part it should support what's needed.


lerno:
Thank you!

PR #360, opened 2021-12-21
improve vscode syntax highlighter

C34A:
This adds better support for {| |} blocks, macros, and adds a `private` keyword


lerno:
Thank you.

PR #367, opened 2022-01-10
Print file's full path and col number in compiler's error message

sudison:
It will be handy for IDE to parse the compiler's error message, then figure out the path of the file, row and column number, then jump to the location directly.


lerno:
Good idea, but can you update "prev" and "warn" with the same thing while you're at it?

lerno:
Nevermind. I'll accept this and add the col for the other ones myself.

lerno:
Thank you.

PR #369, opened 2022-01-10
Use several LLVM versions in CI

data-man:


lerno:
Thanks. I'll accept this and then do the required updates to make it work.

lerno:
It looks like the new script is now having issues on linux @data-man

lerno:
Nevermind, it seems to have been some issue with github CI

PR #370, opened 2022-01-10
Disable mimalloc by the option

data-man:


lerno:
Do you see any issues running LLVM 13 with mimalloc on debug?

data-man:
I'm using LLVM 14 and mimalloc-dev. No issues. :)
Is there an issue with a specific file?

lerno:
Usually it breaks on `LLVMGetTargetMachineTriple(machine)`

data-man:
Can you try this:
```sh
!/bin/sh
mkdir -p build-rel
cd build-rel
cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang -DC3_MIMALLOC_TAG=dev -DC3_USE_MIMALLOC=ON ../ "$@"
```

data-man:
And yes, what OS are you using?

lerno:
Mac, I think a good idea would be to disable on Mac and Win32 then? I have checked and it fails to intercept calls to asprintf. Presumably something similar that happens to LLVM.

data-man:
Mac should be in CI! :)

lerno:
I would be grateful for contributions enabling mac CI (and MSVC!)

Issue #384, opened 2022-01-29
Use LLVM Polly for better optimizations


data-man:
LLVM `opt` has additional [Polly](https://polly.llvm.org) options:
>Polly is a high-level loop and data-locality optimizer and optimization infrastructure for LLVM. It uses an abstract mathematical representation based on integer polyhedra to analyze and optimize the memory access pattern of a program. We currently perform classical loop transformations, especially tiling and loop fusion to improve data-locality. Polly can also exploit OpenMP level parallelism, expose SIMDization opportunities. Work has also be done in the area of automatic GPU code generation.

```
Polly Options:
Configure the polly loop optimizer

  --polly                                                               - Enable the polly optimizer (with -O1, -O2 or -O3)
  --polly-2nd-level-tiling                                              - Enable a 2nd level loop of loop tiling
  --polly-ast-print-accesses                                            - Print memory access functions
  --polly-context=<isl parameter set>                                   - Provide additional constraints on the context parameters
  --polly-dce-precise-steps=<int>                                       - The number of precise steps between two approximating iterations. (A value of -1 schedules another approximation stage before the actual dead code elimination.
  --polly-delicm-max-ops=<int>                                          - Maximum number of isl operations to invest for lifetime analysis; 0=no limit
  --polly-detect-full-functions                                         - Allow the detection of full functions
  --polly-dump-after                                                    - Dump module after Polly transformations into a file suffixed with "-after"
  --polly-dump-after-file=<string>                                      - Dump module after Polly transformations to the given file
  --polly-dump-before                                                   - Dump module before Polly transformations into a file suffixed with "-before"
  --polly-dump-before-file=<string>                                     - Dump module before Polly transformations to the given file
  --polly-enable-simplify                                               - Simplify SCoP after optimizations
  --polly-ignore-func=<string>                                          - Ignore functions that match a regex. Multiple regexes can be comma separated. Scop detection will ignore all functions that match ANY of the regexes provided.
  --polly-isl-arg=<argument>                                            - Option passed to ISL
  --polly-on-isl-error-abort                                            - Abort if an isl error is encountered
  --polly-only-func=<string>                                            - Only run on functions that match a regex. Multiple regexes can be comma separated. Scop detection will run on all functions that match ANY of the regexes provided.
  --polly-only-region=<identifier>                                      - Only run on certain regions (The provided identifier must appear in the name of the region's entry block
  --polly-only-scop-detection                                           - Only run scop detection, but no other optimizations
[...]

PR #386, opened 2022-01-31
Linking with LLVM Polly

data-man:


data-man:
It's for #384.

lerno:
So basically here, while it's interesting to pull in polly, there are no uses of it in the source code. I haven't looked, but I'm unsure if it even has a C API, so wrappers would need to be written and so on, not to mention the pain it is to get *any* extension to LLVM to work on all platforms. Just including lld has been quite a headache. It's great that you've done this work, but it has to be put on the back burner until it can be integrated, because of the maintenance issue.

data-man:
> I haven't looked, but I'm unsure if it even has a C API

`clang` can use `Polly` with `-mllvm -polly` switch, so something like this approach can be used in c3c.

lerno:
I will reject this for now. When we integrate Polly for real we can add the dependencies.

PR #411, opened 2022-03-12
Show '--test' switch in the usage help

data-man:
I'm sorry if my description is incorrect.


data-man:
@lerno are you all right?

lerno:
The --test switch is just for the test runner, so it's not intended for any public use. It *could* get modified to work with tools (the output is fairly amenable to parsing), but it's not something anyone should rely on, so I would prefer to keep it hidden for now.

lerno:
And thanks for asking by the way, everything is fine, just behind checking emails.

data-man:
I discovered this hidden switch in the tester and it's very useful!
Some tests (vectors related) pass the tester successfully, but the compiler crashes when building executables.

lerno:
I don't want to make this available in the help. However it could be useful as a plugin functionality for editors, esp the error format. But it should probably be well defined then.

lerno:
I'll close this with motivation given in the previous comment.

PR #412, opened 2022-03-16
Apply clang-format to all sources and add it to CI

data-man:


lerno:
Clearly the clang-tidy is not completely correct, because it's actually reformatting a lot of code that shouldn't be reformatted. I particularly hate this:

```c
if (foo)
   goto RETRY;
```

Which this checkin just ran with all over the place. So to accept this the clang-tidy must be modified to conform with the current coding standards.

lerno:
So what I mean is that I will accept this pull request when I can see that the tidy isn't actually rewriting the code with a different code style.

data-man:
Got it!

I'll test all supported styles (LLVM, GNU, Google, Chromium, Microsoft, Mozilla, WebKit) and experiment with the `.clang-format` file.

PR #415, opened 2022-03-18
Don't replace digits to underscore in created files

data-man:


data-man:
@lerno what if these lines were deleted completely?
https://github.com/c3lang/c3c/blob/6789fab93c288ac7262102ca791bd5aeac651173/src/compiler/context.c#L69-L76

lerno:
If they are deleted, then the generated module would not possible to use.

PR #417, opened 2022-03-20
Update build-with-docker.sh

data-man:
I don't use Docker, but I hope these are valid changes.

### DIFF
--- a/build-with-docker.sh
+++ b/build-with-docker.sh
@@ -2,12 +2,12 @@
 ## build-with-docker.sh
 ## @author gdm85
 ##
-## Script to build c3c for either Ubuntu 18 or Ubuntu 20.
+## Script to build c3c for either Ubuntu 20, 21 or 22.
 ##
 #
 
 if [ $# -ne 1 -a $# -ne 2 ]; then
-	echo "Usage: build-with-docker.sh (18|20) [Debug|Release]" 1>&2
+	echo "Usage: build-with-docker.sh (20|21|22) [Debug|Release]" 1>&2
 	exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


lerno:
Can we have some confirmation that this works? I don't have a setup to test it.

data-man:
cc @gdm85 :)

gdm85:
I don't think it's a good idea to remove Ubuntu 18, see how far in the future it is supported here: https://ubuntu.com/about/release-cycle

Cutting out an LTS release (and all the people using it) to save a couple lines here does not seem a worthy gain; I also think the `DEPS` approach was more flexible but that's less important.

I verified that this version still works but you need to remove `lib` from the `rm -rf` line (and the corresponding `mv` on the last line in the script) otherwise it will fail:
```
CMake Error at CMakeLists.txt:84 (file):
  file COPY cannot find "/home/c3c/source/lib": No such file or directory.
```

I guess at some point `lib` was versioned so this script should not touch it anymore.

data-man:
> I don't think it's a good idea to remove Ubuntu 18

According to [https://packages.ubuntu.com/search?suite=all&searchon=names&keywords=llvm](https://packages.ubuntu.com/search?suite=all&searchon=names&keywords=llvm) Bionic hasn't LLVM 12.

In any case, `build-with-docker.sh` is incorrect because the minimum required LLVM version must be >= 12 now.

data-man:
> I verified that this version still works but you need to remove lib from the rm -rf line (and the corresponding mv on the last line in the script) otherwise it will fail:

Thank you! Fixed.

gdm85:
I wasn't aware that LLVM12 has become a minimum requirement, in that case no point keeping Ubuntu18.

The last command `mv c3c ../bin/ && mv lib ../bin/"` could be compacted as `mv c3c lib ../bin/`.

lerno:
Anything else to update here? Or is it ready to add? @gdm85

gdm85:
@lerno yes it's fine, it works.

I will submit a PR to add `podman` support after you merge this one.

gdm85:
There it is: https://github.com/c3lang/c3c/pull/418

I/you can rebase it after this is merged.

lerno:
Thank you

PR #418, opened 2022-03-23
Add support for podman

gdm85:
I have been using [podman](https://podman.io/) for a few years now, I like that you can run it in rootless mode.

Nonetheless, it is close to 100% compatible with Docker so little changes are necessary (that `--userns=keep-id` for example, the `localhost/` prefix is optional).

This PR switches to `podman` when using `build-with-docker.sh` (I am aware of the ugliness of that `docker` in the script name) if it is locally available.

`docker` is going to be around with us for many years to come so this offers a graceful optional "upgrade" when `podman` is locally available.

This PR should merge cleanly after #417.

### DIFF
--- a/build-with-docker.sh
+++ b/build-with-docker.sh
@@ -13,6 +13,15 @@ fi
 
 set -e
 
+DOCKER=docker
+DOCKER_RUN=""
+IMAGE="c3c-builder"
+if type podman 2>/dev/null >/dev/null; then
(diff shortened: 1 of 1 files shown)


lerno:
Can the conflicts be resolved?

gdm85:
It is ready now.

lerno:
Thank you

PR #427, opened 2022-04-12
Fix resource leak in check_file function

data-man:
Discovered by cppcheck.
I propose use [some](https://github.com/marketplace?type=actions&query=cppcheck+) C/C++ checkers in CI.


lerno:
If you add them to the CI I'll approve it.

PR #430, opened 2022-04-13
Add more vector tests

data-man:
I don't know how to generate ll code properly. :(


lerno:
I will typically look at the new .ll as reported by the tester tool, copy paste that into the test case and then trim down and manually inspect the result.

data-man:
Aha, thanks!

lerno:
Yes, certainly. There's also the missing features to get dot syntax for some operations. And that's something for an `@operator` extension to do.

data-man:
> There's also the missing features to get dot syntax for some operations. And that's something for an @operator extension to do.

What do you think about operator overloading for any type? I have PoC! :)

lerno:
I'm on the fence. Possibly for the +-/* subset.

Issue #299, opened 2021-09-29
Built-in linker in Linux


data-man:
I'm currently using this patch in my OS.
It's bad, but it works.

```diff
diff --git a/src/compiler/compiler.c b/src/compiler/compiler.c
index 62e2cbf..9ba427b 100644
--- a/src/compiler/compiler.c
+++ b/src/compiler/compiler.c
@@ -414,14 +414,16 @@ void compiler_compile(void)
        {
                if (active_target.arch_os_target == ARCH_OS_TARGET_DEFAULT)
                {
+                       printf("Use platform linker.\n");
                        platform_linker(active_target.name, obj_files, output_file_count);
                }
                else
                {
+                       printf("Use built-in linker.\n");
                        if (!obj_format_linking_supported(platform_target.object_format) || !linker(active_target.name, obj_files,
                                                                                                    output_file_count))
                        {
-                               printf("No linking is performed due to missing linker support.");
+                               printf("No linking is performed due to missing linker support.\n");
                                active_target.run_after_compile = false;
                        }
                }
diff --git a/src/compiler/linker.c b/src/compiler/linker.c
index 4def40f..a9dd077 100644
--- a/src/compiler/linker.c
+++ b/src/compiler/linker.c
@@ -124,26 +124,26 @@ static bool link_exe(const char *output_file, const char **files_to_link, unsign
                                        if (platform_target.pie || platform_target.pic)
                                        {
                                                vec_add(args, "--eh-frame-hdr");
-                                               vec_add(args, "/usr/lib/x86_64-linux-gnu/crt1.o");
-                                               vec_add(args, "/usr/lib/gcc/x86_64-linux-gnu/10/crtbeginS.o");
+                                               vec_add(args, "/usr/lib/crt1.o");
+                                               vec_add(args, "/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/crtbeginS.o");
                                                add_files(&args, files_to_link, file_count);
-                                               vec_add(args, "/usr/lib/x86_64-linux-gnu/crti.o");
-                                               vec_add(args, "/usr/lib/gcc/x86_64-linux-gnu/10/crtendS.o");
+                                               vec_add(args, "/usr/lib/crti.o");
+                                               vec_add(args, "/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/crtendS.o");
                                                vec_add(args, "-pie");
                                        }
                                        else
                                        {
[...]

PR #433, opened 2022-04-14
Fix linking logic

data-man:


data-man:
I'm trying to fix #299.

lerno:
I'm going to try to make a better stab at this. Currently I'm working on library support, which will also affect this whole linker chain.

data-man:
Looking forward to it!

PR #440, opened 2022-05-05
Remove build-msys2-clang from CI

data-man:
It's buggy.


lerno:
The problem here is that they removed a bunch of things from the package, because – they claim – it didn't fit the DLL. I'll work on getting all win targets to use the precompiled files.

data-man:
> I'll work on getting all win targets to use the precompiled files.

IMO, msys2-mingw and msvc are more than enough. :)

lerno:
This is already done.

PR #445, opened 2022-05-09
Updated enums, macro usage / naming

lerno:


lerno:
How do you mean "bad news"?

data-man:
I meant ordinal based enums.

lerno:
The ordinal based enums are complemented by grouped constants. Let me show an example:
```c
```c
// Before
enum Foo { A, B }
enum Bar { X = 123, Y = 235 }
fn void test()
{
   Foo f = A;
   Bar b = X;
}
// After
enum Foo { A, B }
define Bar = distinct int;
module foo.bar;
const Bar X = 123;
const Bar Y = 235;
fn void test()
{
   Foo f = A;
   Bar b = bar::X;
}
```

lerno:
This basically retains the same type-safety as enums have, but they are defined as constants. This helps API stability.

data-man:
> they are defined as constants.

It's exactly bad news! :)

lerno:
Why? There is no difference between those and C style enums.

data-man:
Because

```rust
enum MyEnum : uint
{
    F1 = 1,
    F2 = 4,
    F3 = 10
}
```
is better.

lerno:
The problem with non-ordinal enums occur with associated values. There is no way to write those fast with non-ordinal enums. In addition, non-ordinal enums mess up compile time enum resolution. If you want the above and you are using enum => value (and not value => enum), you can do:

```c
enum MyEnum : uint (uint val)
{
    F1(1),
    F2(2),
    F3(10)
}
```

lerno:
To match value => enum:
```c
foreach (e : MyEnum.values) if (e.val == num) return e;
```

data-man:
```rust
enum MyEnum : uint (uint val)
{
    F1(1),
    F2(2),
    F3(10)
}
```
vs
```rust
enum MyEnum : uint (uint val)
{
    F1 = 1,
    F2 = 2,
    F3 = 10
}
```
or
```rust
enum MyEnum : uint (uint val)
{
    F1: 1,
    F2: 2,
    F3: 10
}
```

data-man:
Also I'm using built-in properties `.min` and `.max`.
Now it's not work.

lerno:
Implementing "max":
```c
int value = 0;
foreach (e : MyEnum.values) value = max(e. value);
```
I can add some macros for these, so that they can be used for any enum and you will be able to use it like `max_enum(MyEnum, value)`. Would that be helpful?

lerno:
Also, I am planning on adding a way to mark a macro/function as a dot function on all enums. That will give the possibility to use `Foo.@max(value)`

data-man:
> Would that be helpful?

Sure, but then \`.elements\` can also be implemented by macros.

data-man:
BTW, I don't like that enum members should be in upper case.
Please relax this restriction.

lerno:
Unfortunately the naming rules are there to make the grammar unambiguous, not because I want to enforce a code style.

data-man:
Ok, please relax the naming rules. :)

data-man:
`$ c3c compile true.c3`

> resulted in a name that is a reserved keyword, try using an explicit module name.

:(

lerno:
I just said that the naming rules cannot be relaxed because they are needed for the grammar.

lerno:
I'll merge this, but feel free to continue using the old version until the enum associated values work. The need for ordinals was blocking adding associated values, but associated values are needed for `@enum_max` obviously. So it's a bit of chicken-egg problem. Can you perhaps share some of the C3 code you've written? It's helpful to see how people use it in order to not make changes that break use cases. I didn't know anyone relied on min/max until now for example.

data-man:
> I didn't know anyone relied on min/max until now for example.

`std.enumset.c3`

lerno:
I forgot about that one. I've fixed it now and it works for the new enums.

lerno:
In regards to enumset, it should be very simple to increase this using arrays as backing data.

data-man:
> using arrays as backing data.

I prefer vectors. And I'm waiting for new (`$$ctpop` and family) builtins. :)

lerno:
Is it ok if I merge this and next I'll work on associated values to have enums feature complete. Any particular builtins you want me to do early let me know.

data-man:
> Any particular builtins you want me to do early let me know.

Thank you! I'll create poll(s) with my proposals.

PR #481, opened 2022-07-17
Add Binary-Heap Based Priority Queue to Standard Library

davecom:
This is an initial design for a priority queue. It builds off of the List type. The API may consider being updated for both List and PriorityQueue to use optionals.


lerno:
Thank you.

PR #483, opened 2022-07-19
Make sure isEmpty() returns when size == 0?

Andersama:
Currently the return value is likely inverted from what's expected.


lerno:
Thanks for reporting it. I updated it with the fixes to the method extensions.

Andersama:
Alright, I'll check to see if it all works.

PR #490, opened 2022-07-20
Fixes crash after using List.free

Andersama:
entries needs to be reset to 0 otherwise the previous pointer will crash when realloc is called again.

If this is intended to prevent use after free that's a neat trick. It's not exactly clear from the docs if lifetime memory management is a feature in the language since it's built on top of C. Is `.free` is called like a destructor if a struct has it?

### DIFF
--- a/lib/std/list.c3
+++ b/lib/std/list.c3
@@ -113,6 +113,7 @@ fn void List.free(List *list)
     mem::free(list.entries);
     list.capacity = 0;
     list.size = 0;
+    list.entries = (Type*)0;
 }
 
 fn void List.swap(List *list, usize i, usize j)


Andersama:
Here's the test example:
```c
module compiler_c3;
import std::array::list;

define IntArray = List<int>;

extern fn void printf(char*, ...);

fn void IntArray.someFunc(IntArray *this, int param) {
//do something
    this.push(param);
}

fn void IntArray.unchecked_reserve(IntArray *this, int capacity) {
    this.entries = mem::realloc(this.entries, capacity*(int.sizeof));
    this.capacity = capacity;
}

fn int main(char[][] argv) 
{
    IntArray stk;
    stk.unchecked_reserve(512);
    stk.someFunc(256);

    for (int i = 0; i < 256; i++) {
        stk.push(i);
    }

    for (;!stk.isEmpty();) {
        int value = stk.pop();
        printf("%i\n", value);
    }

    stk.free(); //may crash after here, program will exit w/ a non-zero value
    printf("entries: %i\n", (int)(iptr)stk.entries);
    stk.unchecked_reserve(64);

    for (int i = 0; i < 64; i++) {
        stk.push(i);
    }

    for (;!stk.isEmpty();) {
        int value = stk.pop();
        printf("%i\n", value);
    }

    return 0;
}
```

lerno:
I can't recall to be honest. The List type was written fairly early as a language test. It hasn't gotten a lot of love since that. The actual behaviour. I think this fix is probably correct.

lerno:
But any reason why you use `(Type*)0` instead of `null`?

Andersama:
No particular reason, should've figured null was part of the language, also makes it fairly clear (at least in my mind) that the whole struct is being 0'd out.

Andersama:
I was more or less just concerned there was some lifetime memory management I was unknowingly messing around with.

lerno:
Compile time reflection yes, try this:

```c
import std::io;

struct Abc
{
  int x;
}
fn void Abc.test() {}

fn void main()
{
  io::printf("%s %s\n", $defined(Abc.test), $defined(Abc.test2));
}
```

lerno:
You should see that it prints "true false"

lerno:
Shall I change to `null` and merge it?

Andersama:
Yeah that's fine. I just tried out adding this:
```
    $if ($defined(Type.free)):
        for (usize i = 0; i < list.size; i++) {
            list.entries[i].free();
        }
    $endif;
    ```
    to the free method, very much enjoying the language.

lerno:
Not yet, but will be, for the moment you can at least check the return type using `$typeof(Type.free(null))`.

lerno:
I am not 100% I understand the question, but... if you mean that ListType is the result of `define ListType = List<SomeType>` then yes. Although it wouldn't be particular meaningful since in that case you already knew the parameterized type. Otherwise it looks fine.

lerno:
BTW the list methods names may change. Currently the stdlib is inconsistent with function names using snake_case and dot methods sometimes having camelCase and sometimes snake_case. I will likely change this to snake_case for methods as well.

Andersama:
Yeah I was curious about that when looking through the libraries. I take it you'd rather have snake_case for methods to distinguish the identifiers from looking too similar to types?

lerno:
The regex for user defined types is `_*[A-Z][_A-Z0-9]*[a-z][a-zA-Z0-9_]*` So basically it starts with an upper case then has at least one lower case. Constants are all upper and everything else is `_*[a-z][a-zA-Z0-9_]*`. This way it's trivial to detect a type in the grammar.

It was an open question whether to use snake_case or camelCase for variables, functions and methods. For C it's more common to use snake_case though and in the C/C++ community there's a preference for snake_case. But when I was writing `List` the naming wasn't yet settled.

Personally I don't have any strong preference, just that code bases are consistent in their use.

As an aside there is a style popular in the C++ community which looks like: `SomeType MyFunction(int parameter)`, but that style cannot differentiate between a function and a type, and so doesn't simplify the grammar, which ruled that out.

lerno:
Oh, and I updated List with that null as I was fixing the names, so it should now be fine.

Andersama:
Well for my own compiler since I didn't want to bother with a preprocessor I used # to denote types, but I'm reconsidering since c3 seems pretty straightforward.

PR #498, opened 2022-07-22
Added a std::vector container to stdlib

raynei86:
Basically C++ `std::vector` but I took the code from C3 `std::list`


lerno:
What's the difference between this one and List?

raynei86:
I was copying and pasting code and forgot the most important part..

PR #511, opened 2022-08-07
Update CI to upload release artefacts

kstrb:
Theres a global variable LLVM_RELEASE_VERSION that should be set to the LLVM-Version that we want to cut the release from (we might be able to replace this by always using - say - the first element of the array "llvm_version"?)

For windows builds, it only uses "build-msvc" for releases.

I'm not too sure about the correct package format for macos, so it uses .zip for now.

It only packages the "c3c" binary, I'm not sure, if that is enough?

The workflow only uses official Github actions, although some of them seem to be "unmaintained" (create-release, upload-release-asset).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,6 +6,9 @@ on:
   pull_request:
     branches: [ master ]
 
+env:
+  LLVM_RELEASE_VERSION: 14
+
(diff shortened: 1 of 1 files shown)


lerno:
I'm merging this and then we might do further work on it.

PR #523, opened 2022-08-11
Use LLVM's tools in CI

data-man:


data-man:
Because lld is faster. :)

lerno:
Thanks!

PR #526, opened 2022-08-11
Improve navigation of messages

Andersama:
When using msvc or visual studio code the full path needs to be used to provide a clickable link*


lerno:
I think this will break a lot of tests.

lerno:
Added separately in this commit: https://github.com/c3lang/c3c/commit/ac7823cabb1afe8f82e4db5ce904ac341167246c

PR #527, opened 2022-08-11
Use debug libraries when doing a debug build on Windows

kstrb:
This results in some warnings related to .pdb, when combining "-g" and "--wincrt=static", which is related to still incomplete support for .pdbs.


lerno:
Thank you

PR #529, opened 2022-08-14
More context for index error messages

Andersama:
Currently the error messages only give the user a hint for the size of the array, not the index's value which may be helpful.


Andersama:
Closing because on second look I'm not sure the logic in the range section makes sense, and I'm not sure how to rewrite the output to make sense.

PR #530, opened 2022-08-14
Add batch file for installing Windows requirements and update README

kstrb:
Added a little .bat-File for Windows to automate installing the Build Tools for Visual Studio. Manually running it is not really a big win but maybe there's a way to better integrate it.

Also updated the README.


lerno:
Thank you

PR #541, opened 2022-08-27
Implement `host` level of vector instructions

data-man:


lerno:
This won't work correctly since the following are set using the type, and "native" would just give crap. The correct way would be to pull the detection from LLVM (or better yet, detect it during compilation and set the "native" to the right setting) and have the X86VectorCapability to pick the right one.

```c
unsigned align_simd_default : 16;
X86VectorCapability x86_vector_capability : 4;
bool soft_float : 1;
bool pass_int128_vector_in_mem : 1;
int native_vector_size_avx;
```

data-man:
> This won't work correctly

Works fine for me. :)
```c
...
if (len > 0) features[len - 1] = 0;
DEBUG_LOG("CPU features: %s", features);
...
```
With `--x86vec=native` option:
>
-- DEBUG: CPU features: -avx512pf,-tsxldtrk,+cx16,+sahf,-tbm,-avx512ifma,-sha,+crc32,-fma4,-vpclmulqdq,+prfchw,+bmi2,-cldemote,+fsgsbase,-avx512bf16,-amx-tile,-uintr,-gfni,+popcnt,-ptwrite,+aes,-avx512bitalg,-movdiri,-widekl,+xsaves,-avx512er,-avxvnni,-avx512fp16,-avx512vnni,-amx-bf16,-avx512vpopcntdq,-pconfig,-clwb,-avx512f,+xsavec,-clzero,-pku,+mmx,-lwp,-rdpid,-xop,+rdseed,-waitpkg,-kl,-movdir64b,-sse4a,-avx512bw,+clflushopt,+xsave,-avx512vbmi2,+64bit,-avx512vl,-serialize,-hreset,+invpcid,-avx512cd,+avx,-vaes,-amx-int8,+cx8,+fma,-rtm,+bmi,-enqcmd,+rdrnd,-mwaitx,+sse4.1,+sse4.2,+avx2,+fxsr,-wbnoinvd,+sse,+lzcnt,+pclmul,-rdpru,+f16c,+ssse3,+sgx,-prefetchwt1,+cmov,-avx512vbmi,-shstk,+movbe,-avx512vp2intersect,+xsaveopt,-avx512dq,+sse2,+adx,+sse3

Ok, what about renaming `native` to `host`?

lerno:
The value is used to determine ABI lowering, so "native" in this version will basically assume we're compiling for SSE or worse.

lerno:
So basically what's needed is quite a bit more work, where the various CPUs are identified with the correct capabilities and translated to the correct featureset.

lerno:
Jumping into this again. A deeper discussion is how the language should handle vector instructions. Ideally I'd like to see that it's easy to write code that is polymorphic over the runtime capabilities. So you essentially get multiple versions compiled and then there's a switch to the right version using cpuid or similar. There is no easy way to model this with the current vector types which are loosely based on the LLVM vectors. But as mentioned elsewhere that abstraction breaks down with SVE vectors even.

data-man:
Honestly this switch is a bit confusing. As far as I understand, optimization options do not only apply to vectors.
I think `--x86vec=` switch should be renamed to `-march=` like gcc, clang, etc.

lerno:
`--x86vec` is also there to be able to turn off floating point usage when doing kernel work if I recall correctly.

data-man:
Renamed to `host`.

lerno:
Yes, we can certainly add this functionality, but it needs a bunch of additional supporting function before it can be implemented.

PR #545, opened 2022-08-28
Add rotate_left and rotate_right macros

data-man:
Thanks for $$f(shl|shr) built-ins!
Test:
```rust
import std::io;
import std::bits;

fn void main()
{
    ushort i = 1;
    ushort irol = rotate_left(i, 1);
    ushort irol2 = rotate_left(i, -1);
    ushort iror = rotate_right(i, 1);
    ushort iror2 = rotate_right(i, -1);

    io::printf("i = %b\n", i);
    io::printf("irol = %b\n", irol);
    io::printf("irol2 = %b\n", irol2);
    io::printf("iror = %b\n", iror);
    io::printf("iror2 = %b\n", iror2);
}
```

> i = 1
irol = 10
irol2 = 1000000000000000
iror = 1000000000000000
iror2 = 10

### DIFF
--- a/lib/std/bits.c3
+++ b/lib/std/bits.c3
@@ -59,3 +59,20 @@ macro fshr(hi, lo, shift) @builtin
 {
 	return $$fshr(hi, lo, ($typeof(hi))shift);
 }
+
+/**
+ * @require types::is_intlike($typeof(i)) && types::is_intlike($typeof(amount)) `The input must be an integer or integer vector`
+ **/
+macro rotate_left(i, amount) @builtin
+{
+	return $$fshl(i, i, ($typeof(i))amount);
+}
+
+/**
+ * @require types::is_intlike($typeof(i)) && types::is_intlike($typeof(amount)) `The input must be an integer or integer vector`
+ **/
+macro rotate_right(i, amount) @builtin
+{
+	return $$fshr(i, i, ($typeof(i))amount);
+}
+


lib/std/bits.c3
```diff
@@ -59,3 +59,20 @@ macro fshr(hi, lo, shift) @builtin
 {
 	return $$fshr(hi, lo, ($typeof(hi))shift);
 }
+
+/**
+ * @require types::is_intlike($typeof(i)) && types::is_intlike($typeof(amount)) `The input must be an integer or integer vector`
+ **/
+macro rotate_left(i, amount) @builtin
+{
+	return $$fshl(i, i, ($typeof(i))amount);
```
lerno, line 68: This cast won't work on vectors

data-man: Copy-n-pasted from new f(shl|shr) macros. ;)

lerno: Yes but it has the last check required check.
```
/**
 * @require types::is_intlike($typeof(hi)) && types::is_intlike($typeof(lo)) && types::is_intlike($typeof(shift)) `The input must be an integer or integer vector`
 * @require $typeof(hi).typeid == $typeof(lo).typeid `Hi and low arguments must have the same type`
 * @require $typeof(hi).typeid == $typeof(shift).typeid `The shift value must have the same type`
 **/
```

lerno:
I'll rename them a bit standardized to `rotl` and `rotr` since they're supposed to be used without prefix.

PR #549, opened 2022-08-29
Rename `$$fabs` to `$$abs`

data-man:
...and support for vectors.


lerno:
Yeah, this is a good change.

PR #550, opened 2022-08-29
Sort some things alphabetically

data-man:
For more human-friendly navigation through code and `--list-*` commands.


lerno:
Looks good. I do the occasional sorting but sometimes it gets messed up.

data-man:
The only thing left to finish is `--list-keywords`. :)

PR #551, opened 2022-08-29
Add new built-ins

data-man:
Added:
```$$copysign, $$exp2, $$floor, $$llrint, $$llround, $$lrint, $$lround, $$nearbyint, $$rint, $$round, $$roundeven```.

For commented lines in tests I got something like this:
> LLVM ERROR: Cannot select: 0x7fd9b80231d0: f64 = llrint 0x7fd9b8023a20


lerno:
There's a reason why I haven't added all the builtins in LLVM yet, and that is because I am waiting to integrate the TildeBackend as an alternative backend for fast compiles. I don't know what TB supports yet, but what it doesn't support we'd need to provide non-builtin implementations for, and I'm not really keen on doing that. So I prefer that we only implement the most necessary ones. So if there's one you really need, then we can add that one, otherwise let's wait for it.

data-man:
> TildeBackend

I didn't know about that backend. Is this it?
https://github.com/RealNeGate/tilde-backend

> then we can add that one

Maybe disable `$$llrint, $$llround, $$lrint, $$lround` only for now?

lerno:
Yes, that's the one. Which ones do you suggest we keep?

data-man:
`$$copysign, $$exp2, $$floor, $$nearbyint, $$rint, $$round, $$roundeven`

data-man:
Might be interesting for you: https://github.com/vnmakarov/mir

lerno:
Thank you

PR #552, opened 2022-09-05
Add `$$LLVM_VERSION` define.

data-man:


src/compiler/compiler.c
```diff
@@ -8,6 +8,8 @@
 #include <unistd.h>
 #endif
 
+#include <llvm/Config/llvm-config.h> // for LLVM_VERSION_MAJOR
```
lerno, line 11: Replace this with:
```c
extern int llvm_version_major;
```

PR #563, opened 2022-09-09
More asm instructions

jb-perrier:
Added instructions : 
- MUL : mulb mulw mull mulq
- SUB : subl subw subl subq

### DIFF
--- a/src/compiler/asm_target.c
+++ b/src/compiler/asm_target.c
@@ -278,6 +278,8 @@ static void init_asm_x86(void)
 {
 	Clobbers rax_mask = clobbers_make(X86_RAX, -1);
 	Clobbers cc_flag_mask = clobbers_make(X86_CC, -1);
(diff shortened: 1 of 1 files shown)


src/compiler/asm_target.c
```diff
@@ -344,6 +345,14 @@ static void init_asm_x86(void)
 	reg_instr_clob("xorw", cc_flag_mask, "rw:r16/mem, r16/mem/imm16");
 	reg_instr_clob("xorl", cc_flag_mask, "rw:r32/mem, r32/mem/imm32");
 	reg_instr_clob("xorq", cc_flag_mask, "rw:r64/mem, r64/mem/immi32/imm64");
+	reg_instr_clob("mulb", cc_flag_mask, "r8/mem");
```
lerno, line 348: I seem to recall mul instructions storing the result in EDX and EAX, so you need to have a clobber mask that has CC + RDX + RAX

lerno:
Thank you!

Issue #425, opened 2022-04-10
Introduce single line function declaration


lerno:
Variant 1:
```c
// Statement
fn int square(int x) = return x * x;
fn void foo(int y) = baz(y);
```

Proposal 2:
```c
fn int square(int x) = x * x;
fn void foo(int y) = baz(y);
```

lerno:
Feel free to vote on this one using 👍 and 👎

data-man:
Proposal 3:

```rust
fn int square(int x) {x * x};
fn void foo(int y) {baz(y)};
```

data-man:
Proposal 4:

```rust
fn int[2] square(int x) (x * 2, x * 3);
```

lerno:
3 is ambiguous. 4 would need special parsing.

lerno:
An alternative uses `=>`

```c
fn int square(int x) => return x * x;
fn void foo(int y) => baz(y);

fn int square(int x) => x * x;
fn void foo(int y) => baz(y);
```

data-man:
Proposal 6: :smile: 

```rust
fn int square(int x) -> return x * x;
```

lerno:
I'm trying to reserve "->" in case it's needed later, plus the problem of that operator when converting C code.

lerno:
Two variants are now available for test:
```c
fn int square(int x) = x * x;
fn int square(int x) => x * x;
```

See if it actually adds any value or we remove it.

lerno:
`=` is more likely to be adopted than `=>`

tommy-mitchell:
Would single line functions also work for anonymous functions?

lerno:
@tommy-mitchell They already do.

lerno:
=> was chosen.

PR #566, opened 2022-09-12
Add single function declaration

jb-perrier:
Implementation of proposal 2 : #425 

Since the logic is very similar between proposals this feature can be easily modified accordingly to the chosen proposal.

I had to use COMPOUND_STMT since we can't directly use a RETURN_STMT.
Actually, `fn void myfunc() = 8 * 6 + 1;` will compile, works and return nothing.


lerno:
Macros?

lerno:
But let me merge this one. I'll add => as an alternative syntax as well.

Issue #567, opened 2022-09-13
Language Server support


jb-perrier:
Having a language server for all famous IDE is a key step for spreading C3 into the programming world.

This issue will enable us to track the needs, how it can be done and anything related to this subject.

For now, `--test` mode enable us to get error reports, in order to have auto completion and any feature AST related we need to access the AST generated in parsing part of c3c.

Few solutions comes to my mind rn:
- Using the compiler as a static lib, with little modification we should be able to access the whole ast
- Outputting the AST in a specific format. ( The language server would have to read / parse the outputted file, performance drop here )

Feels free to express your ideas below.

lerno:
C3 used to have an AST printer, but such a format was more informative than intended for LSPs. It's also the concern that the AST is sometimes rewriting things from the concrete tree and as such might not correctly match the LSP output. But a starting point should probably be trying to create a LSP-format AST printing function.

jb-perrier:
`--lsp` totally make sense to me. Are we talking about a binary format or smth like json ?

jb-perrier:
After some reflexion, we should probably not use `lsp` because it means Language server protocol, it's a rpc json format that enable IDE to query the language server about request like Go to definition, ect ..

jb-perrier:
Atfer discord discussion, the `--lsp` will take a [LSP](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/) request in stdin and respond to it, in the stdout.
We will start by supporting very basic request.

lerno:
What's the state of this?

jb-perrier:
Hi @lerno, i should have a branch somewhere with basic communication using lsp protocol.
I will get back to you when i will find it !

jb-perrier:
Just found it : https://github.com/jb-perrier/c3c/tree/lsp
Will put my head in it, i have more time now.

lerno:
I'm looking forward to it!

Akselmo:
Hi, any news on this? :eyes:

jb-perrier:
> Hi, any news on this? 👀

Unfortunately no, i was just stuck to an issue i wasn't able to fix for some reason. I had to switch to other things.

lerno:
Do you know what issue that was?

resolritter:
FYI there's a WIP language server implementation at https://github.com/pherrymason/c3-lsp. It was recently updated with the stdlib for 0.6.0.

lerno:
[...]

PR #574, opened 2022-09-15
Extending JSON utils

jb-perrier:
This PR aims to extend the json utils : 
- Free json object
- Convert json object to string
- Create, modify easily json object using helper funcs

The needs come from the implementation of LSP #567  that require extensive manipulation of JSON objects.

* Edit : In order to keep code size small as possible only necessary func will be added ( json_free ), json should be built directly as string ( hello scratch_buffer ).


jb-perrier:
In order to keep code size small as possible only necessary func will be added ( json_free ), json should be built directly as string ( hello scratch_buffer ).

data-man:
I think [json.h](https://github.com/sheredom/json.h) is the best alternative to this PR.
And [yajl](https://github.com/anjohnson/yajl) also supports JSON5 and stream parsing.

jb-perrier:
As discussed on discord with @lerno , we should not add any dependency to C3. I've removed the whole json creation helpers funcs i was planning, to keep complexity and code size low. I guess when we will rewrite the compiler in C3 a complete JSON stdlib would be written.

src/utils/stringutils.c
```diff
@@ -232,6 +232,27 @@ void scratch_buffer_append_signed_int(int64_t i)
 	scratch_buffer.len += len_needed;
 }
 
+void scratch_buffer_append_double(double d)
+{
+	uint32_t len_needed = (uint32_t)sprintf(&scratch_buffer.str[scratch_buffer.len], "%f", d);
```
lerno, line 237: I seem to recall that %f has a fixed number of decimals. Are 6 decimals (or whatever it is) sufficient?

jb-perrier: Nice catch, i didn't know about that, we should use "%.15g" instead.

PR #577, opened 2022-09-16
Adding $$shufflevector. Bump to 0.3.44

lerno:


data-man:
Maybe rename it to `$$shuffle`?

lerno:
It's going to be wrapped by a dot function, so `foo.shuffle(0, 1, 2, 3)` and maybe `foo.shuffle_with(other, 0, 1, 2, 3)`

PR #578, opened 2022-09-16
Adding $$reverse. Bump to 0.3.45

lerno:


src/compiler/llvm_codegen_expr.c
```diff
@@ -4561,6 +4562,26 @@ INLINE void llvm_emit_shufflevector(GenContext *c, BEValue *result_value, Expr *
...
+{
+	Expr **args = expr->call_expr.arguments;
+	llvm_emit_expr(c, result_value, args[0]);
+	llvm_value_rvalue(c, result_value);
+	Type *rtype = result_value->type;
+	LLVMValueRef arg1 = result_value->value;
+	LLVMValueRef arg2 = LLVMGetPoison(LLVMTypeOf(arg1));
+	LLVMValueRef buff[128];
+	unsigned elements = rtype->array.len;
+	LLVMValueRef *mask_element = elements > 128 ? MALLOC(sizeof(LLVMValueRef)) : buff;
```
data-man, line 4575: Should be `mask_element` freed if it was allocated?

lerno: No, it's arena allocated.

PR #581, opened 2022-09-17
Deprecate LLVM 12 compatibility

data-man:


lerno:
Included in the previous push.

PR #591, opened 2022-09-21
Add `--benchmarking` and `--testing` switches.

data-man:


lerno:
Options are merged into the active target, overriding options. To trigger tests, and benchmarks the idea is to use the the command type (not hooked up yet) to pick one: COMMAND_BENCH, COMMAND_UNIT_TEST. So please instead use those to wire up the constants.

data-man:
> COMMAND_BENCH, COMMAND_UNIT_TEST. So please instead use those to wire up the constants.

I think options are preferable to commands.

src/compiler/compiler.c
```diff
@@ -497,7 +497,7 @@ static const char **target_expand_source_names(const char** dirs, const char **s
 void compile_target(BuildOptions *options)
 {
 	init_default_build_target(&active_target, options);
```
lerno, line 499: Go into init_default_build_target and you'll see the build options being copied into the active target. Do this rather than pulling the options into compile()

lerno:
Can you you resolve the conflicts?

PR #648, opened 2022-10-26
Add `format` field to `Formatter` struct.

data-man:


lerno:
What is the point of this?

data-man:
An user-defined formatters can support a non-standard syntax.

lerno:
format[] is the whole formatting string. So "%s foo %z". Anyone doing output with custom formatting does not know if "%s" or "%z" is the formatting string. In addition the registered function will only be invoked on %s.

lerno:
So is it ok to close this one? I think being able to inject additional formatting types is fundamentally a good idea, but it needs more than access to `format`.

data-man:
> format[] is the whole formatting string.

Yes, it will allow a custom parsing.

lerno:
No it doesn't. Let's say you have the following: `io::printf("%s %s %d", a, a, a)`. What happens here is that if you registered a type Foo with a formatting function, it will first give a callback with the formatting string set to `"%s %s %d"`, then there will be another callback with `"%s %s %d"` again. The formatting function is never invoked on %d.

So:
1. There is no way for the formatting function to know what formatting location it is in given just the formatting string.
2. Only %s ever invokes formatting functions, so no custom formatters like using "%m" to grab the month for a "Time" type will ever work.

My point is therefore that to support custom formatting %<some letter> hooks would be required. If additional formatting hints should be passed to the formatting function, this information needs to be preparsed or at a minimum the actual formatting substring is passed (e.g. "%s")

PR #668, opened 2022-12-06
Add inlined `memcpy` and `memset` builtins and macros.

data-man:


data-man:
For some reason `$$memset_inline` IR isn't correct for LLVM < 15. :(

data-man:
@lerno Maybe remove LLVM IR from `test/test_suite|suite2/stdlib/map.c3t`?

lerno:
Eventually.

lerno:
The reason this is not passing in those tests is probably because some functions in mem moved, which meant the panic function call got a different line number. Try checking what the line should be and update that particular value in the expected IR

lerno:
Thank you

PR #702, opened 2023-01-14
Add override flag to groupadd to support docker build on mac

DeanHnter:
MacOS docker build fails on the groupadd command:
```
RUN groupadd --gid=20 c3c && useradd --gid=20 --uid=20 --create-home --shell /bin/bash c3c:
#6 0.209 groupadd: GID '20' already exists
```
It can be fixed by adding the -o flag


DeanHnter:
This fixed one issue on mac but it seems it could also be resolved by modifying the build-script to not run as the user running it from commandline.

PR #721, opened 2023-01-30
Update docker script and dockerfile

Its-Kenta:
Using Ubuntu 23 throws an error `groupadd: GID '1000' already exists` when trying to build. Ubuntu 22 works fine.
There should be no difference between building C3 on Ubuntu 22 vs 23.

To avoid issues raised it's best to move to single Ubuntu version that builds the compiler.


lerno:
Thank you

PR #727, opened 2023-02-07
Add `@likely`, `@unlikely` and `@expect` macros.

data-man:
Tests passed only with `-O1-O3` without `--safe` switches.


data-man:
> I was planning on perhaps adding C++ style

But why? I think macros will be enough.

PR #750, opened 2023-03-15
Fix the default project template

Ox512:
``c3c`` now expects ``String[] args`` instead of ``char[][] args``. Fixing this means that a project made with ``c3c init`` will now work by default


lerno:
Great! Thanks for seeing that.

PR #764, opened 2023-05-10
Added some matrix math functions

tonis2:
- Added Quaternion to matrix function
- Improved the code for matrix identity functions


data-man:
Why do you remove trigonometric functions from `math` module?

lerno:
Yes, it looks like a bad merge there?

PR #765, opened 2023-05-12
Improve Matrix identity functions and add Quaternion to matrix function

tonis2:


lerno:
I wonder if this isn't better to do as constants:
```c
const Matrix2 IDENTITY2 = { .m = { [0] = 1, [3] = 1 } };
```
And then
```c
define MATRIX2_IDENTITY = matrix::IDENTITY2<double>;
```

tonis2:
Yep, looks better, I'l change it

tonis2:
I'm not sure, what I'm doing wrong,
I replaced macros with constants like this ` const Matrix2x2 IDENTITY2 = { .m = { [0] = 1, [3] = 1 } };`
Then I define them for the type 
` define MATRIX2_IDENTITY = matrix::IDENTITY2<double>;`

But If I try to use it in program, the compiler tells me 
`You cannot cast 'void' into 'Matrix4' (Matrix4x4<double>) even with an explicit cast, so this looks like an error.`

So the defined constant returns void

lerno:
Hmm, maybe it's a bug! Let me try.

lerno:
No that worked for me @tonis2. Looking at your error that's with Matrix4. How does that definition look like?

lerno:
The bug should be resolved now @tonis2

lerno:
Thank you

PR #787, opened 2023-06-19
lib::std::core::bitorder: add read and write

### LINKED ISSUE #785: std: bigendian and littleendian support
Is there any plan to support reading and writing ints in either big or little endian format? Maybe in the encoding module? WDYT

pierrec:
Fix #785.


lerno:
Thank you

PR #793, opened 2023-06-23
lib/std: fix ByteWriter.read_from method

pierrec:
When reading from a stream which does not have an available method, ByteWriter would not make any progress if its buffer was empty.

### DIFF
--- a/test/unit/stdlib/io/bytestream.c3
+++ b/test/unit/stdlib/io/bytestream.c3
@@ -1,6 +1,6 @@
-module std::io @test;
+module std::io;
 
-fn void! bytestream()
+fn void! bytestream() @test
 {
 	ByteReader r;
 	r.init("abc");
@@ -23,7 +23,7 @@ fn void! bytestream()
 	assert(w.as_str() == "helloworldabcbc");
 }
 
-fn void! bytewriter_buffer()
+fn void! bytewriter_buffer() @test
 {
 	ByteWriter writer;
 	char[8] z;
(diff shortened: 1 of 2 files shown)


test/unit/stdlib/io/bytestream.c3
lerno, line 1: Another way to do this with a test reader is to create another module section. So you have first

`module std::io @test;` with the tests, then you declare *another module section*, like `module std::io;` it is now a different section and has different defaults.

test/unit/stdlib/io/bytestream.c3
```diff
@@ -35,3 +35,43 @@ fn void! bytewriter_buffer()
 	assert(o == "hello");
 	assert(@catchof(s.write("xxxx")));
 }
+
+// TestReader only has the read method to trigger the path
```
lerno, line 39: So basically insert `module std::io;` here.

test/unit/stdlib/io/bytestream.c3
```diff
@@ -1,6 +1,6 @@
-module std::io @test;
```
lerno, line 1: So revert to this

test/unit/stdlib/io/bytestream.c3
pierrec: Ah yes, I keep forgetting about this. Too used to single module per file from other languages :).

lerno:
Thank you for this.

PR #807, opened 2023-07-01
lib/std/io/stream: add ReadBuffer and WriteBuffer streams

pierrec:
ReadBuffer and WriterBuffer buffer stream reads and writes to a stream. Useful in situations where the underlying stream is sensitive to the number of read or write calls.

### DIFF
--- a/lib/std/collections/linkedlist.c3
+++ b/lib/std/collections/linkedlist.c3
@@ -82,8 +82,8 @@ fn void LinkedList.link_last(LinkedList* list, Type value) @private
     list.size++;
 }
 
-fn Type! peek(LinkedList* list) => list.first() @inline;
-fn Type! peek_last(LinkedList* list) => list.last() @inline;
+fn Type! LinkedList.peek(LinkedList* list) => list.first() @inline;
+fn Type! LinkedList.peek_last(LinkedList* list) => list.last() @inline;
 
 fn Type! LinkedList.first(LinkedList *list)
 {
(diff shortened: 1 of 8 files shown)


lib/std/io/stream/buffer.c3
```diff
@@ -0,0 +1,138 @@
...
+fn usz! ReadBuffer.read(ReadBuffer *b, char[] bytes)
+{
+    if (b.read_idx == b.write_idx)
+    {
+        if (b.read_idx == 0 && bytes.len >= b.bytes.len)
+        {
+            // Read directly into the input buffer.
+            return b.stream.read(bytes)!;
+        }
+        ReadBuffer.refill(b)!;
```
lerno, line 41: Do you prefer this over `b.refill()!`

lerno:
Thank you! It was no my to do list. I don't know if you want to change the part I commented on.

lib/std/io/stream/buffer.c3
pierrec: Hmm I think so. It is only triggered upon seeing a "big" read (bytes.len higher than the internal buffer and when there is no buffered data). Go's [implementation](https://cs.opensource.google/go/go/+/refs/tags/go1.20.5:src/bufio/bufio.go;l=220) does this as well.

lerno: I mean calling it as `Readbuffer.refill(b)` rather than `b.refill()`. If you want it as a standalone function, I would usually write it as such (i.e. `fn void! refill(Readbuffer* b)`)

PR #808, opened 2023-07-01
lib/std/io/stream: add ReadBuffer and WriteBuffer streams

pierrec:
ReadBuffer and WriterBuffer buffer stream reads and writes to a stream. Useful in situations where the underlying stream is sensitive to the number of read or write calls.


pierrec:
Messed up my git history, pushing a new pull request.

lerno:
Thank you!

PR #810, opened 2023-07-02
Edit Matrix4 perspective fn

tonis2:
I'm not very confident in Matrix calculations, but the previous version breaks my OpenGL renders and this one works perfectly.


lerno:
Looks correct.

PR #813, opened 2023-07-03
add the std::sort::binarysearch module

pierrec:


lib/std/sort/binarysearch.c3
```diff
@@ -0,0 +1,66 @@
...
+ *  -1 if x < y
+ *  0 if x == y
+ *  1 if x > y
+ */
+def Comparer = fn int (void *x, void *y);
+
+/**
+ * Perform a binary search over the sorted array and return the smallest index
+ * in [0, array.len) where cmp(i) is true and cmp(j) is true for j in [i, array.len).
+ * @require is_array_or_subarray(array) "array must be an array or subarray"
```
lerno, line 14: You can actually simply require that it is possible to use foreach on it. I can accept this and then do the change, or I can explain how you can do it. What would you prefer?

lerno:
Thank you!

PR #816, opened 2023-07-04
add basic quicksort support

pierrec:
The algorithm may be improved but I just want to get the basics in place.

### DIFF
--- a/lib/std/hash/crc32.c3
+++ b/lib/std/hash/crc32.c3
@@ -8,17 +8,17 @@ struct Crc32
     uint result;
 }
 
-fn void Crc32.init(Crc32* this, uint seed = 0)
+fn void Crc32.init(&this, uint seed = 0)
 {
     this.result = ~seed;
 }
 
-fn void Crc32.updatec(Crc32* this, char c)
+fn void Crc32.updatec(&this, char c)
 {
     this.result = (this.result >> 8) ^ CRC32_TABLE[(this.result ^ c) & 0xFF];
 }
 
-fn void Crc32.update(Crc32* this, char[] data)
+fn void Crc32.update(&this, char[] data)
 {
     uint result = this.result;
     foreach (char x : data)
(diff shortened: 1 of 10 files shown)


lib/std/hash/crc32.c3
```diff
@@ -8,17 +8,17 @@ struct Crc32
     uint result;
 }
 
-fn void Crc32.init(Crc32* this, uint seed = 0)
+fn void Crc32.init(&this, uint seed = 0)
```
lerno, line 11: Can you change this to `&self`? I found `&self` was more visually distinct and this is what I've been using pervasively.

lib/std/sort/quicksort.c3
```diff
@@ -0,0 +1,66 @@
+module std::sort;
+
+/**
+ * @require is_searchable(list) "The list must be indexable and support .len or .len()"
+ **/
+macro quicksort(list, $Type)
```
lerno, line 6: Why is the $Type here?

lib/std/sort/quicksort.c3
```diff
@@ -0,0 +1,66 @@
...
+ **/
+macro quicksort_with(list, $Type, cmp)
+{
+    (($Type)(list)).sort(cmp);
+}
+
+module std::sort::quicksort<Type, Comparer>;
+import std::sort;
+
+def Quicksort = distinct Type[];
```
lerno, line 23: Do we need to make a type?

lerno:
I am thinking I'd probably just prefer to do:
```c
quicksort(array);
```
Shouldn't that be possible?

pierrec:
I couldnt find a way without it to keep typed comparers.
The partition function needs a type unless we only use cmp(void*, void*).
Or am I missing something?

lerno:
It should be possible, but I can massage this into shape if you want.

pierrec:
Go ahead! Happy to learn more stuff.

lerno:
You can have a look at the current one. Although arguably it's a bit repetitive. And it reveals something of a weakness in the system. I'll discuss it later on discord.

PR #837, opened 2023-07-08
std/lib/sort: update quicksort to use the new generics

pierrec:


lib/std/sort/quicksort.c3
```diff
@@ -1,99 +1,45 @@
...
-}
-
-fn void sort_ref_fn(Type list, ComparerRef cmp)
-{
-    usz len = sort::@len_from_list(list);
-    qsort_ref(list, 0, (isz)len - 1, cmp);
-}
-
-fn void sort(Type list) @if(ELEMENT_COMPARABLE)
+macro quicksort(list, cmp = null)
```
lerno, line 4: Does this really work when cmp is null?

pierrec:
It does work with cmp set to null. Do you want a dedicated test for ti?

lerno:
Thank you.

lerno:
No, it's fine, I saw it working. I think quicksort could probably be a builtin since it's used so much by its name, so you can add a @builtin to its declaration if you want.

PR #853, opened 2023-07-14
improve the sort and collections libs

pierrec:
- Merge binarysearch and binarysearch_with
- Use a generic parameter to indicate a max or min priority queue

### DIFF
--- a/lib/std/collections/priorityqueue.c3
+++ b/lib/std/collections/priorityqueue.c3
@@ -21,94 +21,114 @@
 //  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 //  SOFTWARE.
 module std::collections::priorityqueue(<Type>);
+import std::collections::priorityqueue::private;
+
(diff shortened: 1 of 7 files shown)


lib/std/collections/priorityqueue.c3
```diff
@@ -20,15 +20,17 @@
 //  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 //  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 //  SOFTWARE.
-module std::collections::priorityqueue(<Type>);
+/**
+ * @require $typeof(MAX).kindof == BOOL
+ **/
+module std::collections::priorityqueue(<Type, MAX>);
```
lerno, line 26: I'm thinking that a better interface could probably be to define an enum, and have have that be the second argument, and it has a default to min heap for example.

lerno:
Thank you!

Issue #794, opened 2023-06-23
Create cross platform socket interface


lerno:

pierrec:
I have started work on this. Any specifics you have in mind though?

lerno:
No, just make sure it works for both UDP and TCP.

lerno:
Oh, and we want to try for making the full capabilities available for all platforms from the cross platform interface, but to be able to query for capabilities where it exists on only one platform.

PR #857, opened 2023-07-15
cross platform socket interface

pierrec:
This is an attempt to address #794.


lerno:
It looks like some things are failing?

pierrec:
Yes I will look into it.

pierrec:
Not sure why it fails on msys2-mingw.

PR #861, opened 2023-07-17
support for List in quicksort; add HashMap.@each

pierrec:


lib/std/collections/map.c3
```diff
@@ -195,7 +200,20 @@ fn Key[] HashMap.key_list(&map, Allocator* using = mem::heap())
 	return list;
 }
 
-
+macro HashMap.@walk(map; @body(key, value))
```
lerno, line 203: Rename it `@each`

lerno:
Thank you!

PR #865, opened 2023-07-19
Add is_even and is_odd to math module

data-man:


lerno:
Otherwise people might pull in a [library for it](https://stackoverflow.com/questions/64493511/why-is-there-a-is-even-package-on-npm)

lerno:
Thank you

data-man:
> Otherwise people might pull in a [library for it](https://stackoverflow.com/questions/64493511/why-is-there-a-is-even-package-on-npm)

My code is highly optimized! :-D

PR #866, opened 2023-07-19
Add `has` function to `Range` and `ExclusiveRange`.

data-man:


lerno:
I've been looking at alternative names, because "has" is rarely used. This is what other classes use:

`includes`
`contains`
`in_range`

There are also functions to see if:

1. A range is completely inside a range
2. A range is overlapping another range

And
- Creating the union of two ranges by taking the max and min of both.

data-man:
Ok.

The universal `contains` macro or additional `contains_range` function?

lerno:
As the method name.

lerno:
I'll accept this and update.

PR #872, opened 2023-07-21
`std::bits` fixes.

data-man:


lib/std/bits.c3
```diff
@@ -1,16 +1,14 @@
...
 /**
- * @require types::is_intlike($typeof(i)) `The input must be an integer or integer vector`
+ * @require types::is_intlike($typeof(self)) `The input must be an integer or integer vector`
  **/
-macro reverse(i) => $$bitreverse(i);
+macro reverse(self) => $$bitreverse(self);
 
 /**
- * @require types::is_intlike($typeof(i)) `The input must be an integer or integer vector`
+ * @require types::is_intlike($typeof(self)) `The input must be an integer or integer vector`
```
lerno, line 9: "self" should not be used on non-methods

lerno:
The bin_literal test needs to be updated to account for the change in variable name.

data-man:
I removed unrelated code for literals.

lerno:
I want to keep that code because it is still testing things that are not tested otherwise.

data-man:
> things that are not tested otherwise

`rotr` and `rotl`?

lerno:
Not just that, sometimes there are small changes that might be correct or not and sometimes a completely unrelated test will catch it.

data-man:
Fixed now, I hope.

lerno:
Thanks!

PR #882, opened 2023-07-25
add is_absolute and absolute methods to path::Path

pierrec:


pierrec:
Yes. Found an issue. Will re-submit once fixed.

pierrec:
Strange that CI fails on the wasm test. It passes on my laptop.

lerno:
I'll accept and check.

PR #883, opened 2023-07-26
use IoError.UNSUPPORTED_OPERATION instead of asserts; improve Path.walk

pierrec:


lib/std/io/os/fileinfo.c3
```diff
@@ -85,7 +85,7 @@ fn bool native_file_or_dir_exists(String path)
                 return posix::access(path.zstr_tcopy(), 0 /* F_OK */) != -1;
             };
         $default:
-            unreachable("Not supported");
+			return IoError.UNSUPPORTED_OPERATION?;
```
lerno, line 88: You need to have unreachable here.

lerno:
Thank you

PR #892, opened 2023-07-27
lib/std/io: output "Invalid enum" if the enum value is out of range i…

pierrec:
…nstead of crashing

### DIFF
--- a/lib/std/io/io_printf.c3
+++ b/lib/std/io/io_printf.c3
@@ -160,7 +160,9 @@ fn void! Formatter.out_str(&self, any arg) @private
 			return self.out_str(*(any*)arg);
 		case ENUM:
 			if (self.print_with_function(arg)!) return;
-			return self.out_substr(arg.type.names[types::any_to_int(arg, usz)!!]);
+			usz i = types::any_to_int(arg, usz)!!;
+			assert(i < arg.type.names.len, "Illegal enum value found, numerical value was %d.", i);
+			return self.out_substr(arg.type.names[i]);
 		case STRUCT:
 			if (self.print_with_function(arg)!) return;
 			return self.out_substr("<struct>");


lerno:
Did this actually happen?

pierrec:
Yep, in a case where the enum was not initialized.

lerno:
As in malloced with trash data?

lerno:
Did you get an "out of bounds" error on --safe mode? Because that would be the preferred error.

pierrec:
Actually I am unusure why it is not initialized.
The error is an OOB as you expected:
```
ERROR: 'Array index out of bounds (array had size 120, index was 200)'
  in function std::core::builtin::panicf (builtin.c3:117)
  in function std::io::out_str (io_printf.c3:165)
  in function std::io::vprintf (io_printf.c3:402)
  in function std::io::printfn (io_printf.c3:41)
  in function doc3::expect_ident (parse.c3:680)
  in function doc3::expect_multi_ident (parse.c3:669)
  in function doc3::expect_module (parse.c3:145)
  in function doc3::parse (parse.c3:58)
  in function doc3::walk (project.c3:33)
  in function std::io::path::walk (path.c3:422)
  in function doc3::load (project.c3:22)
  in function doc3::main (main.c3:8)
zsh: illegal hardware instruction  ../doc3
```

BUT the enum field is part of an union (token):
```
struct Lexer
{
	Allocator* allocator;
	Stream reader;
	char[] buf;
	TokenTrie tokens;
	Ident ident;
	bool peeked;
	Kind kind;
	Token expected;
	uint line;
	uint column;
	uint span;
	union {
		ulong x;       // UINT
		String string; // STRING, IDENTIFIER
		char rune;     // RUNE
		Token token;   // TOKEN
	}
}

```

lerno:
Ok, so what I suggest is to add an assert which bounds checks and has an error `"Illegal enum value found, numerical value was %d.", ordinal`

lerno:
Thanks, that will be much better. Better to crash to let people know their data is messed up.

PR #895, opened 2023-07-28
add std::io::stream::ByteBuffer; fix std::io::Path::walk

pierrec:


lib/std/io/path.c3
```diff
@@ -411,13 +411,16 @@ def PathWalker = fn bool(Path, bool is_dir, void*);
 
 fn bool! Path.walk(self, PathWalker w, void* data, Allocator* using = mem::heap())
 {
-	Path abs = self.absolute(using)!;
-	defer abs.free();
+	Path abs = self;
```
lerno, line 414: Couldn't this be written with the temp allocator instead of doing alloc / free?

lib/std/io/path.c3
```diff
@@ -411,17 +411,19 @@ def PathWalker = fn bool(Path, bool is_dir, void*);
 
 fn bool! Path.walk(self, PathWalker w, void* data, Allocator* using = mem::heap())
```
lerno, line 412: If it uses the temp allocator, then there is no need to provide a "using"

pierrec: I tried to remove it but Path.append provide an invalid path otherwise. I will look into it tomorrow.

lib/std/io/path.c3
```diff
@@ -409,19 +409,30 @@ fn String Path.root_directory(self)
...
-	PathList files = ls(abs, .using = using)!;
-	foreach (f : files)
+/*
+ * Walk the path recursively. PathWalker is run on every file and
+ * directory found. Return true to abort the walk.
+ */
+fn bool! Path.walk(self, PathWalker w, void* data)
+{
+	$if env::LINUX:
+		const PATH_MAX = 4 << 10;
```
lerno, line 419: Set this to a conservatively low value, or it will allocate prohibitively much memory on the stack, I recommend something like 256, if it overflows it will use the heap allocator.

lerno: Or maybe a little more, but you get the point.

pierrec: I actually felt unsure about the right value to set here. Windows (10+) and macOS seem to not limit the path size, Linux keeping it at 4Kb from what I gathered.
If it switch over to the heap then a low value is fine.

lerno:
Thanks!

PR #904, opened 2023-07-31
std/lib/io: add Scanner

pierrec:


lib/std/io/stream/scanner.c3
```diff
@@ -0,0 +1,93 @@
...
+		return self.buf[self.anchor:len];
+	}
+	if (self.split_idx == 0 || self.read_idx < self.buf.len)
+	{
+		// Split character not found with maximized search, abort.
+		// Split character not found and already read as much as possible.
+		return SearchResult.MISSING?;
+	}
+	// Split character not found: maximize the search and try one more time.
+	self.buf[:n] = buf[..];
```
lerno, line 68: Is it possible to use a ring buffer to avoid copies?

pierrec: ah yes, also working on one but not yet ready. Will update when it is.

lerno: Should I wait then?

pierrec: yes. WIll start working on it again.

pierrec: Hmm. Since the match is returned as a contiguous slice, I cannot use a ring buffer or I need to use half of the provided buffer.

pierrec: Doing small refactoring though.

lerno: Let me know when you're ready for it to be added

lib/std/core/string.c3
```diff
@@ -404,7 +404,7 @@ fn Char32[]! String.to_utf32(s, Allocator* using = mem::heap())
 
 fn void String.convert_ascii_to_lower(s)
 {
-	foreach (&c : s) if (*c >= 'A' && *c <= 'Z') *c += 'a' - 'A';
+	foreach (&c : s) if ((*c).is_upper()) *c += 'a' - 'A';
```
lerno, line 407: Hmm, does it need the deref?

pierrec: It does not indeed.

lerno:
Ready to merge?

PR #907, opened 2023-08-01
std/lib: simplify String.{,r}index_of and improve speed for the index…

pierrec:
…_of one

### DIFF
--- a/lib/std/core/string.c3
+++ b/lib/std/core/string.c3
@@ -252,24 +252,12 @@ fn usz! String.rindex_of_char(s, char needle)
 fn usz! String.index_of(s, String needle)
 {
 	usz needed = needle.len;
-	char search = needle[0];
-	if (needed == 1) return s.index_of_char(search);
-	usz match = 0;
-	usz index_start = 0;
-	foreach (usz i, char c : s)
+	if (needed > 0 && s.len >= needed)
 	{
-		if (c == search)
+		char first = needle[0];
+		foreach (i, c: s[..^needed])
 		{
-			if (!match) index_start = i;
-			match++;
-			if (match == needed) return index_start;
-			search = needle[match];
-			continue;
-		}
-		if (match)
-		{
-			match = 0;
-			search = needle[0];
+			if (c == first && s[i:needed] == needle) return i;
 		}
 	}
 	return SearchResult.MISSING?;
@@ -289,24 +277,12 @@ fn usz! String.index_of(s, String needle)
 fn usz! String.rindex_of(s, String needle)
 {
 	usz needed = needle.len;
-	char search = needle[^1];
-	if (needed == 1) return s.rindex_of_char(search);
-	usz match;
-	usz index_start;
-	foreach_r (usz i, char c : s)
+	if (needed > 0 && s.len >= needed)
 	{
-		if (c == search)
+		char first = needle[0];
+		foreach_r (i, c: s[..^needed])
 		{
-			if (!match) index_start = i;
-			match++;
-			if (match == needed) return index_start - needle.len + 1;
-			search = needle[^(match + 1)];
-			continue;
-		}
-		if (match)
-		{
-			match = 0;
-			search = needle[^1];
+			if (c == first && s[i:needed] == needle) return i;
 		}
 	}
 	return SearchResult.MISSING?;
--- a/lib/std/collections/enummap.c3
+++ b/lib/std/collections/enummap.c3
@@ -7,7 +7,7 @@ struct EnumMap
 
 fn void EnumMap.init(&self, ValueType init_value)
 {
-	foreach(&a : self.values)
+	foreach (&a : self.values)
 	{
 		*a = init_value;
 	}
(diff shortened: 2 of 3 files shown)


lib/std/core/string.c3
```diff
@@ -252,24 +252,12 @@ fn usz! String.rindex_of_char(s, char needle)
...
-	char search = needle[0];
-	if (needed == 1) return s.index_of_char(search);
-	usz match = 0;
-	usz index_start = 0;
-	foreach (usz i, char c : s)
+	if (needed > 0 && s.len >= needed)
 	{
-		if (c == search)
+		char first = needle[0];
+		foreach (i, c: s[:^needed - 1])
```
lerno, line 258: Or just `s[..^needed]`

lerno:
Were there already a unit tests on these functions?

pierrec:
Not really although other tests do trigger them (the scanner for instance). WIll add some specific ones for those methods as they are important.

lib/std/core/string.c3
```diff
@@ -252,24 +252,12 @@ fn usz! String.rindex_of_char(s, char needle)
...
-			match++;
-			if (match == needed) return index_start;
-			search = needle[match];
-			continue;
-		}
-		if (match)
-		{
-			match = 0;
-			search = needle[0];
+			if (c == first && s[i:needed] == needle) return i;
```
lerno, line 260: Hmm.. an even simpler solution might be:
```c
usz last = s.len - needed;
for (usz i = 0; i <= last; i++)
{
  if (s[i:needed] == needle) return i;
}
```
I'm not sure which one is fastest: the original, your new one or the above one. It might be worth looking at the final asm at -O2 or -O3

lib/std/sort/sort.c3
```diff
@@ -18,4 +18,34 @@ macro bool is_comparer(cmp, list)
...
+
+/**
+ * Compare two comparable values, returning:
+ * - a negative value if a < b
+ * - a positive value if a > b
+ * - zero if a == b
+ *
+ * @require types::inner_kind($typeof(a)) == types::inner_kind($typeof(b))
+ **/
+macro int cmp(a, b)
```
lerno, line 31: Any comparison functions should go into builtin_comparison.c3

lib/std/core/builtin_comparison.c3
```diff
@@ -112,3 +112,32 @@ macro max(x, ...) @builtin
...
+		$case VECTOR:
+			if (a.len != b.len) return (b.len - a.len);
+			foreach (i, x : a)
+			{
+				if (equals(x, b[i])) continue;
+				return less(x, b[i]) ? -1 : 1;
+			}
+			return 0;
+		$default:
+			$error("Values are not comparable");
```
lerno, line 141: This is incorrect. If the type implements compare_to, then the result is a.compare_to(b)

So a type may implement either "less" or "compare_to" - or both. It then supports comparisons. If it supports "less",
then you can get the compare_to(a, b) by doing `a.less(b) ? -1 : (b.less(a) ? 1 : 0)` if it supports compare_to, then you just return `a.compare_to(b)`.

pierrec: I see. Will update accordingly.

lib/std/core/builtin_comparison.c3
```diff
@@ -112,3 +112,53 @@ macro max(x, ...) @builtin
...
+						int c = x.compare_to(y);
+						if (c != 0) return c;
+					$case $defined(x.less):
+						bool c1 = x.less(y);
+						bool c2 = y.less(x);
+						if (!(c1 && c2)) c1 ? -1 : 1;
+					$default:
+						int c = (int)(x - y);
+						if (c != 0) return c;
+				$endswitch
```
lerno, line 158: Consider simplifying the above to:
```c
foreach (i, x : a)
{
  var c = cmp(x, b[i]);
  if (c != 0) return c;
}

File: lib/std/core/builtin_comparison.c3 [hunk 6]
```diff
@@ -112,3 +112,53 @@ macro max(x, ...) @builtin
...
+			return cmp(*a, *b);
+		$case BITSTRUCT:
+		$case STRUCT:
+			$switch
+				$case $defined(a.compare_to):
+					return a.compare_to(b);
+				$case $defined(a.less):
+					bool c1 = a.less(b);
+					bool c2 = b.less(a);
+					return c1 ? -1 : c2 ? 1 : 0;
```
lerno, line 136: This does one more comparison than strictly necessary though?

lib/std/core/builtin_comparison.c3
```diff
@@ -112,3 +112,53 @@ macro max(x, ...) @builtin
 	$endif
 }
 
+/**
+ * Compare two comparable values, returning:
+ * - a negative value if a < b
+ * - a positive value if a > b
+ * - zero if a == b
+ *
+ * @require types::inner_kind($typeof(a)) == types::inner_kind($typeof(b))
```
lerno, line 121: You should probably not be using `inner_kind` here, it simply removes "distinct" around an alias.

I think you want either that:
1. Both are pointers and the pointees are the same and comparable.
2. Both are arrays, vectors or subarrays and the element type is the same. And that element type is comparable
3. Or the types are the same and the type is comparable.

lerno:
Let's remove the cmp macro for now, that needs more of a discussion, and I'll approve the rest.

lerno:
Thank you!

PR #910, opened 2023-08-03
lib/std/io: Add File.readbytes and File.writebytes

OdnetninI:
It is very common when reading files (using fread) to not read the full file.
Depending on the specific implementation, you can end up with a bigger buffer than the data you want to read.

This patch adds File.readbytes and File.writebytes for these cases.
File.read and File.write reamin identical.

In the case that the amount of bytes to read or write is larger than the buffer, then IoError.OVERFLOW is returned.

### DIFF
--- a/lib/std/io/io_file.c3
+++ b/lib/std/io/io_file.c3
@@ -95,11 +95,22 @@ fn bool File.eof(self) @inline
 }
 
 /**
+ * @require self.file `File must be initialized`
  * @param [in] buffer
  */
 fn usz! File.read(self, char[] buffer)
 {
-	return os::native_fread(self.file, buffer);
+	return os::native_fread(self.file, buffer, buffer.len);
+}
+
+/**
+ * @require self.file `File must be initialized`
+ * @param [in] buffer
+ */
+fn usz! File.readbytes(self, char[] buffer, usz size)
+{
+	if (size > buffer.len) return IoError.OVERFLOW?;
+	return os::native_fread(self.file, buffer, size);
 }
 
 /**
@@ -108,7 +119,17 @@ fn usz! File.read(self, char[] buffer)
  */
 fn usz! File.write(self, char[] buffer)
 {
-	return os::native_fwrite(self.file, buffer);
+	return os::native_fwrite(self.file, buffer, buffer.len);
+}
+
+/**
+ * @param [out] buffer
+ * @require self.file `File must be initialized`
+ */
+fn usz! File.writebytes(self, char[] buffer, usz size)
+{
+	if (size > buffer.len) return IoError.OVERFLOW?;
+	return os::native_fwrite(self.file, buffer, size);
 }
 
 /**
--- a/lib/std/io/os/file_libc.c3
+++ b/lib/std/io/os/file_libc.c3
@@ -70,14 +70,14 @@ fn usz! native_ftell(CFile file) @inline
 	return index >= 0 ? (usz)index : file_seek_errno()?;
 }
 
-fn usz! native_fwrite(CFile file, char[] buffer) @inline
+fn usz! native_fwrite(CFile file, char[] buffer, usz size) @inline
 {
-	return libc::fwrite(buffer.ptr, 1, buffer.len, file);
+	return libc::fwrite(buffer.ptr, 1, size, file);
 }
 
-fn usz! native_fread(CFile file, char[] buffer) @inline
+fn usz! native_fread(CFile file, char[] buffer, usz size) @inline
 {
-	return libc::fread(buffer.ptr, 1, buffer.len, file);
+	return libc::fread(buffer.ptr, 1, size, file);
 }
 
 macro anyfault file_open_errno() @local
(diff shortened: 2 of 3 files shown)


lerno:
Actually this is less of a problem than one might think with the current construct. Let's say you have a buffer with 256 bytes and you only want to read 16:
```c
char[] buffer = ... // Get the buffer in some way.
file.read(buffer); // This would read all.
file.read(buffer[:16]); // This would read up to 16 because we pass a slice from the first buffer which is only 16 bytes.
```

lerno:
If it is necessary to add such a method, then using default arguments is probably better:
```c
fn usz! File.read(self, char[] buffer, usz max = 0)
{
   return os::native_fread(self.file, max ? buffer[:max] : buffer);
}
```

OdnetninI:
The second solution is bassicaly identical to the first one but with the slicing hidden by the library function (I think it is dirtier).
Personally I am not a big fan of array slicing, but I can see how they can be useful.
Also, that solution is cleaner as you don't need an additional check for if it goes out of bounds.

I am not sure if this approach is the correct solution for these cases.
If I find any case when slicing is not possible, I will reopen this discussion.

lerno:
I think a `limit` could probably be nice actually, One thing one might think about a bit would be if the `limit` is greater than the  buffer `len`.

Is this (a) an error or (b) fine, and then the limit would automatically adjust to being the same as the buffer length.

I think the latter might certainly come in handy at times, but I am afraid such behaviour could hide bugs, like you think
`file.read(foo, 16)` is going to read up to 16 bytes, and reading less you assume it's because there is nothing more left to read, but the actual reason is the buffer was too small. One could imagine the code wasn't even clear at the location that this might happen, e.g:

```c
while (true)
{
  usz read = file.read(buffer, max_read)!;
  if (read < max_read) 
  {
     finalize_read(buffer[:read]);
  }  
  add_read(buffer[:max_read]);
}
```
If one would pick the latter behaviour it would not be obvious that this code is assuming `buffer.len >= max_read` in order to work correctly.

But again that depends on what usage is the more common one.

OdnetninI:
Agree, an optional limit could be a cool feature.

I have done a quick search of fread usages and I see the folowing:
- Reading a byte
- Reading a magic (typically 4 bytes)
- Reading the full file
- Reading an stuct (like a file header)
- Reading an arbitrary set of bytes depending on a previously read value or an input from other file/user

I have also seen that sometimes just a hardcoded value is readed, but it can be covered with the curretn array slicing approach.

How do we read structs easily with the current implementaton?

It would be great to do something like:

```C
struct SavedData {
....
} 
```

```C
File file = open(....);
SavedData data_from_file;
file.read(data_from_file); // or file.read(&data_from_file);
```

pierrec:
For this use case, I use the LimitReader to read X amount of bytes into a buffer and a custom macro to load the buffer's data into the struct. From there I can check the magic number and other parameters. No need to anything else specific.

The macro is something like this (geared towards little endian numbers):
```
/**
 * Unsafely load the content of buf into s.
 * @require $typeof(s).kindof == POINTER
 * @require $typeof(buf).kindof == SUBARRAY && buf.len <= $typefrom($typeof(s).inner).sizeof
 **/
macro read_struct(s, buf)
{
	var $Type = $typefrom($typeof(s).inner);
	$foreach ($m : $Type.membersof)
		$if $m.kindof == UNSIGNED_INT:
			$switch ($m.sizeof)
			$case 1:
				s.$eval($m.nameof) = buf[0];
			$case 2:
				s.$eval($m.nameof) = bitorder::read(buf, UShortLE);
			$case 4:
				s.$eval($m.nameof) = bitorder::read(buf, UIntLE);
			$case 8:
				s.$eval($m.nameof) = bitorder::read(buf, ULongLE);
			$endswitch
			buf = buf[$m.sizeof..];
		$endif
	$endforeach
}
```

OdnetninI:
I get the idea of the macro, and for several scenarios, it will work perfectly fine.

But for simple structs with no checks, I was thinking that the C version is much simple and direct:
```C
struct Data data;
fread((char*)&data, 1, sizeof(struct Data), file);
```

lerno:
Yeah, note that we could consider a simple function like this:
```c
macro usz! read_data(self, reference)
{
   char[] buffer = ((char*)reference)[:$sizeof(*reference)];
   return read(self, buffer);
}
```
Now we can do:
```c
Data data;
file.read_data(&data)!;
```

OdnetninI:
I like that one!!
I am thinking that we should check if we read enough data to fill the "reference"... but in that case read_data will return EOF or other errors... So, it should be fine.

I was wondering if this should be inside the stdlib or not..
Either of the case, I think this is common enough... but probably it will be enough adding it as an example in the docs???

lerno:
I think this is a stdlib feature.

OdnetninI:
Then, is it fine as a macro, or should it be a function?

lerno:
Probably better as a function:
```c
fn usz! Fileread_data(self, any ref)
{
   char[] buffer = ((char*)ref.ptr)[:ref.type.sizeof];
   return self.read(buffer);
}
```

OdnetninI:
I have some data and structs to test it.
If you want I can do it and get back to you.

lerno:
That's good then. Maybe it should be renamed `read_any`, `read_value` or something. I don't know what's the best.

OdnetninI:
Maby read_any is better?
read_value seems like it is going to read only one value of the struct.

lerno:
I added read_any / write_any now.

OdnetninI:
I can confirm it passed my test :)
(checked with the "[Fix void[] -> char[] cast.]" commit).

lerno:
Feel free to add unit tests if you want!

PR #914, opened 2023-08-08
Tests:std/lib/io/file: Added unit tests for read_any and write_any

OdnetninI:
Added an unit test for the recently added File.read_any and File.write_any functions.


lerno:
Thank you.

PR #915, opened 2023-08-08
README: Updated instructions for AUR

OdnetninI:
ArchLinux's packaged LLVM has some issues with LLVM projects.
The updated AUR package will download LLVM 16 for Ubuntu 23.04 and use it for compiling the c3c compiler.


lerno:
Thank you!

PR #930, opened 2023-08-15
fix module path for RingBuffer

pierrec:


lerno:
Looks like a test needed update as well.

lerno:
Thank you.

PR #932, opened 2023-08-15
improve tests

pierrec:
First pass at improving tests.

### DIFF
--- a/lib/std/core/runtime.c3
+++ b/lib/std/core/runtime.c3
@@ -2,6 +2,7 @@
 // Use of this source code is governed by the MIT license
 // a copy of which can be found in the LICENSE_STDLIB file.
 module std::core::runtime;
+import libc;
 
 struct StackTrace
 {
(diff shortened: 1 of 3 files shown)


lib/std/core/runtime.c3
```diff
@@ -54,23 +73,40 @@ fn void test_panic(String message, String file, String function, uint line)
 	libc::longjmp(&current_runner.buf, 1);
 }
 
-fn bool TestRunner.run(TestRunner* runner)
+fn bool TestRunner.run(&self)
 {
-	current_runner = runner;
+	assert(self.test_names.len == self.test_fns.len);
+	TestUnit[] units = malloc_checked(TestUnit, self.test_names.len, .using = mem::heap())!!;
```
lerno, line 79: Use `malloc` instead?

lib/std/core/runtime.c3
```diff
@@ -54,23 +73,40 @@ fn void test_panic(String message, String file, String function, uint line)
...
-	int tests = runner.test_names.len;
+	int tests = units.len;
 	io::printn("----- TESTS -----");
-	foreach(i, String name : runner.test_names)
+	DString name;
+	name.tinit();
+	foreach(unit : units)
 	{
-		io::printf("Testing %s ... ", name);
+		defer name.chop(0);
```
lerno, line 101: name.clear()

lib/std/core/runtime.c3
lerno: `TestUnit[] units = malloc(TestUnit, self.test_names.len);`

lerno:
Thank you for this.

PR #935, opened 2023-08-16
add NanoDuration.to_format()

pierrec:


lib/std/time/datetime.c3
```diff
@@ -145,12 +145,7 @@ fn bool DateTime.before(&self, DateTime compare)
 
 fn int DateTime.compare_to(&self, DateTime compare)
 {
-	switch
-	{
-		case self.time > compare.time: return 1;
-		case self.time < compare.time: return -1;
-		default: return 0;
-	}
+	return (int)(self.time - compare.time);
```
lerno, line 148: Unfortunately this assumes that the time diff is less than 2^31. It can be safely extended by using `math::sign`, but will then still fail on ranges where `self.time - compare.time` overflows a Long. If you want things like this switch-less use this:
```c
Time a = self.time;
Time b = compare.time;
return (int)(a > b) - (int)(a < b);
```

lerno: I'll add it to the standard library.

lerno: So, you can use `return compare_to(a, b);` now.

pierrec: nice!

lerno:
Thank you

PR #941, opened 2023-08-22
Reimplement QuickSort (non-recursive modification)

### LINKED ISSUE #931: quicksort makes program segfaults when run
The cmp function might not be correct but the program should not crash.

```
module testing;
import std::io;
import std::sort;

fn void! main()
{
	String[] list = {
		"http::header_test::header_scan",
		"text_test::test_render_tag1",
		"text_test::test_render_template",
	};
	quicksort(list, &cmp);
}

fn int cmp(String a, String b)
{
	usz an = a.len;
	usz bn = b.len;
	if (an < bn)
	{
		foreach (i, c : a)
		{
			if (c > b[i]) return 1;
		}
		return -1;
	}
	foreach (i, c : b)
	{
		if (c > a[i]) return 1;
	}
	return -1;
}
```

data-man:
Based on https://alienryderflex.com/quicksort by Darel Rex Finley, Public Domain.

Maybe closes #931. :)

### DIFF
--- a/lib/std/sort/quicksort.c3
+++ b/lib/std/sort/quicksort.c3
@@ -18,51 +18,68 @@ module std::sort::qs(<Type, Comparer>);
 
 def ElementType = $typeof(Type{}[0]);
 
+struct StackElementItem @private
+{
+	isz low;
+	isz high;
+}
+
+def Stack = StackElementItem[64] @private;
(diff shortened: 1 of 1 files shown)


lerno:
Thank you!

PR #948, opened 2023-08-24
Fix calling `llvm::writeArchive` for LLVM > 17.

data-man:


OdnetninI:
Now it is upstream.
This should be merged now.

lerno:
Thank you.

PR #951, opened 2023-08-24
Tune `@expect`, `@likely`, `@unlikely` and `@prefetch` macros.

data-man:


lib/std/core/builtin.c3
```diff
@@ -193,11 +193,15 @@ macro enum_by_name($Type, String enum_name) @builtin
  **/
 macro bool @likely(bool #value, $probability = 1.0) @builtin
 {
-	$if $probability == 1.0:
+	$if env::BUILTIN_EXPECT_IS_DISABLED:
```
lerno, line 196: The indentation here got screwed up.

lib/std/core/builtin.c3
```diff
@@ -209,11 +213,15 @@ macro bool @likely(bool #value, $probability = 1.0) @builtin
  **/
 macro bool @unlikely(bool #value, $probability = 1.0) @builtin
 {
-	$if $probability == 1.0:
+	$if env::BUILTIN_EXPECT_IS_DISABLED:
```
lerno, line 216: Same here. Note that it's probably better to write this as a $switch:
```c
$switch
$case env::BUILTIN_EXPECT_IS_DISABLED:
    return #value;
$case $probability == 1.0:
    return $$expect(#value, false);
 $default:
    return $$expect_with_probability(#value, false, $probability);
 $endswitch
 ```

File: lib/std/core/builtin.c3 [hunk 3]
```diff
@@ -249,9 +261,11 @@ enum PrefetchLocality
  * @param $locality `Locality ranging from none to extremely local`
  * @param $write `Prefetch for write, otherwise prefetch for read.`
  **/
-macro prefetch(void* ptr, PrefetchLocality $locality = VERY_NEAR, bool $write = false) @builtin
+macro @prefetch(void* ptr, PrefetchLocality $locality = VERY_NEAR, bool $write = false) @builtin
```
lerno, line 264: Why is this changed to an `@`?

PR #953, opened 2023-08-25
std/collections: add RingBuffer.popc()

pierrec:


data-man:
What is meant the suffix "c()" in functions, "char"?
I think this is bad naming, since `RingBuffer` is generic.

lerno:
Yes, that could be improved I think.

pierrec:
Originally I had it for char only and then extended it to any type. And the name stuck. Easy to change and happy to have new  more meani-ng ful names!

PR #956, opened 2023-08-26
Implement builtin benchmarks.

data-man:
For testing:
`test_benchmarks.c3`
```
import std::io;

fn void! bench1() @benchmark
{
}

fn void! bench2() @benchmark
{
}
```
`$ c3c compile-benchmark test_benchmarks.c3`


src/compiler/sema_decls.c
```diff
@@ -2593,7 +2608,7 @@ static inline bool sema_analyse_func(SemaContext *context, Decl *decl, bool *era
 		}
 		if (decl->name == kw_main)
 		{
-			if (is_test) SEMA_ERROR(decl, "Main functions may not be annotated @test.");
+			if (is_test || is_benchmark) SEMA_ERROR(decl, "Main functions may not be annotated @test and @benchmark.");
```
lerno, line 2611: Should be "nor" rather than "and"

data-man: This is "or".

data-man:
```
fn void! main() @test // or @benchmark
{
}
```
`$ c3c compile-test test_bad.c3` // or `c3c compile-benchmark test_bad.c3`
```
1: fn void! main() @test // or @benchmark
            ^^^^
(.../test_bad.c3:1:10) Error: Main functions may not be annotated @test and @benchmark.
```

data-man: Oh, you wrote about an error messages! Fixed, I hope.

lerno: Yes, the message :D

lerno:
Thank you.

PR #958, opened 2023-08-27
Architecture generic Atomics (C11 + extras)

OdnetninI:
Using compare_exchange is possible to implement C11 atomics in a generic way.
The implementation used for these atomics use the well known do_while_cas loop.

- Most of c11 atomics is supported, also max and min have been added.
- Made in a way where floating point numbers is also supported
- Attached tests only account for add, sub, float add, float sub, max, min.

A later benchmark should be done when hardware atomics are implemented.

### DIFF
--- a/lib/std/atomic.c3
+++ b/lib/std/atomic.c3
@@ -8,16 +8,27 @@ module std::atomic;
  * @param [in] y "the value to be added to ptr."
  * @param $ordering "atomic ordering of the load, defaults to SEQ_CONSISTENT"
  * @return "returns the old value of ptr"
+ * 
+ * @require types::is_int($typeof(*ptr)) || types::is_float($typeof(*ptr)) "Only integer/float pointers may be used."
  **/
 macro fetch_add(ptr, y, AtomicOrdering $ordering = SEQ_CONSISTENT)
 {
+	var $StorageType = $typefrom(types::get_atomic_compatible_type($typeof(*ptr)));
+
+  $StorageType* storage_ptr = ($StorageType*)ptr;
+
 	$typeof(*ptr) old_value;
   $typeof(*ptr) new_value;
+
+  $StorageType storage_old_value;
+  $StorageType storage_new_value;
   
   do {
-    old_value = $$atomic_load(ptr, false, $ordering.ordinal);
+    storage_old_value = $$atomic_load(storage_ptr, false, $ordering.ordinal);
+    old_value = bitcast(storage_old_value, $typeof(*ptr));
     new_value = old_value + y;
-  } while (mem::compare_exchange(ptr, old_value, new_value, $ordering, $ordering) != old_value);
+    storage_new_value = bitcast(new_value, $StorageType);
+  } while (mem::compare_exchange(storage_ptr, storage_old_value, storage_new_value, $ordering, $ordering) != storage_old_value);
 
   return old_value;
 }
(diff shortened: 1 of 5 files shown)


lib/std/atomic.c3
```diff
@@ -1,5 +1,325 @@
-// Copyright (c) 2021 Christoffer Lerno. All rights reserved.
+// Copyright (c) 2023 Eduardo José Gómez Hernández. All rights reserved.
 // Use of this source code is governed by the MIT license
 // a copy of which can be found in the LICENSE_STDLIB file.
 module std::atomic;
 
+/**
+ * @param [in] ptr "the variable or dereferenced pointer to the data."
```
lerno, line 7: You can use `[&in]` to require the pointer to be non-null.

lib/std/core/types.c3
```diff
@@ -208,6 +208,32 @@ macro bool may_load_atomic($Type)
 	$endswitch
 }
 
+macro get_atomic_compatible_type($Type)
```
lerno, line 211: "lower_to_atomic_compatible_type" as name perhaps?

lerno:
I was thinking that this could perhaps be a generic type, so rather than the functions, something like this:
```c
module std::atomic::types<Type>;

struct Atomic
{
   Type type; // <- actually, find a suitable type
}

fn Type Atomic.add_and_get(&self, Type value)
{ ... }

fn Type Atomic.sub_and_get(&self, Type value)
{ ... }
```

Etc so then usage would be:

```c
Atomic(<int>) my_int = 0;
int test = my_int.add_and_get(12);
```

OdnetninI:
That is the current way done in C++.
However, there is no direct benefit of it. Most of the time it just bothers programmers.

In C atomics it is just a volatile pointer which is closer to the hardware operation.

The question is to decide creating the type or not is: Is there any benefit of having a new data type for shared data between processors?

OdnetninI:
The question is very tricky because in a lot of parallel programs with shared matrixes or arrays, sometimes only some of the data is shared (borders between processors).

When doing porting some programs to modern programming styles I found that, and the solution is not always clean.
That is why I deceided to use the raw pointer.

lerno:
Sorry, my code example was wrong. I think having an atomic communicates more than the C variant. One does not exclude the other though.

OdnetninI:
I am not sure if I understand what you mean by `atomic communicates`.

OdnetninI:
Also, while implementing the Atomic type, a problem arise.
The ordering is a constant that needs to be known at compile time.
The solution could be something like:

```C
fn Type Atomic.add(&self, Type value, AtomicOrdering ordering = SEQ_CONSISTENT)
{
  switch(ordering) 
  {
     case SEQ_CONSISTENT: atomic::fetch_add(self.data, value, SEQ_CONSISTENT);
     case ACQUIRE:  atomic::fetch_add(self.data, value, ACQUIRE);
     ....
  }
}
```

So I think that, for now. The current C style atomics cover most of the cases in the traditional style.

OdnetninI:
Even being a bit ugly, I just did it.
Also an unit test is added.

lerno:
I would make `add` a macro.

lerno:
Look at the CI error, it looks like there might be something wrong with the ordering.

OdnetninI:
Wait, that is a weird bug, it only triggered in MSVC Debug mode.

Checking the source of the assert: https://llvm.org/doxygen/Instructions_8h_source.html
It does not make sense that it is not trigger in Linux or Mac. So why only MSVC Debug???

I will look into it more.

OdnetninI:
Found the documentation, updating...

https://llvm.org/docs/LangRef.html#cmpxchg-instruction

lerno:
I'm going to push a change that ensures there's a compile time check there. Also, some of the other compile time checks were broken for atomic store/load, so those are updated as well. I'm also changing MONOTONIC to RELAXED, following the more common naming.

lerno:
MSVC debug is the only one running LLVM with full debug info and asserts. So that one triggered an assert in LLVM because of invalid lowering.

lerno:
The update to RELAXED is in now.

OdnetninI:
Let me change it to RELAXED.

lerno:
Thank you!

PR #963, opened 2023-08-31
lib/std/collections: fix comments, List.remove_if() and List.retain_i…

pierrec:
…f() methods


lerno:
Hmm... what was the problem with the old implementation?

pierrec:
I added tests and they didnt pass. Looking at the implementation I couldnt see how it would work.

PR #968, opened 2023-09-01
std/collections: use List.remove_if and List.remove_using_test for Li…

pierrec:
…st.remove and List.compact

### DIFF
--- a/lib/std/collections/list.c3
+++ b/lib/std/collections/list.c3
@@ -415,17 +415,7 @@ fn bool List.contains(&self, Type value) @if(ELEMENT_IS_EQUATABLE)
  **/
 fn usz List.remove(&self, Type value) @if(ELEMENT_IS_EQUATABLE)
 {
-	usz size = self.size;
-	for (usz i = size; i > 0; i--)
-	{
-		if (!equals(self.entries[i - 1], value)) continue;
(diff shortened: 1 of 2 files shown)


lerno:
This would create another function call for every element checked though?

pierrec:
Good point. Adding @inline to the lambda should work, right?

> return self.remove_if(fn bool(p) @inline => *p == null);

lerno:
Can you benchmark using various ways including the first?Sent from my iPhone1 sep. 2023 kl. 18:58 skrev Pierre Curto ***@***.***>:﻿
Good point. Adding @inline to the lambda should work, right?

return self.remove_if(fn bool(p) @inline => *p == null);

pierrec:
```
module bench;
import std::collections::list;

def PtrList = List(<void*>);

fn void main()
{
	const N = 40_000;
	PtrList test;
	for (usz i = 0; i < N; i++)
	{
		test.add_array({ null, &test });
		test.compact();
	}
}
```

Initial method:
./bench  0.63s user 0.01s system 98% cpu 0.648 total

Proposed one:
./bench  3.49s user 0.02s system 99% cpu 3.523 total

So... forget this PR!

lerno:
That seems surprisingly big a difference. Is this really -O3 --fast?

PR #970, opened 2023-09-02
make io.Stream read_byte use read if possible

pierrec:


lerno:
I am making the overall changes to stream now, so this change will likely conflict with that one. Let's look at this one later.

lerno:
Dot should be "precision", right? That's how it works for regular printf I think it's just that "precision" in the string case means string length.

pierrec:
> Dot should be "precision", right? That's how it works for regular printf I think it's just that "precision" in the string case means string length.

oh yes...

lerno:
Thanks for fixing this.

PR #977, opened 2023-09-06
lib/std/collections: add Bitmap

pierrec:


lerno:
"Bitmap" seems like a bad name. Isn't this a packed bool list? Bitmap is a name intimately associated with graphics.

pierrec:
True, it is used in different context. Need to come up with something else.

lerno:
BitList, BoolPack, PackedBoolList, BoolList

lerno:
Looking at other implementations, BitSet is usually fixed size, so I propose:
BitSet - fixed size
GrowableBitSet - current implementation (or DynamicBitSet or BitList)

lerno:
Thank you

PR #980, opened 2023-09-08
lib/std/encoding: remove use of pushback_byte in json

pierrec:


lerno:
Thank you for that improvement.

PR #982, opened 2023-09-08
fix Object.free

pierrec:


lib/std/collections/object.c3
```diff
@@ -199,22 +194,22 @@ fn void Object.set_object(&self, String key, Object* new_object) @private
 	defer
 	{
 		(void)free(entry.key, .using = self.allocator);
-		(void)entry.value.free();
+		(void)free(entry.value, .using = self.allocator);
```
lerno, line 197: This doesn't look right.

pierrec: Oups I missed that one, I noticed it later but only fixed the first instance.

lib/std/ascii.c3
```diff
@@ -53,6 +53,10 @@ fn bool char.is_blank(char c) => is_blank_m(c);
 fn bool char.is_cntrl(char c) => is_cntrl_m(c);
 fn char char.to_lower(char c) => (char)to_lower_m(c);
 fn char char.to_upper(char c) => (char)to_upper_m(c);
+/**
+ * @require c.is_xdigit()
+ **/
+fn char char.from_hex(char c) => c.is_digit() ? c - '0' : 10 + c - (c < 'a' ? 'A' : 'a');
```
lerno, line 59: If you know it to be a hex digit, then `c.is_digit() ? c - '0' : 10 + (c | 0x20) - 'a'` is one branch less.

lib/std/collections/object.c3
```diff
@@ -43,23 +43,23 @@ fn usz! Object.to_format(&self, Formatter* formatter) @dynamic
 			usz n = formatter.printf("[")!;
 			foreach (i, ol : self.array)
 			{
-				n += formatter.printf(i == 0 ? " " : ", ")!;
+				if (i > 0 )n += formatter.printf(",")!;
```
lerno, line 46: Missing space

lib/std/collections/object.c3
```diff
@@ -43,23 +43,23 @@ fn usz! Object.to_format(&self, Formatter* formatter) @dynamic
...
 			return n;
 		case ObjectInternalMap:
 			usz n = formatter.printf("{")!;
 			@pool()
 			{
 				foreach (i, key : self.map.key_tlist())
 				{
-					n += formatter.printf(i == 0 ? " " : ", ")!;
-					n += formatter.printf(`"%s": `, key)!;
+					if (i > 0 )n += formatter.printf(",")!;
```
lerno, line 57: Missing space between `)` and `n`

lib/std/io/stream/scanner.c3
```diff
@@ -2,7 +2,8 @@ module std::io;
 
 struct Scanner
 {
-	Stream* stream;
+	inline Stream stream;
```
lerno, line 5: I didn't know Scanner was supposed to be a stream?

lib/std/io/stream/scanner.c3
```diff
@@ -94,4 +100,25 @@ macro usz! Scanner.refill(&self, buf) @private
 			return err?;
 	}
 	return n;
+}
+
+fn usz! Scanner.read(&self, char[] bytes)
```
lerno, line 105: Can you add tests for these functions as well?

lib/std/io/stream/scanner.c3
pierrec: Yeah true but I found it can help in some cases where you dont want to flush and keep going reading. If you think this is too narrow case, it can be dropped.

lerno: Wouldn't you simply continue reading from the stream then?

pierrec: No because I need what's left in the scanner.

lerno:
Sorry for taking me so long to look at this. Thank you for your contribution.

PR #989, opened 2023-09-13
sema_asm: Do not initialize asm multiple times

OdnetninI:
`init_asm` is called on each asm block. This is not needed.
At some point, with enough asm blocks, asm_target:reg_register will be stuck in the infinite loop.


lerno:
It's a correct observation, it's wrong, but the problem is that `asm_target.initialized` is not set to true after the check in `init_asm`

lerno:
I've updated init_asm now so this fix should not be needed. Have a look!

OdnetninI:
Oh, I haven't notice that.
Yeah, it works now. Thank you :)

PR #992, opened 2023-09-14
Add `log(x, base)` macro to `std::math` module.

data-man:


lerno:
Rename `log` -> `ln` and use `log` for `logb`

lerno:
Thank you!

PR #994, opened 2023-09-15
Add `$$return_address` builtin.

data-man:
Also `$$frameaddress` renamed to `$$frame_address` and supports a parameter.


data-man:
Hmm...

lerno:
Keep the $$frameaddress name and use $$returnaddress and remove the argument as it will almost never work (according to docs)

data-man:
It was a bad idea.
`$$frameaddress(0)` == `$$returnaddress(0)` in my tests.

PR #1006, opened 2023-09-19
lib::std::encoding: add varint::{read,write}

pierrec:


data-man:
Which varint? :)

https://github.com/mattsta/varint

pierrec:
Ah. I was only aware of [Go's](https://pkg.go.dev/encoding/binary@go1.21.1#ReadUvarint).

lerno:
Perhaps better with `read_network_int($Type, char bytes)` and corresponding write methods. What they will do is then to use the corresponding amount of bytes to read/write the integer.

This often done in protocols, like:
```c
uint elements = foo.read_network_int(uint, 3);
List(<Blurb>) list;
for (uint i = 0; i < elements; i++)
{
   list.push(blurb_read(foo));
}
```

...
```c
bar.write_network_int(list.len(), 3);
foreach (&blurb : list)
{
  blurb_write(foo, blurb);
}
```

lerno:
Oh, I see, but still you might want to make stream methods for it simply?

data-man:
I think `std::serialization` is the best way.

pierrec:
@lerno gentle ping.

lerno:
Thank you!

PR #1010, opened 2023-09-22
std::lib::os: add get_home_dir() and get_config_dir()

pierrec:


lib/std/os/env.c3
```diff
@@ -63,4 +81,50 @@ fn String! executable_path(Allocator *using = mem::heap())
...
+	$endif
+	return get_var(home);
+}
+
+/**
+ * Returns the current user's config directory.
+ **/
+fn String! get_config_dir(Allocator* using = mem::heap())
+{
+	$if env::WIN32:
```
lerno, line 105: Use "Path" for this

lerno: No, I meant to use Path to append paths together. I think your code is also incorrect, forgetting to add '/' in some cases. Path ensures valid paths and can concatenate into new paths.

pierrec: I have already fixed the missing / for Linux&co.
I dont think Path is required as the concatenation occurs when it is known that this is a non Windows code path.

lib/std/os/env.c3
```diff
@@ -39,6 +63,53 @@ fn void set_var(String name, String value, bool overwrite = true)
 	$endif
 }
 
+/**
+ * Returns the current user's home directory.
+ **/
+fn String! get_home_dir()
```
lerno, line 69: What is the ownership and allocation origin of String here?

lib/std/os/env.c3
```diff
@@ -39,6 +63,53 @@ fn void set_var(String name, String value, bool overwrite = true)
...
+			else
+			{
+				s = get_var("HOME")!;
+			}
+			const DIR = "/.config";
+		$endif
+		str.init(s.len + DIR.len, using);
+		str.append(s);
+		str.append(DIR);
+		return str.as_str();
```
lerno, line 108: This is returning string as a String, but if you try to release this String then that will fail, because the string is part of a DString. `as_str()` is getting the view. Maybe we should rename all of these `.str_view()` to clarify it.

lib/std/os/env.c3
pierrec: Good question. I was wondering the same about get_var(). I think an allocator may be required for those functions.

lib/std/os/env.c3
pierrec: It makes sense as there is already .array_view(). Maybe this should be done before this PR?

lerno: I've made the update.

lib/std/os/env.c3
lerno: get_var() is grabbing it from the os which then might change it, but there is no need to deallocate it.

pierrec: > get_var() is grabbing it from the os which then might change it, but there is no need to deallocate it.

It currently does for non-WIndows os'es but with WIN32 you need to allocate it.

lerno: Best is probably to make the non-Win32 also allocate.

lib/std/os/env.c3
```diff
@@ -8,17 +8,38 @@ import libc;
...
+		DString buffer;
+		buffer.init(N, using);
+		usz n = get_env_win32(name, buffer);
+		if (n == 0) return SearchResult.MISSING?;
+		if (n > N)
+		{
+			buffer.reserve(n - N);
+			get_env_win32(name, buffer);
+		}
+		return buffer.str_view()[:n];
```
lerno, line 31: Still buggy, it will not be possible to release this string. You want to allocate the buffer with a simple tmalloc, create a string from it and copy it with using.

lib/std/os/env.c3
```diff
@@ -28,38 +49,98 @@ fn String! get_var(String name)
...
+			else
+			{
+				s = get_var("HOME")!;
+			}
+			const DIR = "/.config";
+		$endif
+		str.init(s.len + DIR.len, using);
+		str.append(s);
+		str.append(DIR);
+		return str.str_view();
```
lerno, line 114: Returns the view of a dstring, so cannot be released.

lib/std/os/win32/process.c3
```diff
@@ -83,7 +83,15 @@ extern fn Win32_BOOL getOverlappedResult(
 	Win32_LPDWORD      lpNumberOfBytesTransferred,
 	Win32_BOOL         bWait
 ) @extern("GetOverlappedResult");
-
+extern fn Win32_DWORD getEnvironmentVariable(
```
lerno, line 86: Rename: win32_GetEnvironmentVariableA

lib/std/os/win32/process.c3
```diff
@@ -83,7 +83,15 @@ extern fn Win32_BOOL getOverlappedResult(
...
+	Win32_LPCSTR lpName,
+	Win32_LPSTR lpBuffer,
+	Win32_DWORD nSize
+) @extern("GetEnvironmentVariableA");
+extern fn Win32_BOOL setEnvironmentVariable(
+	Win32_LPCSTR lpName,
+	Win32_LPCSTR lpBuffer
+) @extern("SetEnvironmentVariableA");
```
lerno, line 94: rename to win32_SetEnvironmentVariableA

lib/std/os/win32/process.c3
```diff
@@ -83,7 +83,15 @@ extern fn Win32_BOOL getOverlappedResult(
 	Win32_LPDWORD      lpNumberOfBytesTransferred,
 	Win32_BOOL         bWait
 ) @extern("GetOverlappedResult");
-
+extern fn Win32_DWORD getEnvironmentVariable(
+	Win32_LPCSTR lpName,
+	Win32_LPSTR lpBuffer,
+	Win32_DWORD nSize
+) @extern("GetEnvironmentVariableA");
```
lerno, line 90: Don't use the A functions, since they are not UTF8 safe. Use the W variants and do the string conversion.

lerno:
Let me see if I can make a version of this.

lib/std/os/env.c3
pierrec: Then:

> return buffer.str_view()[:n].copy(using);

lib/std/os/win32/process.c3
pierrec: I did try with GetEnvironmentVariable but then it couldnt be found.

pierrec:
> Let me see if I can make a version of this.

Go for it :).

lerno:
set / get / clear env should probably be impoved by returning an optional on error.

pierrec:
Thanks for this and sorry for the long back and forth.

lerno:
No not at all.

lerno:
I think it resulted in good improvements for the stdlib.

PR #1016, opened 2023-09-25
get_env for Win32, @pierrec's get_config_dir and get_home_dir

lerno:


lerno:
This should include all features of #1010

PR #1026, opened 2023-10-03
add DString.insert_at

pierrec:


test/unit/stdlib/conv_tests.c3
```diff
@@ -4,7 +4,7 @@ import std::io;
 fn void! comparison_helper_32_to_8(Char32 c32, String expected_output)
 {
 	char[8] out;
-	usz len = conv::char32_to_utf8(c32, &out, 4)!;
+	usz len = conv::char32_to_utf8(c32, out[:4])!;
```
lerno, line 7: Can be `&out` actually.

lib/std/core/dstring.c3
```diff
@@ -175,8 +153,8 @@ fn DString DString.copy(self, Allocator* using = null)
 		if (using) return new_with_capacity(0, using);
 		return (DString)null;
 	}
-	if (!using) using = mem::heap();
 	StringData* data = self.data();
+	if (!using) using = data.allocator;
```
lerno, line 157: The correct behaviour is to use mem::heap()

PR #1031, opened 2023-10-05
std::core: rewrite DString

pierrec:
This PR aims at making the use of DString easier. More precisely, now a DString can be created and built and the resulting string from str_view() or zstr_view() can be freed without having to keep a reference to the original DString.

### DIFF
--- a/lib/std/core/dstring.c3
+++ b/lib/std/core/dstring.c3
@@ -2,6 +2,16 @@ module std::core::dstring;
 
 def DString = distinct void*;
 
+struct StringData @private
+{
+	Allocator* allocator;
+	usz len;
+	usz capacity;
+}
+
+macro char* StringData.start(&data) => (char*)((void*)data - data.capacity);
+macro char* StringData.end(&data) => (char*)((void*)data - data.capacity + data.len);
+
 const usz MIN_CAPACITY @private = 16;
 
 /**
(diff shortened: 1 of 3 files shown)


lib/std/core/dstring.c3
```diff
@@ -10,37 +20,43 @@ const usz MIN_CAPACITY @private = 16;
 fn void DString.init(&self, usz capacity = MIN_CAPACITY, Allocator* using = mem::heap())
 {
 	if (capacity < MIN_CAPACITY) capacity = MIN_CAPACITY;
-	StringData* data = malloc(StringData, 1, .using = using, .end_padding = capacity);
-	data.allocator = using;
-	data.len = 0;
-	data.capacity = capacity;
-	*self = (DString)data;
+	char* start = malloc(char, capacity + StringData.sizeof, .using = using);
```
lerno, line 23: This doesn't guarantee alignment.

lerno:
Let me look at rewriting it, as there are some gnarly things to consider. Keep the conv.c3 update though. If you have time, look at converting Streamable to a protocol.

lerno:
Uh, I also realize that it won't work.

pierrec:
> Let me look at rewriting it, as there are some gnarly things to consider. Keep the conv.c3 update though. If you have time, look at converting Streamable to a protocol.

Will have a look.

Thanks for considering this approach!

PR #1033, opened 2023-10-06
misc

pierrec:
Changes to conv::char32_to_utf8_unsafe(); add tests for DString
add tests for DString

### DIFF
--- a/lib/std/core/dstring.c3
+++ b/lib/std/core/dstring.c3
@@ -136,8 +136,7 @@ fn void DString.append_char32(&self, Char32 c)
 {
 	char[4] buffer @noinit;
 	char* p = &buffer;
-	conv::char32_to_utf8_unsafe(c, &p);
-	usz n = p - (char*)&buffer;
+	usz n = conv::char32_to_utf8_unsafe(c, &p);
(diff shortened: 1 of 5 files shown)


lib/std/core/dstring.c3
```diff
@@ -308,18 +308,18 @@ fn void DString.insert_at(&self, usz index, String s)
 	data.len += s.len;
 
 	char* start = data.chars[index:s.len].ptr; // area to insert into
-	mem::move(start + s.len, start, len - index); // move existing data
+	mem::move((void*)start + s.len, start, len - index); // move existing data
```
lerno, line 311: Why the void casts?

pierrec: When working on DString with the header at the tail, I had miscalculations if I didnt cast.

lerno: That... should never be. Do you have a repro? `void*` and `char*` works the same way.

pierrec: Wait. I think it was a StringData*... I'll remove the void*.

lerno:
Thank you.

PR #1038, opened 2023-10-08
List capacity and SubProcess field name change

pierrec:


lib/std/os/subprocess.c3
```diff
@@ -31,7 +31,7 @@ struct SubProcess
 	Pid_t child @if(!env::WIN32);
 	int return_status @if(!env::WIN32);
 
-	bool is_alive;
+	bool alive;
```
lerno, line 34: I prefer is_alive, as it reads better: `self.is_alive` as opposed to `self.alive`

pierrec: The problem is that it clashes with the is_alive function.

lerno: Hmm.. let's change the function name instead.

lib/std/collections/list.c3
```diff
@@ -352,11 +352,11 @@ fn void List.set(&self, usz index, Type value) @operator([]=)
 fn void List.ensure_capacity(&self, usz added = 1) @inline @private
 {
 	usz new_size = self.size + added;
-	if (self.capacity > new_size) return;
+	if (self.capacity >= new_size) return;
 
 	assert(new_size < usz.max / 2U);
-	usz new_capacity = self.capacity ? 2U * self.capacity : 16U;
-	while (new_size >= new_capacity) new_capacity *= 2U;
+	usz new_capacity = self.capacity ? 2 * self.capacity : 16;
```
lerno, line 358: A good rule is to use unsigned constants with unsigned values. In *this* case it works, but it could just as well not work.

lerno:
Thank you!

PR #1047, opened 2023-10-15
Add more paths to search for the standard library

pinicarus:
Module `std` is searched first in a `c3` subdirectory. Search directories are, in order and relative to the compiler executable location:
- `lib/c3` relative to the parent directory
- `lib` relative to the parent directory
- `/lib/c3`
- `/lib`
- `/c3`
- `/`
- `c3` relative to the parent directory
- the parent directory
- `lib/c3` relative to the grand-parent directory
- `lib` relative to the grand-parent directory


lerno:
Thank you.

PR #1074, opened 2023-11-19
feat(lib/std): get backtrace for static binaries on linux

pinicarus:
*all elements below where tested both with glibc 2.37 and musl 1.2.3)*

When build a static executable, `dladdr` is a stub that always returns 0, which then gives this kind of backtrace:
```
ERROR: 'hello'
  in ???
  in ???
  in ???
  in ???
```

This patch fallbacks on checking the given address against the ELF executable file to get location information also for static executable:
```
ERROR: 'hello'
  in hello.test2 (/tmp/hello.c3:6) [/tmp/hello]
  in hello.test1 (/tmp/hello.c3:10) [/tmp/hello]
  in hello.main (/tmp/hello.c3:15) [/tmp/hello]
  in main (/usr/lib/c3/std/core/private/main_stub.c3:19) [/tmp/hello]
```

However, there is still a limitation when building a static PIE executable because the actual address is different from the mapping in the ELF file:
```
ERROR: 'hello'
  in ??? (source unavailable) [/tmp/hello]
  in ??? (source unavailable) [/tmp/hello]
  in ??? (source unavailable) [/tmp/hello]
  in ??? (source unavailable) [/tmp/hello]
```
Maybe this can be circumvented by retro-mapping the segment from `/proc/[pid]/maps` the address is within into the segment from the ELF file, but I'm not sure if it'd work and if it does, it'd be a harder and longer PR.

### DIFF
--- a/lib/std/os/linux/linux.c3
+++ b/lib/std/os/linux/linux.c3
@@ -131,45 +131,72 @@ fn ulong! elf_module_image_base(String path) @local
 	return 0;
 }
 
+fn Backtrace! backtrace_load_from_exec(void* addr, Allocator* allocator) @local
+{
+	char[] buf = mem::temp_array(char, 1024);
+
+	String exec_path = process::execute_stdout_to_buffer(buf, {"realpath", "-e", string::tformat("/proc/%d/exe", posix::getpid())})!;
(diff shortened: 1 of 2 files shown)


lerno:
Thank you!

Issue #779, opened 2023-06-13
c3c compilation fails on Fedora 38


Ashvith10:
Here's the log:
```sh
$ cmake ..
C3C version: 0.4.527
-- Found LLVM 16.0.5
-- Using LLVMConfig.cmake in: /usr/lib64/cmake/llvm
-- Libraries located in: /usr/lib64
-- using find_library
-- linking to llvm libs LLD_COFF-NOTFOUND;LLD_COMMON-NOTFOUND;LLD_WASM-NOTFOUND;LLD_MINGW-NOTFOUND;LLD_ELF-NOTFOUND;LLD_MACHO-NOTFOUND
-- Found lld libs LLD_COFF-NOTFOUND;LLD_COMMON-NOTFOUND;LLD_WASM-NOTFOUND;LLD_MINGW-NOTFOUND;LLD_ELF-NOTFOUND;LLD_MACHO-NOTFOUND
-- using gcc/clang warning switches
-- The following OPTIONAL packages have been found:

 * CURL
 * FFI
 * Terminfo
 * ZLIB

-- The following REQUIRED packages have been found:

 * LLVM (required version >= 16)

-- Configuring done (0.2s)
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
LLD_COFF
    linked by target "c3c" in directory /home/<user>/Projects/c3c
    linked by target "c3c_wrappers" in directory /home/<user>/Projects/c3c
LLD_COMMON
    linked by target "c3c" in directory /home/<user>/Projects/c3c
    linked by target "c3c_wrappers" in directory /home/<user>/Projects/c3c
LLD_ELF
    linked by target "c3c" in directory /home/<user>/Projects/c3c
    linked by target "c3c_wrappers" in directory /home/<user>/Projects/c3c
LLD_MACHO
    linked by target "c3c" in directory /home/<user>/Projects/c3c
    linked by target "c3c_wrappers" in directory /home/<user>/Projects/c3c
LLD_MINGW
    linked by target "c3c" in directory /home/<user>/Projects/c3c
    linked by target "c3c_wrappers" in directory /home/<user>/Projects/c3c
LLD_WASM
    linked by target "c3c" in directory /home/<user>/Projects/c3c
    linked by target "c3c_wrappers" in directory /home/<user>/Projects/c3c

-- Generating done (0.0s)
CMake Generate step failed.  Build files cannot be regenerated correctly.
```

[...]

PR #1077, opened 2023-11-20
Add CMake option to link c3 dynamically to LLVM/LLD

Wiguwbe:
To link dynamically, use
`cmake -DC3_LINK_DYNAMIC=ON <other-options> ../`

As discussed in #779


Wiguwbe:
To note, not tried on Windows, it would probably need the `*.dll` besides the `*.so` (and the same for Mac?)

lerno:
Thank you!

PR #1130, opened 2024-02-13
Fixing row major for some matrix functions

tonis2:
This pull request fixes my own mess, I made some time ago, cause I did not understand how row major / column major works on matrixes.

Basically all Matrixes should follow a same direction while doing calculations, otherwise it will break the math.

And then when sending the matrix data to GPU, the Matrixes should be transposed with .transpose() function, cause glsl shaders at-least read the values in column-major.

For short, all matrixes in the commit are converted from column-major to row-major, and tested that they work this way.


lerno:
Could you rebase on the latest?

lerno:
Wait, I probably need to implement those by hand anyway, since we can't break compatibility with 0.5.x which removing some of those constants would.

lerno:
No unit tests?

lerno:
I'm going to close this one for now. Tests have to be added for 0.5.5

PR #1148, opened 2024-02-19
Change matrix function to row major, add tests for them

tonis2:
Overview what this pull request contains

* Change some matrix functions to use row-major layout, cause previously (cause my own mistake) some of them were in column-major

* New `roundf` macro, it's very useful when writing tests for matrixes for example, cause all the values have lots of decimal places.

* New `radians` macro

* Add identity constant for Quaternion

### DIFF
--- a/lib/std/math/math.c3
+++ b/lib/std/math/math.c3
@@ -392,14 +392,14 @@ macro nearbyint(x) => $$nearbyint(x);
 
 /**
  * @require values::@is_promotable_to_floatlike(x) `The input must be a number or a float vector`
- * @require $assignable(exp, $typeof(x)) || values::@is_int(exp) `The input must be an integer, castable to the type of x`
+ * @require $assignable(exp, $typeof(values::promote_int(x))) || values::@is_int(exp) `The input must be an integer, castable to the type of x`
  **/
 macro pow(x, exp)
 {
 	$if types::is_floatlike($typeof(exp)):
-		return $$pow(x, ($typeof(x))exp);
+		return $$pow(values::promote_int(x), ($typeof(values::promote_int(x)))exp);
 	$else
-		return $$pow_int(x, exp);
+		return $$pow_int(values::promote_int(x), exp);
 	$endif
 }
 
--- a/releasenotes.md
+++ b/releasenotes.md
@@ -7,6 +7,7 @@
 
 ### Fixes
 - Struct/union members now correctly rejects members without storage size #1147.
+- `math::pow` will now correctly promote integer arguments.
 
 ### Stdlib changes
 None


lib/std/math/math.c3
```diff
@@ -97,8 +97,6 @@ def complex_identity = complex::identity(<double>);
 
 def Quaternionf = Quaternion(<float>);
 def Quaternion = Quaternion(<double>);
-def quaternionf_identity = quaternion::identity(<float>);
```
lerno, line 100: This should not be removed until 0.6.0. At most add a `@deprecated`

lib/std/math/math.c3
```diff
@@ -120,6 +118,14 @@ def MATRIX3F_IDENTITY = matrix::IDENTITY3(<float>);
 def MATRIX4_IDENTITY = matrix::IDENTITY4(<double>);
 def MATRIX4F_IDENTITY = matrix::IDENTITY4(<float>);
 
+
+/**
+ * @require types::is_numerical($typeof(x)) `The input must be a numerical value or numerical vector`
+ **/
+macro radians(x) {
+    return x * PI / 180;
```
lerno, line 126: `deg_to_rad` as name perhaps?

lib/std/math/math_matrix.c3
```diff
@@ -9,7 +9,7 @@ struct Matrix2x2
 			Real m00, m01;
 			Real m10, m11;
 		}
-		Real[4] m;
+		Real[<4>] m;
```
lerno, line 12: Note that this changes the alignment of the matrix. This might be ok, but it should be mentioned.

tonis2: Glad you mentioned, tested it and my vulkan renderer started giving below error cause of this change

```c
Fatal glibc error: malloc assertion failure in sysmalloc: (old_top == initial_top (av) && old_size == 0) || ((unsigned long) (old_size) >= MINSIZE && prev_inuse (old_top) && ((unsigned long) old_end & (pagesize - 1)) == 0)
```

Only reason I made this change was, so I could use `math::round_to_decimals(matrix.m, 4)`
I guess, I'll have to use good old for loop, to make the rounding.

lerno: You can do a cast: `math::round_to_decimals((Real[<3>])matrix.m, 4)`

lerno:
Can you add the same tests but for the float variants? So there aren't some problems with conversions with that one.

tonis2:
@lerno 
Do you mean adding tests like below, so both double and float are tested ?

```c
assert(math::round_to_decimals((double[<16>])result.m, 4) == math::round_to_decimals((double[<16>])perspective.m, 4));

assert(math::round_to_decimals((float[<16>])result.m, 4) == math::round_to_decimals((float[<16>])perspective.m, 4));
```

data-man:
What if instead of adding a new function, add a new parameter to `math::ceil/floor/round/etc`?

lerno:
Yes, but I think especially you want to test both Matrix4 and Matrix4f etc.

lerno:
Can you rebase and squash this into a single commit?

PR #1152, opened 2024-02-24
Add support [rgba] properties in vectors

data-man:
Because colors are often needed.


lerno:
I did research for languages that natively support swizzling. Aside from shaders, they all only support only one scheme: `xyzw`. There are actually many different naming schemes that could be considered: rgba, uv, ijkl etc. The line has to be drawn somewhere, and just supporting `xyzw` is the simplest and most straightforward solution.

(Also, note that your code allows things like `foo.graywarz`)

data-man:
> I did research for languages that natively support swizzling.

It's been a while since I checked, but Odin has supported rgba in the past.

> they all only support only one scheme: xyzw

In coordinates, not colors.

data-man:
> Also, note that your code allows things like foo.graywarz

Fixed.

PR #1158, opened 2024-02-26
Remove numbering in --list* commands

data-man:


lerno:
Is this needed?

data-man:
Why not? When updating syntax highlighting, additional actions are required.

lerno:
I think better we remove the numbering completely for all that are numbered, I'll accept such a PR.

data-man:
> we remove the numbering completely

Done.

src/compiler/compiler.c
```diff
@@ -740,33 +740,32 @@ void print_syntax(BuildOptions *options)
...
-			printf("%2d %s\n", i + 1, attribute_list[i]);
+			printf("%s\n", attribute_list[i]);
 		}
 	}
 	if (options->print_builtins)
 	{
 		for (int i = 0; i < NUMBER_OF_BUILTINS; i++)
 		{
-			printf("%3d $$%s\n", i + 1, builtin_list[i]);
+			printf("%s\n", builtin_list[i]);
```
lerno, line 757: You removed too much here and also below.

src/compiler/compiler.c
```diff
@@ -740,33 +740,32 @@ void print_syntax(BuildOptions *options)
...
 		}
-		puts("---");
 		for (int i = 0; i < NUMBER_OF_BUILTIN_DEFINES; i++)
 		{
-			printf("%2d $$%s\n", i + 1, builtin_defines[i]);
+			printf("%s\n", builtin_defines[i]);
```
lerno, line 761: $$ is missing.

src/compiler/compiler.c
```diff
@@ -740,33 +740,32 @@ void print_syntax(BuildOptions *options)
...
 		}
 	}
 	if (options->print_type_properties)
 	{
 		for (int i = 0; i < NUMBER_OF_TYPE_PROPERTIES; i++)
 		{
-			printf("%2d .%s\n", i + 1, type_property_list[i]);
+			printf("%s\n", type_property_list[i]);
```
lerno, line 768: . is missing.

data-man: Because dot isn't a part of name.

lerno:
Can we put this in 0.6.0 (so dev) rather than 0.5.5 (master)?

PR #1171, opened 2024-03-17
Fix aarch64 return type ABI bug

davidgm94:


lerno:
Thank you for finding this!

Issue #940, opened 2023-08-21
Cmdline argument feature to list all files used for compilation


fvpfvpfvp:
**Proposal**
Given a file C:/path/to/test.c3:

```
import std::io;
fn int main(char[][] argv) {
	io::println("Hello World!");
	return 0;
}

```
Calling 'c3c compile test.c3 --print-input' will yield the following output:

```
# input-files-begin
C:/path/to/test.c3
C:/path/to/std/io.c3
...
# input-files-end
```

And so forth, in the same style as the --print-output command. So all input files specified on the cmdline, all imported module files, all $include'd files, etc.

**Relevance**
This is useful for external build systems, so they know they should rebuild when a file changes.

lerno:
Would it be fine to return it as json?

fvpfvpfvp:
Yes, sure!

lerno:
This is now in 0.5.6

PR #1175, opened 2024-03-25
Added print-input command line argument

Wrimo:
Ads command line argument print-input to print all files used for compilation. 

Resolves issue #940

### DIFF
--- a/releasenotes.md
+++ b/releasenotes.md
@@ -4,6 +4,7 @@
 
 ### Changes / improvements
 - Support `defer (catch err)`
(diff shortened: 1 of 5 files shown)


lerno:
I think the path is probably more correct than the name, as an example, the standard library has at least two "files.c3", but in different directories.

lerno:
Also, you should add the feature to the release notes for 0.5.6

Wrimo:
Alright, I've just made those changes.

lerno:
This looks good! Thank you.

PR #1214, opened 2024-06-30
add is_keyer macro

Andersama:
Adds a macro to try to detect a key-like structure.

TODO: key-like structures are any integral, bitset, boolean, float, double, tuple of keys or array of keys.

Also changes `is_comparer` to take unsigned integers.

### DIFF
--- a/lib/std/sort/sort.c3
+++ b/lib/std/sort/sort.c3
@@ -13,7 +13,7 @@ macro usz @len_from_list(&list)
 macro bool @is_comparer(#cmp, #list)
 {
 	var $Type = $typeof(#cmp);
-	$if $and($Type.kindof == FUNC, $Type.returns.kindof == SIGNED_INT):
+	$if $and($Type.kindof == FUNC, $or($Type.returns.kindof == SIGNED_INT, $Type.returns.kindof == UNSIGNED_INT)):
 		var $params = $Type.params;
 		$if $params.len != 2:
 			return false;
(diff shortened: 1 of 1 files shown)


lib/std/sort/sort.c3
```diff
@@ -13,7 +13,7 @@ macro usz @len_from_list(&list)
 macro bool @is_comparer(#cmp, #list)
 {
 	var $Type = $typeof(#cmp);
-	$if $and($Type.kindof == FUNC, $Type.returns.kindof == SIGNED_INT):
+	$if $and($Type.kindof == FUNC, $or($Type.returns.kindof == SIGNED_INT, $Type.returns.kindof == UNSIGNED_INT)):
```
lerno, line 16: This is incorrect, a comparator function should return -1, 0, 1. And unsigned int cannot ever return -1

lerno:
What should is_keyer do?

lib/std/sort/sort.c3
Andersama: Ah, ok, I didn't realize that the macro was testing for three way comparisons.

lib/std/sort/sort.c3
```diff
@@ -35,4 +35,44 @@ macro bool @is_comparer(#cmp, #list)
...
-}
\ No newline at end of file
+}
+
+macro bool @is_key(#item)
+{
+	var $key_type = $typeof(#item);
+	$switch
+		$case $key_type == @typeid(char):
+			return true;
```
lerno, line 45: Two alternatives:
1. Use kindof, or just `.is_int()`:  `$typeof(#item).is_int()` returns true for all integer types.
2. If you want to do a type switch, it easy:

```c
$switch $typeof(#item)
$case char: return true;
$case ushort: return true;
```

lerno:
Added as @is_cmp_key_fn in `dev`

PR #1216, opened 2024-07-01
Fix formatter dereferencing null ZString, print "(null)" for null pointers.

cbuttner:
Closer to printf-like behavior.


lerno:
Can I ask for a unit test too? I think we might detect other regressions too if we have that.

lerno:
I'll accept this and I add the test and release notes for it, but try to add that next time.

cbuttner:
Thanks, will keep in mind, although you could've just waited until I got around to it.

lerno:
Hot on the trigger you know 😄. I feel bad when pull requests aren't merged quickly.

PR #1217, opened 2024-07-01
Added a `--no-headers` option.

cbuttner:


lerno:
Good addition! Ideally we should add that as project setting too later on.

lerno:
I'll add the info to the release notes, but that should be kept in mind for later.

PR #1218, opened 2024-07-01
Implicitly cast distinct inline to index.

cbuttner:


src/compiler/sema_casts.c
```diff
@@ -611,6 +611,13 @@ bool cast_to_index(SemaContext *context, Expr *index)
 		case TYPE_ENUM:
 			type = type->decl->enums.type_info->type->canonical;
 			goto RETRY;
+		case TYPE_DISTINCT:
```
lerno, line 614: Even simpler: use "type_flat_distinct_inline" on the type.

cbuttner: Better?

lerno: Yes, thank you!

lerno:
Thank you

PR #1220, opened 2024-07-02
Add `$debugtrap` builtin.

cbuttner:


lerno:
Can we rename it simply $breakpoint and also have a simple codegen test for it?

cbuttner:
We could, but that's what the builtin is called in Clang, so I'm not sure?

Just noting that in the future there should be a `is_debugger_attached()` and `breakpoint_if_debugging()`, currently the program just exits (with SIGTRAP on Linux) if no debugger is attached.

Another option for a name would be `debug_break()`.

lerno:
I don't like tying myself to LLVM names. I think the first was debug_break in MSVC, which is also a good name. I think the macro should be named the same as the builtin. So either `$debugbreak` `$debug_break` or `$breakpoint`.

cbuttner:
Understandable.
`breakpoint` is used by C++20, Rust, Zig and Python.
`debug_break`/`debugbreak` is used by MSVC and some libraries.

Probably going with `breakpoint` then.

lerno:
Thank you.

PR #1224, opened 2024-07-03
Fix `$$unaligned_store` arg check and add test.

cbuttner:


lerno:
Aaah I changed the order of the arguments initially and then I only tested with an int 🤦  Thank you for fixing it.

PR #1225, opened 2024-07-04
add insertion sort

Andersama:
NOTE: I'm following along with quicksort as an example. Will be used as a fallback sort for other algorithms.


lerno:
Can you add unit tests too? You could use the tests for quicksort and apply them to this.

PR #1231, opened 2024-07-09
sort: make countingsort.c3 even more branchless

Andersama:
Track even more specific partitions, such that we can define loops without if-statements within them.

If there's a branchless-ish way to do a 3 way partition of the provided indexs that'd be perfect. Here we're using a partition algorithm to split which indexs use which fallback.

May be better to iterate over the counts a second time?


lerno:
This push failed because the code doesn't compile. It really should have unit tests as well. Finally, can I ask you to follow the coding standard used in the rest of the stdlib so it's consistent: space between operators (`foo+bar` => `foo + bar`), space after `,` and `{` on new line if it's not a lambda definition?

Andersama:
I think I'll discard this one, because I'm fairly confident iterating over counts a second time is in fact better, and also I obviously failed to write this one, but you can get the picture.

PR #1233, opened 2024-07-09
Make countingsort.c3's recursion stage branchless

Andersama:
Tracks the three potential cases for each fallback, item counts ranging from [2,32], [33,128], [128, ...] and uses a loop specifically for each fallback.


lerno:
Do you still want to add this?

Andersama:
Yeah, this one is worth doing.

lerno:
It says "draft". Does it mean it's not ready to be merged?

Andersama:
Oh, that's what you meant.

lerno:
Thank you!

PR #1235, opened 2024-07-11
`@nopadding` and `@compact` attributes

cbuttner:
`@nopadding`
If applied to a struct/union, checks for any padding in the struct. If applied to a member, checks if there is padding before the member.

`@compact`
If applied to a struct/union: Same as `@nopadding`, but extends the check to all members of type struct/union recursively.


cbuttner:
Will add some tests later.

src/compiler/compiler_internal.h
```diff
@@ -427,8 +427,9 @@ typedef struct
 typedef struct
 {
 	TypeSize size;
-	Decl **members;
 	MemberIndex union_rep;
+	Decl **members;
+	Decl *padded_decl;
```
lerno, line 432: Reduce the memory footprint of this by using a DeclId instead?

src/compiler/sema_decls.c
```diff
@@ -269,6 +273,23 @@ static inline bool sema_analyse_struct_member(SemaContext *context, Decl *parent
...
+{
+	if (member_type->type_kind == TYPE_STRUCT || member_type->type_kind == TYPE_UNION) 
+	{
+		if (member_type->decl->strukt.padded_decl) 
+		{
+			if (!decl->strukt.padded_decl) decl->strukt.padded_decl = member_type->decl->strukt.padded_decl;
+			if (decl->attr_compact) 
+			{
+				sema_error_at(context, member->span, "This member has holes.");
+				RETURN_SEMA_ERROR(member_type->decl->strukt.padded_decl, "Padding would be added for this.");
```
lerno, line 286: RETURN_SEMA_ERROR will print a sema_error_at, so so what you want to do instead is probably using SEMA_NOTE:

```
Decl *padded = member_type->decl->strukt.padded_decl; 
if (padded->name) 
{ 
   SEMA_ERROR(member, "This member is not compact, padding would be added for '%s'", padded->name);
}
else
{
   SEMA_ERROR(member, "This member is not compact.");
}
SEMA_NOTE(padded, "Padding would be added for this.");
return false;
```

lerno: That is why the SEMA_NOTE will show that definition.

cbuttner: So why print the name, if it's unhelpful and redundant.

lerno: Oh, it was just a suggestion, feel free to ignore it.

lerno:
Let me accept this as-is and I might tweak it a bit later.

PR #1240, opened 2024-07-15
Make countingsort.c3's recursion stage branchless

lerno:
Tracks the three potential cases for each fallback, item counts ranging from [2,32], [33,128], [128, ...] and uses a loop specifically for each fallback.


Andersama:
I'm an idiot, I don't know why the tests showed it was all good on the draft, I typed the result for count wrong.

PR #1241, opened 2024-07-15
Correct typing of count and use usz over ulong

Andersama:


lerno:
This still seems to be failing?

Andersama:
I didn't pay attention to the first failure location, I thought it was in the recursion stage, but on second viewing that didn't make sense.
The apparent error is on line 112: that also doesn't make sense.

Andersama:
#1242 is this, but without the junk commits

PR #1242, opened 2024-07-16
Use usz and fix out of bounds access in branchless loop

Andersama:


lerno:
Is this the final version now?

PR #1249, opened 2024-07-23
Add type property `is_substruct`.

cbuttner:
This is needed to determine if a struct is a a substruct at compile time.


lerno:
Thank you.

PR #1268, opened 2024-07-29
Add Riscv Example

chuckb:
I took a guess as to how you might organize these kinds of examples. Also, one thing I could add to this example is the ability to cleanly exit the qemu host via semihosting (I have an example working in another project). That way, this could be integrated with CI as a multi-arch test suite.


lerno:
Can we get this into the CI in some way? Like during build on linux it downloads qemu and runs the example?

chuckb:
Yep. That can be done - next few days...I'll get back to you.

chuckb:
So, not being too familiar with github ci...it looks like starting at line 219 of .github/workflows/main.yml I'd add dependencies for qemu and the risc-v toolchain. Then I'd add a runner entry somewhere after line 274 for `make run` within my project, assuming that my app kills qemu when Hello World is printed and then presumably checks for that string on the console?

So I've forked your project...can I somehow test my changes to CI before another PR to you?

lerno:
At the top of the CI there is this:
```
on:
  push:
    branches: [ master, dev, ci_testing, experiments ]
  pull_request:
    branches: [ master, dev ]
```
So if you rename your branch "ci_testing" then the CI ought to immediately kick in when you push.

lerno:
Can you place the CI together with the rest of the CI? Right now it's a separate yml file.

chuckb:
Yes, I will. I am trying just to strip this down to just what I need to get my piece working, then I'll add my additions to your master file. But I am stuck on a Risc-v ISA version mismatch at the moment and researching it. If you can shed light, that would be great.

chuckb:
Ok, I have the CI merged and my addition working, altho testproject is failing...I don't think that was me ;-). Sorry for my messy commits. I'd like to do some more examples...maybe using opensbi and 64-bit. Or perhaps other archs. I like this project.

lerno:
Ok, I'll grab this onto my machine and see if I can figure why the testproject fails. Thanks for all the hard work so far.

lerno:
Some broken CI, but this is now due to LLVM making a bad release (not the first time)

lerno:
Thank you!

PR #1270, opened 2024-07-30
Fix docker build issue in ubuntu kinetic

MashyBasker:
The official ubuntu image in DockerHub does not contain `22.10` aka `Kinetic`. This merge request changes the `UBUNTU_VERSION` to `22.04` aka `Jammy`.

Changelog:

- Update `UBUNTU_VERSION` to `22.04` to build on Jammy
- Update `LLVM_VERSION` from 15 to 17 due to this [line](https://github.com/c3lang/c3c/blob/8d6dabf65c6df420b9030bc049f5ce3092cec01c/CMakeLists.txt#L84C44-L84C46) in `CmakeLists.txt`
- Add an `UBUNTU_CODENAME` variable which is used to pull LLVM for a specific Ubuntu version

### DIFF
--- a/build-with-docker.sh
+++ b/build-with-docker.sh
@@ -26,13 +26,21 @@ else
     CMAKE_BUILD_TYPE="$config"
 fi
 
-UBUNTU_VERSION="22.10"
-LLVM_VERSION="15"
+UBUNTU_VERSION="22.04"
+UBUNTU_CODENAME=$( \
+    if [ "${UBUNTU_VERSION}" = "22.04" ]; then echo "jammy"; \
+    elif [ "${UBUNTU_VERSION}" = "20.04" ]; then echo "focal"; \
+    elif [ "${UBUNTU_VERSION}" = "18.04" ]; then echo "bionic"; \
+    else echo "unknown"; fi)
+
+LLVM_VERSION="17"
 
 IMAGE="$IMAGE:22"
 
 cd docker && $DOCKER build -t $IMAGE\
-    --build-arg DEPS="llvm-$LLVM_VERSION-dev liblld-$LLVM_VERSION-dev clang-$LLVM_VERSION libllvm$LLVM_VERSION llvm-$LLVM_VERSION-runtime" \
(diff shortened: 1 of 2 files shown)


lerno:
Thank you. Can we get this into the CI somehow?

I don't use this myself, so I can't judge if we're missing something essential with the removal of 22.10. @Its-Kenta did the previous updates, and so might have a better opinion on how it should work.

MashyBasker:
Would you like me to add a job to build the project in Ubuntu 22.04? I can start working on an MR right away or wait for more reviews, whichever you prefer.

lerno:
Getting a CI for it would be awesome.

lerno:
Something broke in the new ci:

> You have an error in your yaml syntax on line 503

MashyBasker:
Yeah, it was indentation problem. Sorry, about that

MashyBasker:
On second thought, the CI for Ubuntu 22 already exists in the job `build-linux`. Wouldn't my addition be redundant?

lerno:
I wanted the CI to run the docker script. That's what I meant.

lerno:
What is needed to make this work?

MashyBasker:
It seems we need to add a command to the script to create the build directory where CMakeOutput.log is supposed to be written. The other errors also appear to be caused by the absence of this build directory.

lerno:
So just a mkdir? Do you know the path?

MashyBasker:
This is the path expected by Cmake for the `CmakeOutput.log`:
` /home/c3c/source/build/CMakeFiles/CMakeOutput.log`

However, the `build-with-docker.sh` script already has the command `mkdir -p build bin` which creates these directories.

lerno:
So what does that mean? That is should work but doesn't?

MashyBasker:
I suspect the docker container is running as a non-root user, and the script is trying to write to a directory that the user doesn't have write permissions to. I adding a potential fix.

lerno:
Now we seem to be getting:
> chown: cannot access 'build': No such file or directory

MashyBasker:
I am adding the `chmod` and `chown` commands directly in the `build-with-docker.sh` script. This should provide the necessary permissions to write a file inside the `build` directory.

lerno:
Ok! Crossing my fingers here :D

MashyBasker:
Failed again. It must be a permission error because the `build` and `bin` directory gets created when I run the script locally.

lerno:
So, are there still options to try?

lerno:
#1347 should have merged these changes, if that merge looks good, please close this @MashyBasker

PR #1271, opened 2024-07-30
adding libc's termios to lib/std/libc

PalsFreniers:
I added a binding for the termios library that is a terminal IO control from libc


cbuttner (changes requested):
You should use `CInt` for `int` and `CUInt` for `uint`. `int` in C3 is always 4 bytes but in C it technically may not be.

lib/std/libc/termios.c3
```diff
@@ -0,0 +1,135 @@
+module libc::termios @if(env::LIBC);
+
+def Cc = char;
+def Speed = uint;
+def Tcflags = uint;
```
cbuttner, line 5: You may want to look into using `distinct` instead of `def` to add type safety where it makes sense, and changing the types of the corresponding constants to that type.

lib/std/libc/termios.c3
```diff
@@ -0,0 +1,135 @@
...
+const int IXOFF = 0010000;
+const int IMAXBEL = 0020000;
+const int IUTF8 = 0040000;
+
+extern fn int getAttr(int fd, Termios* self) @extern("tcgetattr");
+extern fn int setAttr(int fd, int optional_actions, Termios* self) @extern("tcsetattr");
+extern fn int sendBreak(int fd, int __duration) @extern("tcsendbreak");
+extern fn int drain(int fd) @extern("tcdrain");
+extern fn int flush(int fd, int queue_selector) @extern("tcflush");
+extern fn int flow(int fd, int action) @extern("tcflow");
```
cbuttner, line 119: You can use `Fd` (defined in std/libc/libc.c3) instead of `int` for the file descriptor type here.

cbuttner (changes requested):

PalsFreniers:
pull request needed more revision

PR #1272, opened 2024-07-30
Adding Termios library as std::libc::termios (posix libc functions)

PalsFreniers:
fixed version of posix libc's termios binding, I putted it inside libc/os/posix.c3 because it is a posix only part of the libc


lerno:
People will look for functions using their C name, e.g. `cfgetospeed`, so I would do it like this:

1. First just create the `extern` declarations as is, e.g. `extern fn CInt tcdrain(Fd fd);`
2. Then create the abstraction variant, e.g. `fn int drain(Fd fd) => tcdrain(fd);` or similar, and place it outside of the libc::os as an actual stdlib feature.

Because the idea is that people should in general not need to use libc functions directly, but instead go for (preferably) cross platform functionality that is *based on* libc.

So I could either accept this as-is and I'll do the changes afterwards, or you do modifications. What would be the best? If you use it then you probably have a better idea of how you think the "normal" stdlib interface should look like.

PalsFreniers:
I made the following changes:
- added extern functions corresponding to the c symbols;
- added abstraction variants for all;
- added unreachable function for all the symbols if either libc or posix is not available;

lib/std/libc/termios.c3
```diff
@@ -0,0 +1,115 @@
...
+fn int flush(Fd fd, int queue_selector) => tcflush(fd, queue_selector);
+fn int flow(Fd fd, int action) => tcflow(fd, action);
+fn Speed Termios.getOSpeed(Termios* self) => cfsetospeed(self);
+fn Speed Termios.getISpeed(Termios* self) => cfgetispeed(self);
+fn int Termios.setOSpeed(Termios* self, Speed speed) => cfsetospeed(self, speed);
+fn int Termios.setISpeed(Termios* self, Speed speed) => cfsetispeed(self, speed);
+fn int Termios.getAttr(Termios* self, Fd fd) => tcgetattr(fd, self);
+fn int Termios.setAttr(Termios* self, Fd fd, int optional_actions) => tcsetattr(fd, optional_actions, self);
+
+module libc::termios @if($or(!env::LIBC, !env::POSIX))
```
lerno, line 14: I don't think this will work, since the struct isn't defined when this module is active. So either remove it or create a dummy Termios struct. Like `struct Termios { void* dummy; }` would work.

lerno:
Thank you for your work. Now there's just the part when not on posix/libc

lib/std/libc/termios.c3
PalsFreniers: fixed it and also the missing semicolon at the end of the module

cbuttner:
Maybe some of the constants should also use the distinct types.

lerno:
Thank you!

PR #1275, opened 2024-07-30
Remove extra item from project.json boolean error message

Dodzey:
The error message for specifying an incorrect value for a boolean field was malformed due to the extra argument to format.

`In file 'project.json': An invalid '(null)' field that was not a boolean, please correct it.`


lerno:
Thank you!

PR #1281, opened 2024-07-31
Fix a warning/error where C compilers can not predict that a variable is initialized

yhyadev:
I had an issue while trying to build the project because warnings are set to errors and the compiler can't tell that it is always initialized because of  the switch after it


yhyadev:
image
Here is the error that was occuring before this commit

lerno:
Thank you

Issue #1286, opened 2024-07-31
LLVM assertion with duplicate function pointer signatures


Dodzey:
I am seeing the following assertion from LLVM:

```
% c3c-devel compile-run test.c3
Assertion failed: (be_value->type->canonical == type), function llvm_emit_parameter, file llvm_codegen_expr.c, line 5283.
```

This seems to be related to multiple function pointer defs with matching signatures.
In the reproducer below, if I modify the signature of FnB (Add an int argument), and modify b_func accordingly, the crash goes away.

Reproducible with:

 `c3c compile-run test.c3`

test.c3
```
module repro;

import std::collections::list;

def FnA = fn void(Foo*);
def FnB = fn void(Foo*);

struct Foo()
{
    List(<FnA>) aFuncs;
    List(<FnB>) bFuncs;
}

fn void! Foo.deinit(&self)
{
    self.aFuncs.free();
    self.bFuncs.free();
}

fn void Foo.add_fn_a(&self, FnA f) => self.aFuncs.push(f);
fn void Foo.add_fn_b(&self, FnB f) => self.bFuncs.push(f);

fn void a_func(Foo* foo) {}
fn void b_func(Foo* foo) {}

fn void! main()
{
    Foo foo;
    defer foo.deinit()!!;
    foo.add_fn_a(&a_func);
    foo.add_fn_b(&b_func);
}
```

lerno:
Can you make another issue for the casting error?

PR #1288, opened 2024-08-01
Function prototype may be NULL in `type_lowering` function

yhyadev:
This should fix a segmentation fault when dereferencing a NULL pointer after the last commit, I tried it also with #1286 reproduction code and it worked.

### DIFF
--- a/src/compiler/codegen_internal.h
+++ b/src/compiler/codegen_internal.h
@@ -37,6 +37,7 @@ static inline Type *type_lowering(Type *type)
 				type = type->decl->enums.type_info->type;
 				continue;
 			case TYPE_FUNC_PTR:
+                if (type->pointer->function.prototype == NULL) return type;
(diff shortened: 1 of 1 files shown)


lerno:
I'm afraid this is incorrect, the reason this would happen (and annoyingly I can't reproduce it), is that the raw prototype isn't created and it should be.

lerno:
Where you able to trace deeper when this happens? The behaviour is odd, as if there is some overwriting happening and the data is invalid.

yhyadev:
> Where you able to trace deeper when this happens? The behaviour is odd, as if there is some overwriting happening and the data is invalid.

It was occurring in every possible script I compile, opened GDB and catched the segfault as it happens, GDB gave me the location and I checked what type contain, it was a function called `SignalFunction` and it doesn't have any signature nor prototype

yhyadev:
image

yhyadev:
Fixed in master

PR #1289, opened 2024-08-01
Fix posix virtual memory reserve.

cbuttner:
Use `mprotect` to "commit" pages that are required. Even though pages are not actually committed to physical memory until touched, setting `PROT_WRITE` on the initial reserve appears to to matter for overcommit handling.

This fixes the 2 GB initial reserve issue for environments with low physical memory and/or where overcommitting is not allowed.

Additional reference:
https://www.kernel.org/doc/html/latest/mm/overcommit-accounting.html

### DIFF
--- a/src/utils/vmem.c
+++ b/src/utils/vmem.c
@@ -11,9 +11,9 @@
 
 #if PLATFORM_WINDOWS
 #include <windows.h>
-#define COMMIT_PAGE_SIZE 0x10000
 #endif
 
+#define COMMIT_PAGE_SIZE 0x10000
 
 static inline void mmap_init(Vmem *vmem, size_t size)
 {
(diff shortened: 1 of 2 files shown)


lerno:
Before this is added, please test the performance against abnormal data, e.g. one produced by https://github.com/nordlow/compiler-benchmark so that there is no performance regression.

cbuttner:
Testing with compiler-benchmark (default settings) on my Linux machine, checking takes 30ms longer, which is roughly 20% slower.
I'm getting the same performance penalty (140ms -> 170ms) on the frontend time for my own project too.

lerno:
As previously discussed, I'm going to reject this because of the performance regression, ok?

lerno:
We use the back off strategy then.

PR #1296, opened 2024-08-02
Implement passing arguments to program via `compile-run`

### LINKED ISSUE #1273: `compile-run` does not accept arguments to pass to the run program.
The current signature of it is `compile-run <file1> [<file2> ...]` and according to the implementation of `run_after_compile` it indeed does not accept any additional arguments to pass to the program: https://github.com/c3lang/c3c/blob/8d6dabf65c6df420b9030bc049f5ce3092cec01c/src/compiler/compiler.c#L622-L680

Are there any plans on allowing passing the arguments? Something akin to `compile-run <file1> [<file2> ...] [-- <arg1> [<arg2> ...]]`? If so may I implement it? :)

rexim:
I replaced system() with custom run_subprocess() which should work on both POSIX and Windows. It's a combination of some ideas I took from [subprocess.h](https://github.com/sheredom/subprocess.h) and [nob.h](https://github.com/tsoding/musializer/blob/f106c92934172096ed6822b6b9b276410cd99a31/nob.h). Subprocess.h is Public Domain. nob.h is my thing under MIT, but this specific piece of code is trivial enough, so I release it to Public Domain as well. :)

Please, let me know if there are any problems with the approach.

Close #1273


src/build/build_options.c
```diff
@@ -250,6 +250,16 @@ void append_file(BuildOptions *build_options)
 	vec_add(build_options->files, current_arg);
 }
 
+void append_arg(BuildOptions *build_options)
+{
+	if (vec_size(build_options->args) == MAX_FILES)
```
Martin005, line 255: A copy-paste typo from `append_file`

```suggestion
	if (vec_size(build_options->args) == MAX_ARGS)
```

lerno:
Ah.. old winnt.h and it's hogging of `TokenType`. Anything grabbing the whole of windows.h must be isolated into its own quarantined area. Maybe you could extract the windows code in a separate .c file which just exposes a function, and imports windows.h in that file only? Or just include the necessary .h files?

lerno:
Thank you!

rexim:
@lerno thank you! for making such a cool language and giving me an opportunity to submit the PR! The session was awesome! :D

lerno:
I only had a chance to see the beginning of your session, it was already useful to see what was obvious and what wasn't in the source code. I like to do cleanup passes every now and then, but it's been quite a while since I revisited the command-line part of the code and your session really reminded me that there is a build up of cruft there. 👍

lerno:
A weird issue occurs when running the compiler from an IDE (in my case CLion) – I get waitpid returns with a -1 and errno says there's an interrupt. I've previously used popen instead, and I have a vague recollection trying fork / execv before and seeing weirdness with this. I have to look at your session, but maybe you can tell me why you picked that over popen on Linux? Did popen work poorly?

rexim:
@lerno Hm... Before it was using `system()`, not `popen()`. I chose to do fork/execv because it's the standard way to just run a child process on POSIX system. I never used `popen()`, but from the man it feels like it's designed pipe stdout so you can read it. We don't care about the stdout of the child process.

I'm not really sure what exactly is the problem. Maybe we should create an issue and I'll investigate?

src/build/build_options.c
```diff
@@ -58,26 +58,26 @@ static void usage(void)
...
-	PRINTF("  headers <file1> [<file2> ...]           Analyse files and generate C headers for public methods.");
-	PRINTF("  vendor-fetch <library> ...              Fetches one or more libraries from the vendor collection.");
+	PRINTF("  compile <file1> [<file2> ...]                       Compile files without a project into an executable.");
+	PRINTF("  init <project name>                                 Initialize a new project structure.");
+	PRINTF("  init-lib <library name>                             Initialize a new library structure.");
+	PRINTF("  build [<target>]                                    Build the target in the current project.");
+	PRINTF("  benchmark                                           Run the benchmarks in the current project.");
+	PRINTF("  test                                                Run the unit tests in the current project.");
+	PRINTF("  clean                                               Clean all build files.");
+	PRINTF("  run [<target>]                                      Run (and build if needed) the target in the current project.");
```
atharvakale343, line 68: add the new help info to `c3c run [<target>]` as well?

lerno:
It's working from the terminal on MacOS but not when I run it in Clion that outputs to a terminal window. So there's something happening there I think. The interrupt corresponds to the user cancelling the fork on the console, so I think something is happening there.

lerno:
It turns out that LLDB is really poor at following forks, and I typically always use compile-run through the debugger when I test things. From what I can gather, this fork makes LLDB/Clion just lose connection and everything ends as soon as the forked process stops. The GDB situation is better I think, but I just can't use it easily on a non x64 Mac

rexim:
@lerno could you give me the steps to reproduce the problem with LLDB?

shanyssys:
@lerno not sure if it will help but: https://anadoxin.org/blog/lost-connection-when-debugging-with-lldb-on-mavericks.html/

shanytc:
@lerno  not sure if it will help but: https://anadoxin.org/blog/lost-connection-when-debugging-with-lldb-on-mavericks.html/

lerno:
Here is a super short code example:
```c
module main;
import std::io;

fn void main()
{
	io::printn("Test");
}
```
When I run this one in debug mode in CLion (which should be attaching to the compiler process implicitly), this happens *after C3 program stops running*

```
Launching ./testexe
Test
Could not wait on ./testexe (pid 32362): Interrupted system call
```

If I don't attach to the compiler at all, there is no problem and it works as expected.

However, I usually always run with the compiler attached when I do development, so I see this a lot.

Here is my repro with LLDB:

```
prompt$ lldb <path-to-c3c>/c3c -- compile-run testfile.c3
(lldb) r
Launching ./test
Test
Could not wait on ./test (pid 32693): Interrupted system call
Process 32679 exited with status = 1 (0x00000001) 
```

rexim:
@lerno I can't reproduce it still. Maybe it's something specific to an lldb version...

But here is a thing, subprocess.h and the musl's implementation of the function `system()` (which we used before my PR)  both start the new process with [posix_spawn(3)](https://man7.org/linux/man-pages/man3/posix_spawn.3.html). Maybe we should use it too here? May I move this to a separate issue and work on it?

lerno:
It can also be specific to MacOS. Moving to posix_spawn seems like an interesting idea. Like... in one way fork/exec is good. I'd like to be able to debug the c3 app directly when it's run. But the fact that it's having this issue isn't. And if it had meant the debugging actually correctly switched to the c3 app process, then it would have been worth it, but it seems like this isn't happening either.

So if you'd like to code the switch to spawn, that would be great.

PR #1303, opened 2024-08-03
Add methodsof to type info

Dodzey:
Adding draft pull request to get feedback on implementation, and discuss.
This is a useful feature, similar to `membersof` but returns the names of methods associated with the type.
Has previously been discussed on discord.

Usage:
```
$foreach ($mm : $Type.methodsof)
   $echo $mm;
$endforeach
```


data-man:
There are no tests. Does this support interfaces?

lerno:
I think `methodsof` could probably just return the names, this prevents the proliferation of compile time types. Because you can then get the actual function like this: 

```c
$foreach ($name : $Type.methodsof)
   $echo $name;
   // Let's grab the function pointer!
   void* func_ref = &$Type.$eval($name);   
$endforeach
```

lerno:
Look at test/unit/regression, you can add it there in the same manner. Use "assert" to test the expected results.

lerno:
CONST_METHOD should not be needed now I think.

lerno:
One final thing: it needs to be mentioned in the release notes. Can you add a line under 0.6.2?

lerno:
Thank you!

PR #1320, opened 2024-08-06
I want make library for http lib such as C++

archanaberry:
Please refactoring my code and fix it because, i'm still a beginner (⁠｡⁠ŏ⁠﹏⁠ŏ⁠)


lerno:
Please write the project in C3 first and make sure it works cross platform, on Windows, Linux and MacOS. Add unit tests as well. Then inclusion can be considered.

lerno:
I will close this for the time being.

PR #1327, opened 2024-08-08
Update string_iterator.c3 to include extra convenience methods

ellipse12:
Added peek: returns the next character without incrementing current
Added has_next: checks if the iterator has another element
Added get: gets the current element (the same one that was returned with the previous call to next).
Most of this is to make it easier to keep unicode support with common operations.


ellipse12:
Oops! the get function has a bug it should be:
```c3
fn Char32! StringIterator.get(&self)
{
    usz len = self.utf8.len;
    usz current = self.current;
    usz read = (len - current < 4 ? len - current : 4);
    Char32 res = conv::utf8_to_char32(&self.utf8[current - read > 0 : current - read ? 0], &read)!;
    return res;
}
```
(changed it so that current is clamped to 0 and also gets the correct element)

lerno:
Just update it and the pull request should update. I always squash pull requests in the end, so it doesn't matter if there are multiple commits.

ellipse12:
Ok good to know! Just fixed it.

lerno:
Looks like there's still an error?

ellipse12:
Thank you for your patience. :)

lerno:
Oh, I noticed there were no tests, can you add some tests? You can look into /test/unit/stdlib for examples. Just something to check what happens if each is called in the beginning, in the end and at the end where there is an invalid UTF8 sequence.

ellipse12:
Ok ill do that.

ellipse12:
Done! Sorry for all of the hassle this is my first PR so thank you for being patient with me. Also just wondering how do you build the Debian binaries? The default Debian package repo does not have llvm-17. (I ended up just using a virtual machine to test with)

lib/std/core/string_iterator.c3
```diff
@@ -1,5 +1,5 @@
 module std::core::string::iterator;
-
+import std::io;
```
lerno, line 2: Is this needed?

lerno:
I grab the llvm binaries from the LLVM org directly. If you look at this:
https://github.com/c3lang/c3c/blob/master/.github/workflows/main.yml

You can scroll down and see exactly the commands that are run to build the compiler for the different targets. You should be able to figure it out even if you don't know how Github defines things, because the actual commands are just direct commands send to the command line for each machine.

lerno:
Also, the first PR and getting the hang of what's needed is always a bit of a learning experience. Don't worry about it.

lib/std/core/string_iterator.c3
lerno: If not then we should remove it.

ellipse12: Yeah no it's not I just forgot to remove it after testing.

ellipse12:
Thanks for your understanding!

lerno:
Thank you!

Issue #1318, opened 2024-08-06
Adding the new arg passing syntax to `c3c run`


DrSloth:
We can pass arguments to `compile-run` using a `--`. This would also be nice to have on the `c3c run` command such that you can quickly try out your programs and pass arguments to them.

prithvidiamond1:
Hi! Can I have a go at this?

prithvidiamond1:
Created a PR for this.

Honestly, very little needed changing. Credit must go to @rexim for this.

I also implemented the arg passing syntax for the `clean-run` command cause I am sure someone will ask for the feature sometime in the future and it was once again minimal effort.

PR #1328, opened 2024-08-08
Implemented arg passing to `clean-run` and `run` commands

prithvidiamond1:
This change is in response to #1318.

Changed `COMMAND_CLEAN_RUN` and `COMMAND_RUN` to return true when passed to function `command_passes_args()`.
Also, updated the compiler usage message to reflect that `clean-run` and `run` commands can be passed args.

Did some local testing and everything seems to work fine on my end (this was on windows-x64).

Let me know if anything needs to be changed.

### DIFF
--- a/src/build/build_options.c
+++ b/src/build/build_options.c
@@ -66,11 +66,11 @@ static void usage(void)
 	PRINTF("  benchmark                                           Run the benchmarks in the current project.");
 	PRINTF("  test                                                Run the unit tests in the current project.");
 	PRINTF("  clean                                               Clean all build files.");
-	PRINTF("  run [<target>]                                      Run (and build if needed) the target in the current project.");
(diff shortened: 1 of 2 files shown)


prithvidiamond1:
@lerno Apologies for the late reply. Thanks for merging the PR. I wanted to ask about this but wasn't sure if it was relevant here or if it needs to be a separate issue. From my quick look at the test folder, there don't seem to be any tests for testing the compiler commands (if I am wrong about this do let me know). I was wondering if you want to create tests for all the compiler commands and their optional arguments. It might be worth making it an issue for anyone to tackle.

lerno:
There are some sanity testing, like running a test project and so on, but there is nothing really organized. It's a good idea, I'll create an issue for it.

lerno:
Here it is: #1336

prithvidiamond1:
Great, I will take a look at it. Also, you may want to consider closing #1318 since the changes have been merged now.

PR #1333, opened 2024-08-08
Fix the win32 native_clock() function in std::time::os

cheese3660:
The win32 `native_clock()` function had a bug where it was retrieving the performance frequency via `win32_QueryPerformanceFrequency` but never using it, leading to incorrect timestamps using `Clock.mark`, this resolves that bug by changing the `native_clock()` function to use the performance frequency in converting from the performance counter value to nanoseconds.


lerno:
Thank you

PR #1334, opened 2024-08-08
Move safe_mul_div macro and make it generic on integer types

cheese3660:
This moves the recently added `safe_mul_div` macro in `std::time::os` (win32) to `std::math` and then should make it generic for integer types.


lerno:
Add a unit test and rename it mul_div

lerno:
Or possibly int_mult_div. One can also consider adding it as a macro method on all numbers, so you can do `foo.mult_div(a, b)` on all integers and integer vectors.

cheese3660:
I think I'll see about doing the latter option there

cheese3660:
should the integer vector variants apply it as a scalar over all members, or do it with another vector of numbers, or be able to do both?

cheese3660:
Alright, with some testing, I have found that these functions are only guaranteed to work correctly for any number even if `N*mul` results in a larger bitwidth than the result can hold *if* `mul*div` does not result in a number with a larger bitwidth than the type the operation is being done on, or the operation is being done on a scalar type that would promote to int. But the changes I just pushed should implement what you asked

lerno:
Thank you!

lerno:
I think it was squashed as there was the same fix in master?

PR #1338, opened 2024-08-09
Add `c-include-dirs` project/manifest setting.

cbuttner:
Set the include directories to be used when compiling C sources.


lerno:
This looks nice, I assume you tested it with some project?

cbuttner:
Yes, it's for this: https://github.com/c3lang/tree-sitter-c3/blob/main/bindings/c3/tree_sitter_c3.c3l/manifest.json

PR #1342, opened 2024-08-10
Add block asm support for RV32I and RV64I

chuckb:


lerno:
Adding these limited bit sized types has a profound effect on the compiler, the language semantics and the standard library, so we need to find a different solution.

I don't quite see why you would need such a drastic solution though?

chuckb:
I initially implemented editing of these types at code gen time. So while I could print an error, code still generated, which ultimately caused a downstream crash (of either clang or llvm...was not exactly sure). When I looked at your lexical analysis and parsing, I saw where you had edit checks already in place and so I just leveraged what you had. I agree that changes were more impactful than seemed warranted. Can you suggest a better mechanism and location at which this could be done with less impact?

lerno:
I probably need to see how it looked before this and see the diff to see what you added.

chuckb:
The first commit I made here: https://github.com/c3lang/c3c/commit/3db37207a6f219c70e7f8b98032324b7730d8935

If you put in an immediate value outside the range of -2048 to 2047, I believe I issue an error message, and then code is generated anyway, and then I think the compiler segfaults or crashes. You also gave me a note to do a semantic check instead, which I am now doing with the type system. If there is another easier, less disruptive alternative, let me know.

lerno:
If you look at: 
`static inline bool sema_check_asm_arg_const_int(SemaContext *context, AsmInlineBlock *block, AsmInstruction *instr, AsmArgType arg_type, Expr *expr, Expr *int_expr)`, that's where the check happens I believe. Extending that one to check for limited type size should be the right thing.

chuckb:
Cool. Let me see what I can do.

lerno:
I made some refactorings right now, they should be straightforward to resolve, you need to pass a `PlatformTarget` around.

src/compiler/types.c
```diff
@@ -75,8 +75,11 @@ Type *type_int_signed_by_bitsize(BitSize bitsize)
 {
 	switch (bitsize)
 	{
+		case 5: return type_ichar;
```
lerno, line 78: This should not still be in here or?

src/compiler/sema_asm.c
```diff
@@ -78,6 +78,80 @@ static inline bool sema_reg_float_suported_type(AsmArgType arg, Type *type)
...
+		}
+		else if (arg_type.imm_arg_ibits & (ARG_BITS_8|ARG_BITS_16|ARG_BITS_32|ARG_BITS_64|ARG_BITS_128|ARG_BITS_256|ARG_BITS_512|ARG_BITS_80)) {}
+		else UNREACHABLE;
+	}
+	else 
+	{
+		if (arg_type.imm_arg_ubits > 0)
+		{
+			if (arg_type.imm_arg_ubits & ARG_BITS_20) max = (Int128){ 0, UINT20_MAX };
+			else if (arg_type.imm_arg_ubits & ARG_BITS_12) max = (Int128){ 0, UINT12_MAX };
```
lerno, line 116: Please use { } in an if-else like this.

src/utils/stringutils.c
```diff
@@ -418,4 +418,14 @@ char *scratch_buffer_copy(void)
 	return str_copy(scratch_buffer.str, scratch_buffer.len);
 }
 
-
+void scratch_buffer_backspace(size_t len)
+{
+	if (scratch_buffer.len < len)
```
lerno, line 423: I think this looks wrong? I'd set an assert that len <= scratch_buffer.len and just keep scratch_buffer.len -= len

lerno: Because no one sane should be giving a larger backspace than there is space.

src/compiler/sema_asm.c
```diff
@@ -78,6 +78,80 @@ static inline bool sema_reg_float_suported_type(AsmArgType arg, Type *type)
 	return type_bit_size(type) == arg_bits_max(arg.float_bits, 0);
 }
 
+static inline bool sema_check_imm_fits(Int imm, AsmArgType arg_type)
+{
+	Int128 min = {0};
```
lerno, line 83: This looks fine, but I think it can be semantically compressed somewhat?

src/compiler/types.c
chuckb: Only used by asm, right? I am allowing the asm instruction grammar to accept immu5, for example. I could just send immi16 and then hard code by instruction. Seems kludgy tho.

src/compiler/sema_asm.c
chuckb: Ok, will fix.

src/compiler/parse_expr.c
```diff
@@ -1520,7 +1520,7 @@ Expr *parse_integer(ParseContext *c, Expr *left)
 	{
 		int min_bits = type_size(type_cint) * 8;
 		Int test = { .i = i };
-		for (int type_kind = 0; type_kind < 5; type_kind++)
+		for (int type_kind = 0; type_kind < 8; type_kind++)
```
lerno, line 1523: This looks like a leftover from before?

src/utils/stringutils.c
chuckb: This was extra code...not needed now anyway. Will remove.

src/compiler/parse_expr.c
chuckb: Oops. Yep. Missed one. Will fix.

src/compiler/types.c
lerno: Can you explain?

chuckb: max_supported_imm_int and arg_bits_max will return null if the odd bit length (5 bits in this case) is not defined. So it means you'd have to declare immediates as immu8 in this case.

src/compiler/sema_asm.c
chuckb: Probably, but I just duped the flow I saw in int_fits.

chuckb:
Although what I have works, I don't like the hardcoding of immediate instructions in codegen_create_riscv_arg. But I can't seem to get easy access to the types, thus I have to duplicate knowledge of signedness and sign extension. Ugly. Is there a better way?

src/compiler/types.c
chuckb: Finally realized how I could factor this out.

lerno:
Is this ready for review now?

chuckb:
Yes, I hope so.

lerno:
Sorry for being late in reviewing this, I've had a hectic day and then I had to test the last of what I had in dev.

chuckb:
No worries. Thanks for the heads up. I have several show and tell vids planned.

src/compiler/codegen_asm.c
```diff
@@ -112,6 +116,103 @@ static inline void codegen_create_aarch64_arg(AsmInlineBlock *block, unsigned in
...
+static inline void codegen_create_riscv_arg(PlatformTarget* platform_target, 
+	AsmInlineBlock *block, unsigned input_offset, unsigned arg_idx, AstAsmStmt* asm_stmt)
+{
+	Expr** args = asm_stmt->args;
+	Expr *expr = args[arg_idx];
+	ExprAsmArg *arg = &expr->expr_asm_arg;
+	switch (arg->kind)
+	{
+		case ASM_ARG_INT:
+			if (strcmp(asm_stmt->instruction, "addi") == 0 ||
```
lerno, line 128: All instruction name will be interned. This means that if you store the interned name, you can do a simple:
```c
if (asm_stmt->instruction == KW_ADDI) 
```
But since you're already here, why not grab the instruction and encode this information in the instruction itself? 

Search for this
```c
AsmInstruction *instr = asm_instr_by_name(&compiler.platform, asm_stmt->asm_stmt.instruction);
```
Then given that it is already encoded, all you need to do is:
```c
if (instr->cast_extend) scratch_buffer_append_signed_int(CAST_AND_EXTEND(arg->value, instr->cast_extend));
```

src/compiler/codegen_asm.c
```diff
@@ -112,6 +116,71 @@ static inline void codegen_create_aarch64_arg(AsmInlineBlock *block, unsigned in
...
+			if (instr->param[arg_idx].imm_arg_ibits > 0) {
+				scratch_buffer_append_signed_int(CAST_AND_EXTEND(arg->value, get_imm_bits(true, instr->param[arg_idx])));
+			} 
+			else
+			{
+				scratch_buffer_append_unsigned_int(arg->value);
+			}
+			return;
+		case ASM_ARG_REG:
+			if ((strcmp(asm_stmt->instruction, "lw") == 0 ||
```
lerno, line 138: Can you do the same and integrate these into the instruction? The instruction should encode the rules, and this code should run the rules. Here the instructions are hacked into the rule code, which isn't the way it was intended.

src/compiler/compiler_internal.h
```diff
@@ -3670,3 +3670,33 @@ INLINE bool check_module_name(Path *path)
 	return true;
 }
 
+static inline unsigned int get_imm_bits(bool is_signed, AsmArgType arg)
+{
+	if (is_signed)
```
lerno, line 3675: This seems to do the same thing as arg_bits_max? What is the difference? Use `unsigned` rather than `unsigned int`, since that's consistent with the rest of the code.

chuckb: If you return a "irregular" sized imm from arg_bits_max, type_int_signed_by_bitsize will complain, and you did not want me to tinker with that fn. And, I don't want the bits of the largest "regular" type that will fit given bits. I want the given bits. I will make the return type change as indicated.

lerno: But that's not how imm_bits work, it's saying all allowed sizes?

lerno: But I could fix this afterwards if you want.

chuckb:
I'll need to look into these CI problems tomorrow.

chuckb:
Failures related to changing ExprAsmArg structure. Tried to change static assert. It resolved MS but broke linux. How to resolve?

src/compiler/compiler_internal.h
```diff
@@ -873,7 +873,11 @@ typedef struct
 			ExprId idx;
 			uint64_t offset;
 		};
-		uint64_t value;
+		struct {
+			uint64_t value;
+			bool is_neg;
+			AsmArgType bits;
+		};
```
chuckb, line 880: Failures are related to changing ExprAsmArg structure. I tried to update the static assert below. It resolved MS but broke linux. How to resolve?

lerno: What was the change?

chuckb: I added those two values (is_neg and bits) into the ExprAsmArg structure. There is a static_assert on line 1116 guarding the size of Expr. This assert now fails for Windows platforms unless I change the size. And when I do, linux platforms fail. Alignment/padding issue?

lerno: Wow, you placed the huge `AsmArgType` into the expression itself. It's data for a particular instruction and show never be in an expression. This would be similar to copying types by value into resolved type infos rather than having a link to the unique type. I'll have a look and see if I can suggest an alternative.

chuckb: The only thing I actually need is the bit size for a signed int, so the whole structure is not necessary. It was just there and convenient. I can factor it out.

chuckb: Perfect. Let me yank the change down, run a few more tests, and then I flip this back to review when ready. Thanks!

chuckb:
I'm passing around platform_target and it's now unnecessary. One more commit and I hope this will be ready.

lerno:
Anything else needed, or is it ready to be merged?

chuckb:
> I need you to squash this one to a single commit. Right now trying to rebase this on top of master is a nightmare because of things added and removed and added. I want this to be a single clean commit.

Will do.

chuckb:
Evidently I did something wrong in the squashing process. Investigating...

PR #1344, opened 2024-08-11
added the necessary library to build on void linux in the README.md

Sergwest585:
there was a problem during the build on void linux in the form of missing libraries: lld17-devel llvm17-devel


lerno:
Thank you

PR #1351, opened 2024-08-12
Update README.md

mmsh1:
Duplicate package removed from Void Linux section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -323,7 +323,7 @@ You can try it out by running some sample code: `./c3c compile ../resources/exam
 
 #### Compiling on Void Linux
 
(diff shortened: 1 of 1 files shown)


kjloveless:
Definitely a duplicate, but I'm not sure any of the lld17-* things are needed if llvm17 is included.

https://github.com/void-linux/void-packages/blob/master/srcpkgs/llvm17/template

It looks like the lld17* packages are included as subpackages by default?

Maybe I'm wrong though... If so maybe lld17 was supposed to listed once, instead of this duplicate.

lerno:
Isn't that perhaps just lld and not lld-devel?

lerno:
Thanks for the pull request BTW

PR #1358, opened 2024-08-14
Escape arguments to platform linker/compiler.

cbuttner:
For the most part fixes problems when a file path passed to platform linker or compiler has whitespace in it.

I assume `linker.c` is going to be cleaned up/refactored in the future and argument handling may be handled a bit differently, so this is a quick fix until then.

Tested on Linux, macOS and Windows with a variety of inputs.

### DIFF
--- a/src/compiler/linker.c
+++ b/src/compiler/linker.c
@@ -665,6 +665,7 @@ const char *concat_string_parts(const char **args)
 	scratch_buffer_clear();
 	FOREACH(const char *, arg, args)
 	{
+		assert(arg != &scratch_buffer.str && "Incorrectly passed a scratch buffer string as an argument.");
 		if (PLATFORM_WINDOWS && arg == *args)
 		{
 			scratch_buffer_append(arg);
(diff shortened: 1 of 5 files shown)


src/compiler/linker.c
```diff
@@ -673,22 +662,19 @@ bool obj_format_linking_supported(ObjectFormatType format_type)
...
 const char *concat_string_parts(const char **args)
 {
-	unsigned size_needed = 0;
-	FOREACH(const char *, arg, args)
-	{
-		size_needed += strlen(arg) + 1;
-	}
-	char *output = malloc_string(size_needed);
-	char *ptr = output;
+	scratch_buffer_clear();
```
lerno, line 665: I am not sure this is safe. Are the arguments all guaranteed to not be in the scratch buffer?

cbuttner: I don't know, you should be more familiar with the scratch buffer's usage patterns.

lerno: I added a check now, so that should tell if the assumption is ever wrong.

PR #1365, opened 2024-08-16
Add `types::is_signed`, `is_unsigned` and `inner_type`.

cbuttner:
By the way, I think `inner_type` and `inner_kind` should be renamed since `inner` is a bit overloaded with meaning, but these macros only remove distinct. Maybe `flatten` and `flatten_kind`? Or `remove_distinct` and `remove_distinct_kind` to be more obvious.


lerno:
Yes, you might do so here actually before we do a pull on this. You need to keep a copy of the function called `inner_type` but it can be marked `@deprecated`. If you do this though, then all other stdlib uses of inner_type needs to be fixed up.

PR #1372, opened 2024-08-20
Add RISC-V block asm support

chuckb:
Add initial block assembler support for RISC-V 32 and 64 bit arch.


chuckb:
After a couple of unsuccessfully squash sessions, I just pulled a diff, did a forced reset, and reapplied the diff. That blasted the PR with the conversation history. Unfortunate. But here we are. This should be ready.

lerno:
Sometimes that's the only way. It's worse trying to deal with weird merges where suddenly changes disappear and other fun things.

lerno:
Thank you for all your hard work!

PR #1373, opened 2024-08-20
Simplify some asm code and update releasenotes for RISCV

lerno:
@chuckb Does this one look fine? I tried to simplify the code a bit.

### DIFF
--- a/releasenotes.md
+++ b/releasenotes.md
@@ -118,6 +118,7 @@
 - Added `--list-manifest-properties` to list the available properties in `manifest.json`.
 - Indexing into a constant array / struct now works at compile time.
 - Improved error message when trying user foreach with an untyped list.
+- RISCV asm support.
 
 ### Fixes
 - Error with unsigned compare in `@ensure` when early returning 0 #1207.
--- a/src/compiler/sema_asm.c
+++ b/src/compiler/sema_asm.c
@@ -78,88 +78,51 @@ static inline bool sema_reg_float_suported_type(AsmArgType arg, Type *type)
 	return type_bit_size(type) == next_highest_power_of_2(arg_bits_max(arg.float_bits, 0));
 }
 
-static inline bool sema_check_imm_fits(Int imm, AsmArgType arg_type)
+static inline bool sema_check_npot_imm_fits(Int imm, AsmArgType arg_type)
 {
-	Int128 min = {0};
-	Int128 max = {0};
-	bool is_signed = false;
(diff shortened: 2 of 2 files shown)


lerno:
Pinging @chuckb again

chuckb:
Sorry, I did not see this notification till now. Will look...

chuckb:
Oh yes I remember now...

If I do a test like:
```
// #target: elf-riscv32
module test;

fn void main(String[] args)
{
	asm
	{
		and $t0, $t1, $t2;
		or $a0, $a1, $a2;
		xor $s7, $s6, $s5;
		andi $s1, $s2, -2049; // #error: 'andi' expected 'short' limited to 12 bits.
		ori $s3, $s4, 2047;
		xori $t3, $t4, 1;
		not $t3, $t4;
	}
}

/* #expect: test.ll

"and t0, t1, t2\0Aor a0, a1, a2\0Axor s7, s6, s5\0Aandi s1, s2, -2048\0Aori s3, s4, 2047\0Axori t3, t4, 1\0Anot t3, t4\0A", "~{x5},~{x9},~{x10},~{x19},~{x23},~{x28}"()
```

I get:
```
- 1/0 test_suite/asm/asm_bit_rv.c3t: Failed.
Unexpected error in asm_bit_rv.c3 line 10:"'andi' expected 'short' limited to 12 bits."
Expected errors that never occurred:
1. asm_bit_rv.c3 line: 11 expected: "'andi' expected 'short' limited to 12 bits."
Found 1 tests: 0.0% (0 / 1) passed (0 skipped).
```

And now looking at the problem more closely, I think it is related to the first target line comment throwing off the line counting.

lerno:
That test is weird. Use .c3 on tests that aren't expected to compile / doesn't need to generate a binary, and .c3t on tests that actually has and #expect clause at the end.

chuckb:
Ah, and now I see. The actual test is being copied to temp and the target comment line is missing (on purpose evidently but it's throwing off the line count).

chuckb:
Ahhhh, ok. Got it.

lerno:
Yes because .c3t files are not supposed to error.

chuckb:
So, now that I understand how to write those negative tests, how would you like for me to proceed? Your change looks fine and I did one manual test. I'd like to get a few into the suite.

lerno:
I think I can merge this one then, but please add more tests and make a pull request for them. The more tests, the more stable the compiler will be.

PR #1378, opened 2024-08-22
RISCV: Correct auipc imm; clarify signed imm error; add imm negative t…

chuckb:
…ests.

Note that specifying an immediate integer without the u suffix for an imm defined as unsigned will generate an error. I'm not sure if this behavior is desired...seems pedantic but maybe needs to be. Usually one would be using hex anyway, so likely will not be noticed.


test/test_suite/asm/asm_imm_err_rv.c3
```diff
@@ -0,0 +1,66 @@
...
+	{
+		auipc $a0, -1; // #error: 'auipc' does not support a signed integer constant here.
+	}
+}
+
+fn void test7()
+{
+	asm
+	{
+		auipc $a0, 1; // #error: 'auipc' does not support a signed integer constant here.
```
lerno, line 56: This does not make sense to me. Why wouldn't 1 work. It's implicitly possible to convert to an unsigned integer.

chuckb: I agree. But it never worked. Or perhaps it was due ongoing modifications. So I will change it so it does.

lerno: I only had asm with arguments that were signed, and those were simply range checked. Note that C3 implicitly converts constants back and forth between unsigned and signed *as long as they fit*. So i/u is just a matter of calculating the valid range.

chuckb:
Assuming CI is happy, this should be ready...

lerno:
Thank you!

PR #1387, opened 2024-08-27
Print Git Hash on `--version`

rexim:
Since I primarily use Pre-Release (the Release does not have the features that I need yet) I want to be aware on which commit I currently am. I think it could be generally useful for anybody to have this information in `--version`. I don't know if this needs any tests. Please let me know if there are any issues with this approach.


lerno:
Nice addition, thank you!

PR #1391, opened 2024-08-29
Properly persist git hash on each build

rexim:
From https://github.com/c3lang/c3c/pull/1387#issuecomment-2315386781

The approach for embedding the git hash into the executable I introduced is flawed. It only works at the configuration step. So if you fetch your changes and simply rebuild the compiler, you end up with the previous hash still being displayed on `--version`. This is a very important issue as it makes the application display misleading information. I extremely apologize for wasting time and not testing my PRs properly.

Here is the new approach:
- Instead of passing git hash via a macro definition we save it into a separate `./build/git_hash.h` file that is then included into the source code. 
- The `./build/git_hash.h` file is generated by a standalone script `./git_hash.cmake`. 
- This script is not part of `CMakeLists.txt` and it is meant to be run via `cmake -P ./git_hash.cmake` from `CMakeLists.txt`. 
- We introduce a custom target `generate_git_hash` that runs `./git_hash.cmake` and produces `./build/git_hash.h` as a byproduct. 
- We then make `c3c` executable depend on `./build/git_hash.h` which triggers `generate_git_hash` on every build.

Tested locally on Linux and everything seems to be file. Let's see how it performs on CI. Please let me know if you see any flaws. Sorry again for wasting time.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -101,19 +101,6 @@ if(C3_USE_TB AND GIT_FOUND AND EXISTS "${CMAKE_SOURCE_DIR}/.git")
     endif()
 endif()
 
-if(GIT_FOUND AND EXISTS "${CMAKE_SOURCE_DIR}/.git")
-    # Get git hash
-    execute_process(COMMAND git rev-parse HEAD
-            OUTPUT_VARIABLE GIT_HASH
-            RESULT_VARIABLE GIT_REVPARSE_RESULT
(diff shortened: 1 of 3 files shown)


rexim:
Had to change the logic again, as it was acting a bit annoying. I'm starting to think that backing a git hash into the executable using cmake is an unsolved computer science problem. But I hope I nailed it this time.

lerno:
CI is happy, so let's try it.

PR #1397, opened 2024-09-01
Adding check of HTTP response code for Windows in download_file()

chamaeleon:
Adding a check for the HTTP response code for Windows builds. Without it, when a download fails (incorrect vendor library name, for example), a file with the expected name remains, but may contain the HTML content the web server sends when specifying an invalid url, etc. Checking for a successful 200 response allows the removal of the file when a download fails.


lerno:
Thank you!

PR #1400, opened 2024-09-02
Add fn decl alternatives for parameter in/out doc contracts

Dodzey:
This adds support for supplying in/out parameter attributes next to the parameter in the function declaration.
Creating as a draft to open discussion about possibility of introducing this to the language.

```
/**
 * @param [in] i
 * @param [out] o
 */
fn void foo_a(int* i, int* o) { *o = *i; }

// Equivalent to the above foo_a
fn void foo_b(int* i @in, int* o @out) { *o = *i; }

fn void foo_error(int* i @in) { *i = 123; }

fn void main() {
    int i = 123;
    int o = 0;
    foo_a(&i, &o);

    o = 0;
    foo_b(&i, &o);

    foo_error(&o); //  Error: 'in' parameters may not be assigned to.
}
```


joshring:
For regular functions would that be like?

```c
/**
 * @param [in] i
 * @param [out] o
 */
fn void foo_a(int* i, int* o) 
{
    return *o = *i;
}

// Equivalent to the above foo_a
fn void foo_b(int* i @in, int* o @out)
{
    return *o = *i;
}

fn void foo_error(int* i @in)
{
    return *i = 123;
} 

fn void main() {
    int i = 123;
    int o = 0;
    foo_a(&i, &o);

    o = 0;
    foo_b(&i, &o);

    foo_error(&o); //  Error: 'in' parameters may not be assigned to.
}
```

lerno:
As per discussion, I think we'll hold off with this change, ok?

lerno:
As per discussion I'm closing this one. However, let's keep the discussion alive on how to support brief functions in a nice way.

Issue #1401, opened 2024-09-02
Compile time slicing of strings does not work in $eval


rexim:
Do I understand correctly that as of right now there is no way to do compile time slicing of strings? Something opposite to `$concat` sort of speak. `$s[$start..$end]` is considered a runtime expression.

lerno:
No, that should be possible. File a bug?

rexim:
bug.c3:

```c
import std::io;

macro test($name) {
    int ab;
    $eval($name[0..1]) = 69;
    io::printn(ab);
}

fn void main() {
    test("abc");
}
```

```console
$ c3c compile bug.c3
 2: 
 3: macro test($name) {
 4:     int ab;
 5:     $eval($name[0..1]) = 69;
              ^^^^^^^^^^^
(/home/rexim/Programming/probe/c3-compile-slice/bug.c3:5:11) Error: Expected a compile time expression.

 7: }
 8: 
 9: fn void main() {
10:     test("abc");
        ^^^^^^^^^^^
(/home/rexim/Programming/probe/c3-compile-slice/bug.c3:10:5) Note: Inlined from here.
```

Please let me know if I'm doing something too stupid. :D

rexim:
Updated the title to emphasize that this is bug.

lerno:
I've reworked a lot of the code and there are some interesting complications.

First of all I tackled this as a general problem of slicing any compile time constant. So for example doing this at compile time: `(int[6] { 1, 22, 33, 4, 55, 66 })[2..4]` (or the equivalent, like `$b[2..4]` for $b containing the same array).

However, in general slicing anything will produce a slice, which lacks a length. So `$b[2..4]` should internally be an `int[3] { 22, 33, 4 }` but have the type `int[]`.

It is tempting to say that runtime slicing `$b[2..4]` produces an `int[3]`, but since an array doesn't implicitly convert into a slice, that's somewhat problematic.

So there are roughly 3 possibilities:

1. Allow compile time slicing of arrays and vectors, and have them produce slices, but add support for "compile time slices" (beyond bytes and Strings)
2. Allow compile time slicing of arrays and vectors but have them produce arrays. However, constant arrays may be implicitly converted to slices (this is not that weird, this is already ok: `int[] a = { 1, 2, 3 };`
3. Only allow compile time slicing on String and bytes constants.

Any opinion on what you would prefer?

lerno:
Compile time slicing should now work. I'll add a new issue for the remaining.

[...]

PR #1404, opened 2024-09-03
Quick hack to make string slices work as compile time values #1401

rexim:
Just quickly debugged the problem and came up with this quick hack that makes my example from #1401 compile and work as expected. I'm not sure how proper of a solution this is, 'cause I'm obviously lacking the full knowledge about the code base. Feel free to reject it if it's too stupid. Just wanted to share my findings if they are of any help.

The essence of the hack:

`sema_expr_flatten_assign()` is not recognizing `EXPR_SLICE` as valid compile time constructions at all. What we add is handling of `EXPR_SLICE` almost like `EXPR_SUBSCRIPT` ('cause they apparently almost the same thing), but we specifically expect the parent expression to be `CONST_STRING`. All of this results in a Slice Expression over a `CONST_STRING` being folded into a `CONST_STRING`.

### DIFF
--- a/src/compiler/sema_expr.c
+++ b/src/compiler/sema_expr.c
@@ -4435,6 +4435,25 @@ static bool sema_expr_flatten_assign(Expr *expr)
 			return true;
 
 		}
+		case EXPR_SLICE:
+		{
+			if (!range_is_const(&expr->subscript_expr.range)) return false;
+			Expr *parent = exprptr(expr->subscript_expr.expr);
+			Type *flat_type = type_flatten(parent->type);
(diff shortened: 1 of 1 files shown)


lerno:
I could add this if you need it quickly otherwise the long term solution is to expand on const handling a bit. I have some code that does this locally, but I realized I wanted to clean up the code to unify the constant handling in a safer and simpler way. So that's why I postponed it. What do you think? Do you want it in now, knowing it will be replaced later?

rexim:
Nah, I don't need it now. I was just curious what exactly is going on in here. :) It's better to do the proper thing.

PR #1405, opened 2024-09-03
Rename `inner_kind` and `inner_type`.

cbuttner:
Rename to `remove_distinct_kind` and `remove_distinct`.

Are these names good? It might not be obvious they remove distinct *recursively*.
The alternative contenders would be `flatten_kind` and `flatten`, but then it's not entirely obvious what it does either.

### DIFF
--- a/lib/std/core/mem.c3
+++ b/lib/std/core/mem.c3
@@ -373,10 +373,10 @@ macro void set_inline(void* dst, char val, usz $len, usz $dst_align = 0, bool $i
 	$$memset_inline(dst, val, $len, $is_volatile, $dst_align);
 }
 /**
- * @require values::@inner_kind(a) == TypeKind.SLICE || values::@inner_kind(a) == TypeKind.POINTER
- * @require values::@inner_kind(b) == TypeKind.SLICE || values::@inner_kind(b) == TypeKind.POINTER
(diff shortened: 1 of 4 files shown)


lerno:
I'm thinking "base" might be better - talking about the "base type" of something.

lerno:
Any thoughts?

lerno:
In the compiler, removing the `distinct`-ness is known as `flattening`, whereas reducing types to their backend representations is known as lowering (and should only be done in the backend), maybe this can be used if it is only distinct?

cbuttner:
`base_type` is not bad but it's also not super clear, just from the name alone one might think it returns the underlying type of an enum or optional, or it could resolve subtyped structs with `inline` to it's "base type".
I think it's a similar problem with `flatten` or `lower`.
I personally like `remove_distinct` for it's clarity but I'm open to whatever.

lerno:
`flatten` and `lower` does have the advantage of being well defined within the compiler.

lerno:
`distinct_of`?

sa-tasche:
`distinctof` to be consistent.

lerno:
@sa-tasche That seems to be a bit reverse of what is meant though? It's getting the type the distinct is using as a based.

lerno:
I'm going to close this because I don't think there is a better name for now. Let's revisit this later when the compiler have gone through further refactorings.

PR #1409, opened 2024-09-03
Add QOI to the standard library

Hema2-official:
Hiyya!

Here's QOI from [qoi-c3](https://github.com/Hema2-official/qoi-c3).
I've done a _minor_ cleanup to match other modules just a little bit better.

As soon as this is merged, I'll archive my repository and continue improving it here (sometime in the future, maybe).

Take a look and see what else should be changed, if anything.

<33


lib/std/compression/qoi.c3
```diff
@@ -0,0 +1,471 @@
...
+	char colorspace;	// 0 = sRGB with linear alpha, 1 = all channels linear
+}
+
+const char[*] END_OF_STREAM = {0, 0, 0, 0, 0, 0, 0, 1};
+
+// inefficient, but it's only run once at a time
+macro @enumcast($Type, raw)
+{
+	foreach (value : $Type.values)
+		if (value.id == raw) return value;
```
lerno, line 413: Use { } around this

lib/std/compression/qoi.c3
```diff
@@ -0,0 +1,471 @@
...
+ * and use of the data should be wrapped in a @pool() { ... }; block.
+ * 
+ * @param [in] filename `The file's name to read the image from`
+ * @param [&out] desc `The descriptor to fill with the image's info`
+ * @param channels `The channels to be used`
+ */
+fn char[]! read(String filename, Desc* desc, Channels channels = AUTO) @if(!$feature(QOI_NO_STDIO))
+{
+	// read file
+	char[]! data = file::load_new(filename);
```
lerno, line 118: It's probably worthwhile to let the user provide their own allocator, but this can be a later improvement.

lib/std/compression/qoi.c3
```diff
@@ -0,0 +1,471 @@
...
+	char[] image = mem::alloc_array(char, image_size);
+	defer catch mem::free(image);
+
+	for (loc = 0; loc < image_size; loc += channels.id)
+	{
+		// get chunk tag
+		tag = data[pos];
+
+		// check for chunk type
+		switch {
```
lerno, line 332: Just a small thing, put this { on the next line.

lib/std/compression/qoi.c3
```diff
@@ -0,0 +1,471 @@
...
+// inefficient, but it's only run once at a time
+macro @enumcast($Type, raw)
+{
+	foreach (value : $Type.values)
+		if (value.id == raw) return value;
+	
+	return QOIError.INVALID_DATA?;
+}
+
+def Pixel = char[<4>];
```
lerno, line 418: You probably want `distinct Pixel = inline char[<4>]` here to make it distinctly typed.

lib/std/compression/qoi.c3
```diff
@@ -0,0 +1,471 @@
...
+			case tag == OP_RGBA:
+				OpRGBA* op = @extract(OpRGBA, data, &pos);
+				p = { op.red, op.green, op.blue, op.alpha };
+				palette[p.hash()] = p;
+				
+			case (tag >> 6) == OP_INDEX:
+				OpIndex* op = @extract(OpIndex, data, &pos);
+				p = palette[op.index];
+
+			case (tag >> 6) == OP_DIFF:
```
lerno, line 350: Note that the precedence is different in C3, so you can write `tag >> 6 == OP_DIFF` if you want.

lib/std/compression/qoi.c3
```diff
@@ -0,0 +1,471 @@
...
+	foreach (value : $Type.values)
+		if (value.id == raw) return value;
+	
+	return QOIError.INVALID_DATA?;
+}
+
+def Pixel = char[<4>];
+macro char Pixel.hash(Pixel p) {
+	return (p.r * 3 + p.g * 5 + p.b * 7 + p.a * 11) % 64;
+}
```
lerno, line 421: This should also work: `return (p * { 3, 5, 7, 11 }).sum() % 64;`

lib/std/compression/qoi.c3
```diff
@@ -0,0 +1,471 @@
...
+
+struct OpRGB @packed
```
lerno, line 423: `@packed` here should do nothing, since the alignment is 1 anyway.

lib/std/compression/qoi.c3
Hema2-official: In non-optimized runs, they decrease performance. It's not a significant hit though.
But I also want to keep the implementation closer to the reference document...
I'd leave it as it is, if that's okay.

lib/std/compression/qoi.c3
Hema2-official: I used the attribute for some kind of consistency, I think.
I'll replace it with an explanation comment.
Thanks!

lib/std/compression/qoi.c3
Hema2-official: Thanks!

lib/std/compression/qoi.c3
Hema2-official: Nah, I've done it now. I wanted to do it anyways.
Thanks for reminding me!

test/unit/stdlib/compression/qoi.c3
```diff
@@ -0,0 +1,56 @@
...
+        if (catch f = decoded) {
+            abort("qoi: decode failed: %s", f.nameof);
+        }
+
+        assert(test_desc.width == 340 && test_desc.height == 169, "qoi: image dimensions mismatch");
+
+        // encode the decoded data
+        char[]! encoded = qoi::encode(decoded, &test_desc);
+        if (catch f = encoded) {
+            abort("qoi: encode failed: %s", f.nameof);
```
lerno, line 28: Don't use abort, use assert and no printing to stdout

lerno: Or even better just throw:
`char[] encoded = qoi::encode(decoded, &test_desc)!;`
If the test function is `void!` then returning a fault here would be a failed test already.

lerno:
One final thing: add QOI to the release notes!

PR #1412, opened 2024-09-04
Add paramnames to compile time type info for fn types

Dodzey:
This adds paramnames to compile time type info for fn types.
This works in tandem with params, which gives you the type ids.


lerno:
So I've been thinking about this a lot. It's not great to have multiple functions, so the question is if it wouldn't be better to return an untyped list with { type, name }. Now the downside of this would be that you'd index it like $param[0] => type $param[1] => name, and this is kind of opaque. I'll bring this up on the discord.

lerno:
I've replaced .params and .paramnames now with `.paramsof` which returns a `ReflectedParam` which is a struct in `runtime.c3` you can use .name to get the name and .type to get the typeid.

PR #1414, opened 2024-09-06
Add CMake option C3_ENABLE_CLANGD_LSP

id3nom:
Added a CMake option to be able to use Clangd LSP with neovim and vscode.


lerno:
Thank you

PR #1430, opened 2024-09-11
Add initial FreeBSD support

wilsonk:
This seems to build the compiler on FreeBSD fine. Then all of the regression tests pass for the unit tests, and most of the stdlib tests pass, with a few errors. It is a start, at least. 

Not tested on Net/Open BSD's at this point. The demo1 and diagnostic testfragments seemed to need a change also...not sure if this is FreeBSD specific or not?

### DIFF
--- a/lib/std/core/env.c3
+++ b/lib/std/core/env.c3
@@ -146,6 +146,7 @@ const bool POSIX = LIBC && os_is_posix();
 const bool OPENBSD = LIBC && OS_TYPE == OPENBSD;
 const bool FREEBSD = LIBC && OS_TYPE == FREEBSD;
 const bool NETBSD = LIBC && OS_TYPE == NETBSD;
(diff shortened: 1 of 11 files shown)


lerno:
Rather than having the `BSDS` const in fileinfo, move it to env.c3 and add a `const bool BSD_FAMILY` in the same way, which then can be used like `env::BSD_FAMILY`

lerno:
And was dirent wrong? I think I checked against the netbsd and openbsd headers?

wilsonk:
Ok, I made the changes for the BSD_FAMILY in env.c3 (only saw the os/env.c3 and not that second file in core...duh). Makes more sense now.

I have updated the dirent struct to the most current versions of each BSD. I think I was looking at an older github version before from a third party...but this should be correct now. It may be different for older versions of the BSD's, but I assume we would want to target the newest version unless someone needs older compatibility.

lerno:
Thank you!

PR #1433, opened 2024-09-12
ability to disable llvm at compile time

alex-s168:
adds the `C3_WITH_LLVM` option (default to true) which can be set to off to disable LLVM.
this is useful for when you want to use other backends or do some backend unrelated testing and don't have LLVM on your computer (when building)


lerno:
Thank you! I'll fix up the formatting a bit so that it conforms to coding standards, so don't worry about that. But for next time have a look?

PR #1435, opened 2024-09-12
Bsd family fixes

wilsonk:
This gets NetBSD (and I assume OpenBSD, but I didn't check yet) up to about the same point as FreeBSD. This means all the regression tests pass and most of the stdlib tests pass except one memory test and then two pthread tests fail and the runner crashes. Still, most of the tests are passing, at least (and most of the resources/examples that I tried).


lib/std/libc/libc.c3
```diff
@@ -202,7 +202,30 @@ macro CFile stdin() => __stdin;
 macro CFile stdout() => __stdout;
 macro CFile stderr() => __stderr;
 
-module libc @if(env::DARWIN || env::BSD_FAMILY);
+module libc @if(env::FREEBSD);
+extern CFile __stdinp;
+extern CFile __stdoutp;
+extern CFile __stderrp;
```
lerno, line 208: Shouldn't this be shared with Darwin?

wilsonk: Changed. Didn't look too closely after extracting the NetBSD/OpenBSD specific code.

lerno:
Thank you!

Issue #1437, opened 2024-09-12
replace actions/create-release with another action


Real-Packet:
`actions/create-release` is archived and deprecated. We should replace it with something else.
**EDIT**:
> This would be a good replacement: https://github.com/softprops/action-gh-release.
> **EDIT**:
> > Can't test it because stupid GitHub decided it would be a great idea to complete every other workflow job except release.

Real-Packet:
this would be a good replacement: https://github.com/softprops/action-gh-release

Real-Packet:
can't test it because stupid GitHub decided it would be a great idea to complete every other workflow job except release.

Real-Packet:
> `actions/create-release` is archived and deprecated. We should replace it with something else. **EDIT**:
> 
> > This would be a good replacement: https://github.com/softprops/action-gh-release.
> > **EDIT**:
> > > Can't test it because stupid GitHub decided it would be a great idea to complete every other workflow job except release.

the reason why it wasn't running is because of https://github.com/c3lang/c3c/blob/6b2ce6de6f4f62bb239645f3fb9378af8968da8b/.github/workflows/main.yml#L666-L669

PR #1438, opened 2024-09-13
chore: migrate to softprops/action-gh-release (untested)

Real-Packet:
Hopefully this works.
Fix c3lang/c3c#1437


lerno:
Crossing my fingers here :D

Issue #1422, opened 2024-09-09
Vendor-fetch download to lib directory specified in project.json


brian-sinquin:
By default, `c3c vendor-fetch` downloads the libraries at the root of the project. Here are some proposition :

 * The command could download in the folder  specified (first entry of `dependency-search-paths` for exemple, `lib/` by default) in the `project.json` file 
 * Add the libraries as global dependencies ( `dependencies` entry) by default.
 * Specifying a target to which add the dependencies with an additional argument. 
 * ~~Download location could also be changed with an additional argument~~. Already done with `--path` argument.

lerno:
In addition I think there should be a "project dependency fetch" which will look at the dependencies in a project and retrieve the ones missing from the dependency search paths.

Can we add this as well?

Feel free to work with the first part already. The code style guide here https://github.com/c3lang/c3c/blob/master/CODESTYLE.md can be good to review before making the PR.

There are already a set of project manipulation sub commands, this one should probably join that list of commands.

brian-sinquin:
> In addition I think there should be a "project dependency fetch" which will look at the dependencies in a project and retrieve the ones missing from the dependency search paths.
> 
> Can we add this as well?
> 
> Feel free to work with the first part already. The code style guide here https://github.com/c3lang/c3c/blob/master/CODESTYLE.md can be good to review before making the PR.
> 
> There are already a set of project manipulation sub commands, this one should probably join that list of commands.

Working on that, PR soon ;)

lerno:
This seems to have completed the issue! Thank you.

PR #1441, opened 2024-09-13
Vendor-fetch download to lib directory specified in project.json (#1422)

brian-sinquin:
This PR refers to the issue #1422 
* [x] vendor-fetch should download to `dependency-search-path` field set in `project.json` by default
* [x] vendor-fetch should add libraries in the target library list in `project.json` (~~with an option to specify a target~~)
    * [ ] Proper JSON file update (`dependencies` field of root or specified target) after fetching
    * [ ] Create `dependencies` field if it does not exist
* [ ] Fetch missing libraries that are specified in `project.json`

### DIFF
--- a/src/compiler/compiler.c
+++ b/src/compiler/compiler.c
@@ -3,6 +3,7 @@
 // a copy of which can be found in the LICENSE file.
 
 #include "compiler_internal.h"
+#include "../build/project.h"
 #include <compiler_tests/benchmark.h>
 #include "../utils/whereami.h"
 #if PLATFORM_POSIX
@@ -848,23 +849,43 @@ static void setup_bool_define(const char *id, bool value)
 	}
 }
 #if FETCH_AVAILABLE
+
+const char * vendor_fetch_single(const char* lib, const char* path) 
+{
+	const char *resource = str_printf("/c3lang/vendor/releases/download/latest/%s.c3l", lib);
+	const char *destination = file_append_path(path, str_printf("%s.c3l", lib));
+	const char *error = download_file("https://github.com", resource, destination);
+	return error;	
+}
+
 void vendor_fetch(BuildOptions *options)
 {
 	unsigned count = 0;
+	if (options->path == DEFAULT_PATH)
+	{
+		const char** lib_dirs =  proj_lib_dirs_get();
+		int num_lib = vec_size(lib_dirs); 
+		if (num_lib > 0) options->vendor_download_path = lib_dirs[0];
+
+	}
 	FOREACH(const char *, lib, options->libraries_to_fetch)
 	{
-		const char *resource = str_printf("/c3lang/vendor/releases/download/latest/%s.c3l", lib);
+		//TODO : Implement progress bar in the download_file function.
 		printf("Fetching library '%s'...", lib);
 		fflush(stdout);
-		const char *error = download_file("https://github.com", resource, str_printf("%s.c3l", lib));
+
+		const char *error = vendor_fetch_single(lib, options->vendor_download_path);
+
 		if (!error)
 		{
-			puts("ok.");
+			puts("finished.");
+			//TODO! add_library_to_project
+			// add option options->vendor-target for specific dependency
 			count++;
 		}
 		else
 		{
-			printf("FAILED: '%s'\n", error);
+			printf("Failed: '%s'\n", error);
 		}
 	}
 	if (count == 0)	error_exit("Error: Failed to download any libraries.");
--- a/src/compiler/compiler.c
+++ b/src/compiler/compiler.c
@@ -867,9 +867,9 @@ void vendor_fetch(BuildOptions *options)
 		if (file_exists(PROJECT_JSON5) || file_exists(PROJECT_JSON))
 		{
 
-			const char** lib_dirs =  proj_lib_dirs_get();
-			int num_lib = vec_size(lib_dirs); 
-			if (num_lib > 0) options->vendor_download_path = lib_dirs[0];
+			const char** deps_dirs =  get_project_dependency_directories();
+			int num_lib = vec_size(deps_dirs); 
+			if (num_lib > 0) options->vendor_download_path = deps_dirs[0];
 
 		}
 
@@ -896,8 +896,9 @@ void vendor_fetch(BuildOptions *options)
 		}
 	}
 	
-	//TODO! add_library_to_project
-	add_libraries_project(fetched_libraries, options->project_options.target_name);
+	// add fetched library to the dependency list
+	add_libraries_to_project_file(fetched_libraries, options->project_options.target_name);
+
 	if (count == 0)	error_exit("Error: Failed to download any libraries.");
 	if (count < vec_size(options->libraries_to_fetch)) error_exit("Error: Only some libraries were downloaded.");
 }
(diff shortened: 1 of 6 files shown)


brian-sinquin:
For now, i worked on :

* The default vendor-fetch download path is set using the first dependency-path specified in the project JSON file.
* This behavior is skipped if --path is specified (vendor-fetch in the specified path)
* Adds the missing libraries in the global dependency entry of the project JSON file

I think i need a review at that point and think about refining the code, and making it more versatile.

lerno:
Just let me know when you need the review.

brian-sinquin:
Test should be running now, i fixed the case where my code does not find the project JSON file. Please review when you can, then i will add the fetching of missing dependencies. Also, the json API works well for reading but is lacks some helper functions for editing.

lerno:
It looks like there are some issues to address

brian-sinquin:
> It looks like there are some issues to address

I manage to compile well on my computer, i updated const types that were not before.
I face an edge case in my JSON project file edition (to add fetched libraries), all empty arrays of the file are modified while i want to only alter the `dependencies` entry. Something may wring in my json object manipulation but i dont see where.

lerno:
Things that have to be fixed for it to compile on all compilers:
```
D:\a\c3c\c3c\src\compiler\compiler.c:878:20: error: result of comparison against a string literal is unspecified (use an explicit string comparison function instead) [-Werror,-Wstring-compare]
  878 |         if (options->path == DEFAULT_PATH)
      |                           ^  ~~~~~~~~~~~~
D:\a\c3c\c3c\src\compiler\compiler.c:907:4: error: assigning to 'char *' from 'const char *' discards qualifiers [-Werror,-Wincompatible-pointer-types-discards-qualifiers]
  907 |                         vec_add(fetched_libraries, lib);
      |                         ^                          ~~~
D:\a\c3c\c3c\src\compiler/../utils/lib.h:339:29: note: expanded from macro 'vec_add'
  339 |         (vec_)[vec_size(vec_) - 1] = value_; \
      |                                    ^ ~~~~~~
D:\a\c3c\c3c\src\compiler\compiler.c:917:32: error: passing 'char **' to parameter of type 'const char **' discards qualifiers in nested pointer types [-Werror,-Wincompatible-pointer-types-discards-qualifiers]
  917 |         add_libraries_to_project_file(fetched_libraries, options->project_options.target_name);
      |                                       ^~~~~~~~~~~~~~~~~
D:\a\c3c\c3c\src\compiler/../build/project.h:6:49: note: passing argument to parameter 'libs' here
    6 | void add_libraries_to_project_file(const char** libs, const char* target_name);
      |        
```

lerno:
One CI doesn't work because vendor-fetch now requires an argument:
```
Run build\Release\c3c.exe vendor-fetch raylib5
  build\Release\c3c.exe vendor-fetch raylib5
  shell: C:\Windows\system32\cmd.EXE /D /E:ON /V:OFF /S /C "CALL "{0}""
  env:
    LLVM_RELEASE_VERSION_WINDOWS: 18
    LLVM_RELEASE_VERSION_MAC: 17
    LLVM_RELEASE_VERSION_LINUX: 17
    LLVM_RELEASE_VERSION_UBUNTU20: 17
    LLVM_DEV_VERSION: 20
No project file found, use --path to specify custom destination.
```
But this looks wrong: it should be possible to use vendor-fetch without a project?

brian-sinquin:
> One CI doesn't work because vendor-fetch now requires an argument:
> 
> ```
> Run build\Release\c3c.exe vendor-fetch raylib5
>   build\Release\c3c.exe vendor-fetch raylib5
>   shell: C:\Windows\system32\cmd.EXE /D /E:ON /V:OFF /S /C "CALL "{0}""
>   env:
>     LLVM_RELEASE_VERSION_WINDOWS: 18
>     LLVM_RELEASE_VERSION_MAC: 17
>     LLVM_RELEASE_VERSION_LINUX: 17
>     LLVM_RELEASE_VERSION_UBUNTU20: 17
>     LLVM_DEV_VERSION: 20
> No project file found, use --path to specify custom destination.
> ```
> 
> But this looks wrong: it should be possible to use vendor-fetch without a project?

You are right. It is fixed, it will download in the default path (cwd) if it does not find a project file.

src/compiler/compiler.c
```diff
@@ -861,7 +861,7 @@ const char * vendor_fetch_single(const char* lib, const char* path)
 void vendor_fetch(BuildOptions *options)
 {
 	unsigned count = 0;
-	if (options->path == DEFAULT_PATH)
+	if (strcmp(options->path, DEFAULT_PATH) == 0)
```
lerno, line 864: There is a str_eq which just returns a bool, could you use that one instead?

brian-sinquin: done.

lerno: It's not a show-stopper though. I can do it later.

brian-sinquin:
So far i added the two first features of the initial PR.

Missing features : 
- Fetch a library that is needed but not present in the search path
- Add fetched libraries in the `dependencies` member of a specified target (JSON root for the moment)

Know bugs/edge cases :

- No checks / correction if the "depencies", "dependencies-search-path" ... do not exists
- Something is wrong with my JSON editing code (populating wrong array entries if they are empty)

lerno:
The JSON bug seems worrying?

brian-sinquin:
> The JSON bug seems worrying?

Let's say you begin with no search-path : 
```json
{
  "langrev": "1",
  "warnings":  [ "no-unused" ],
  "dependency-search-paths":  [ ],
  "dependencies":  [ ],
  "authors":  [ "John Doe <john.doe@example.com>" ],
  "version": "0.1.0",
  "sources":  [ "src/**" ],
  "output": "build",
  "targets": {
    "myAAAGame": {
      "type": "executable",
    },
  },
  "cpu": "generic",
  "opt": "O0",
}
```

After `c3c vendor-fetch raylib5` you will end up with : 
```json
{
  "langrev": "1",
  "warnings":  [ "no-unused" ],
  "dependency-search-paths":  [ "raylib5" ],
  "dependencies":  [ "raylib" ],
  "authors":  [ "John Doe <john.doe@example.com>" ],
  "version": "0.1.0",
  "sources":  [ "src/**" ],
  "output": "build",
  "targets": {
    "myAAAGame": {
      "type": "executable",
    },
  },
  "cpu": "generic",
  "opt": "O0",
}
```
The compilator will search for `./raylib5/raylib5.c3l` while it is actualy at `./raylib5/raylib5.c3l`.
I dont know why the `dependencies-search` gets populated, i may not use the JSONObject properly.

lerno:
Ok I know why. First of all, the JSON object were never intended to be manipulated – to just be read-only structure. This creates lots of problems when it's mutated. In this case, an empty array is reused. So you're actually reusing the single empty array in the JSON structure!

lerno:
The JSON stuff really needs to be brought in to use the general C3 allocator.

lerno:
Thank you!

brian-sinquin:
> Ok I know why. First of all, the JSON object were never intended to be manipulated – to just be read-only structure. This creates lots of problems when it's mutated. In this case, an empty array is reused. So you're actually reusing the single empty array in the JSON structure!

Yeah, I saw it was more a read API but I tried to use the abstract object. Thank you for the fix

lerno:
I updated it further, it should be straightforward to manipulate now.

PR #1446, opened 2024-09-14
CI: Update actions/*-artifact to v4

Real-Packet:
Because v3 is being deprecated and uses Node 16 instead of Node 20. v4 is also a drop in replacement so we can just find & replace the usages without any changes to easily upgrade to v4.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -91,7 +91,7 @@ jobs:
           dir msvc_sdk
 
       - name: upload artifacts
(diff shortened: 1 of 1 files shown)


Real-Packet:
Based on the current CI results (19 in progress, 15 successful), this probably works.

Real-Packet:
5 years later: The MSVC Debug check has completed

lerno:
Thank you!

PR #1449, opened 2024-09-16
fix install instructions on ubuntu's latest LTS

joshring:
fix install instructions on ubuntu's latest LTS, using LLVM 17 does not trivially produce an executable and LLVM 18 is what I would recommend as it works out of the box.

LLVM 17 will work with additional work to the system or install script, since clang is at a different path for clang-17 as it's the non-default version and does not have a symlink to `/bin/clang`, instead is installed at: `/bin/clang-17`

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,6 +205,12 @@ More platforms will be supported in the future.
 2. Unpack executable and standard lib.
 3. Run `./c3c`.
 
+#### Installing on Ubuntu with precompiled binaries
+1. Download tar file: [https://github.com/c3lang/c3c/releases/download/latest/c3-ubuntu-20.tar.gz](https://github.com/c3lang/c3c/releases/download/latest/c3-ubuntu-20.tar.gz)
+   (debug version [here](https://github.com/c3lang/c3c/releases/download/latest/c3-ubuntu-20-debug.tar.gz))
+2. Unpack executable and standard lib.
+3. Run `./c3c`.
+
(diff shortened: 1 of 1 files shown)


lerno:
c3c stopped supporting LLVM 16, so maybe there is need for new instructions for Arch too?

Caleb-o:
I'm currently using LLVM 18.x on Arch for C3 atm

lerno:
So could the instructions be updated with that?

Caleb-o:
To use an 18.x version? It would seem that might be the best option if Josh is also saying that 18 is better off. I haven't had any issue using 18.

joshring:
LLVM 18 is best on Ubuntu 24.04 yeah, because it's installed at a path which is detected easily and sets the correct symlinks by default.

If it works out of the box with either LLVM 17 or 18, maybe provide a command for either?

lerno:
@joshring Will you update it?

joshring:
Tried to test on Docker without luck, SadeAli confirms that those steps are still working correctly and install with LLVM 17. So I'll leave Arch for now

lerno:
But this is what it says:

> Due to some issues with the LLVM packaged for Arch Linux, the AUR package will download and use LLVM 16 from Ubuntu-23.04 to compile the c3c compiler.

Note "LLVM 16"

joshring:
Sorted, thanks for clarifying I was a bit lost

lerno:
Thank you!

PR #1469, opened 2024-09-20
Project fetch missing libraries

brian-sinquin:
Additional feature to #1422 

> c3c project fetch

Search for dependencies used in the project file and fetch those that are not present in the project tree.
Works both on root dependencies and target specific dependencies.


brian-sinquin:
Done.
It should compile on every machine and work only on windows for the moment because of cURL.

src/build/build_options.c
```diff
@@ -187,6 +188,7 @@ static void project_usage()
 	PRINTF("Project Subcommands:");
 	PRINTF("  view                                          view the current projects structure");
 	PRINTF("  add-target <name>  <target_type>              add a new target to the project");
+	PRINTF("  fetch              							fetch missing project libraries");
```
lerno, line 191: This should only be visible on `FETCH_AVAILABLE` right?

brian-sinquin: Yep

lerno: But it looks like it's printed even if FETCH_AVAILABLE isn't set?

src/build/project_manipulation.c
```diff
@@ -252,6 +267,95 @@ static void view_target(const char *filename, const char *name, JSONObject *targ
...
+
+	// dependency check tree
+	while (vec_size(deps) > 0)
+	{
+	
+		FOREACH(const char*, dir, libdirs)
+		{
+			
+			const char* dep = deps[vec_size(deps)-1];
+			if(file_exists(file_append_path(dir, str_printf("%s.c3l", dep))))
```
lerno, line 326: Nitpicking: space between `if` and `(`

lerno:
Oh, and please add this to the release notes as well, including a not of the other feature you implemented.

src/build/build_options.c
brian-sinquin: Yes I forgot this, I'll fix it in the next hours

brian-sinquin:
> Oh, and please add this to the release notes as well, including a not of the other feature you implemented.

Done

lerno:
Thank you!

PR #1482, opened 2024-09-25
add x86 APX features

alex-s168:
APX support was recently merged into LLVM, making C3C output warning messages, because it doesn't know these features.
Note that I unified the LLVM APX features in the standard library, because they are all on the same feature bit, and it doesn't make any sense to split them (like in LLVM).


lerno:
Thank you!

PR #1494, opened 2024-09-28
Fix incorrect `to_gmt_offset` result

hkalexling:
There was a bug in `to_gmt_offset` where conversion between two non UTC timezones would give incorrect results. This PR fixes it.


lerno:
Merged!

PR #1498, opened 2024-09-29
Fix escape sequence handling in encoding::json

chri-k:


lerno:
Thank you for finding this.

lerno:
I added release notes and tests for this one.

PR #1509, opened 2024-09-30
Made shebangs be ignored

ellipse12:
shebangs are now treated as comments. This allows single file scripts to be made using the `compile-run` command.
Example:
```c3
#!/bin/c3c compile-run
import std::io;
fn int main(String args)
{
    io::printn("Hello, World!");
    return 0;
}
```
It does **not** make `#` into comments only `#!`

### DIFF
--- a/src/compiler/lexer.c
+++ b/src/compiler/lexer.c
@@ -250,6 +250,14 @@ static void skip_whitespace(Lexer *lexer)
 	{
 		switch (peek(lexer))
 		{
+            //Case for shebang line #!
+            case '#':
+                if(peek_next(lexer) == '!')
+                {
+                    skip(lexer, 2);
+                    parse_line_comment(lexer);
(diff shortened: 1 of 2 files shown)


lerno:
I've thought about that feature, but decided not to include it.

sa-tasche:
Also for files in folder `scripts`?

lerno:
How do you mean?

ellipse12:
@lerno Sorry I didn't realize you had already thought about it, do you mind sharing your reservations?

lerno:
Originally #! was smart, because comments started with `#` in many scripting languages. I know D adds this, but I feel it would be something that would be too unix-y to properly be in the language.

ellipse12:
Oh ok that makes sense, though I might keep it in my own fork if that's ok, I see myself using it fairly often.

PR #1510, opened 2024-10-01
Add variants of `DString.insert_at` to match `.append`

chri-k:
Adds more methods to insert things in the middle of a DString:
```
insert_string_at(usz, DString)
insert_char_at(usz, char)
insert_char32_at(usz, Char32)
insert_utf32_at(usz, Char32[])
```
And to match `.append`, i renamed `.insert_at(usz, String)` to `insert_chars_at`, and instead made `.insert_at` a macro that picks a function based on type, like `.append` (it will pick the renamed one for `String` so nothing should break).

I also updated `append_utf32` and `append_char32`  to now return the size of the utf8 they appended for convenience (and of course the insert versions do the same). That shouldn't affect anything either

### DIFF
--- a/lib/std/core/dstring.c3
+++ b/lib/std/core/dstring.c3
@@ -154,13 +154,15 @@ fn String DString.str_view(self)
 	return (String)data.chars[:data.len];
 }
 
-fn void DString.append_utf32(&self, Char32[] chars)
+fn usz DString.append_utf32(&self, Char32[] chars)
 {
 	self.reserve(chars.len);
+	usz end = self.len();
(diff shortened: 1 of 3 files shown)


chri-k:
And of course i somehow managed to include six lines of space indentation in the PR, despite already having undone the commit once to fix those exact same six lines of space indentation.

lerno:
Can you add something in releasenotes about this as well please?

chri-k:
My bad. Added line to release notes (and fixed the spaces)

lerno:
Thank you!

PR #1521, opened 2024-10-05
fix int formatting in std::collections::object

Renerick:
Reported at https://discord.com/channels/650345951868747808/650347464498610189/1291898838772088832

> When the value is a signed integer, it's still formatted as unsigned integer:
> ```
> io::printn(b_value.type.kindof);
> io::printn(b_value);
> io::printn((int128)b_value.i);
> ```

> ```
> SIGNED_INT
> 340282366920938463463374607431768211414
> -42
> ```


lerno:
Ideally there is also a note in the releasenotes for this, but I'm going to accept this as is. But something to think about in the future.

lerno:
Thank you!

PR #1531, opened 2024-10-08
fix(string): use heap allocator for ZString.copy

konimarti:
Use the heap allocator for ZString.copy() instead of the temp allocator to stay consistent across the code base.

The temp allocator is used for ZString.tcopy().


lerno:
This is a bug actually, it should definitely be the heap allocator.

lerno:
Thank you!

PR #1543, opened 2024-10-10
fix(csv): read_new_row uses heap alloc / update @each_row

fernandolguevara:


lerno:
Ah, darn, I was working on adding to this one.

PR #1550, opened 2024-10-12
Small fixes to AST json

hwchen:
Some small fixes to AST json.

```
    c3c compile -P ~/c3dev/c3c/lib/std | jq
```
now works.

Besides some basic formatting, one commit is also a fix to correctly print types for struct/union fields.

### DIFF
--- a/src/compiler/json_output.c
+++ b/src/compiler/json_output.c
@@ -81,10 +81,10 @@ static inline void emit_type_data(FILE *file, Module *module, Decl *type)
 			PRINTF("\t\t\t\t{\n");
 			if (member->name)
 			{
-				PRINTF("\t\t\t\t\t\"name:\": \"%s\",\n", member->name);
+				PRINTF("\t\t\t\t\t\"name\": \"%s\",\n", member->name);
 			}
 			// TODO, extend this
-			PRINTF("\t\t\t\t\t\"type:\": \"%s\"\n", type->name);
+			PRINTF("\t\t\t\t\t\"type\": \"%s\"\n", type->name);
 			PRINTF("\t\t\t\t}");
 		}
 		fputs("\n\t\t\t]", file);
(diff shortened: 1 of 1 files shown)


src/compiler/json_output.c
```diff
@@ -191,8 +191,8 @@ static inline void emit_func_data(FILE *file, Module *module, Decl *func)
 	fputs("\t\t\t\"params\": [\n", file);
 	FOREACH_IDX(i, Decl *, decl, func->func_decl.signature.params)
 	{
-		if (i != 0) fputs(",\n", file);
```
hwchen, line 194: having this comma before the `decl` check means that a trailing comma can apear.

src/compiler/json_output.c
```diff
@@ -255,7 +255,7 @@ static inline void emit_globals(FILE *file)
 					PRINTF("\t\t\"%s::%s\"", module->name->module, decl->name);
 		FOREACH_DECL_END;
 	}
-	PRINTF("\n\t],");
+	PRINTF("\n\t]");
```
hwchen, line 258: This comma, and the one at 244, are dependent on the ordering of the globals/types sections. I just made a hardcoded fix here, but if the order of the sections changes the commas will be wrong again.

Probably the easiest better fix would be to handle commas between calls to each section in `emit_json_to_file`, let me know your preference.

src/compiler/json_output.c
```diff
@@ -182,6 +159,32 @@ void print_type(FILE *file, TypeInfo *type)
...
+		{
+			if (i != 0) fputs(",\n", file);
+			PRINTF("\t\t\t\t{\n");
+			if (member->name)
+			{
+				PRINTF("\t\t\t\t\t\"name\": \"%s\",\n", member->name);
+			}
+			// TODO, extend this
+			PRINTF("\t\t\t\t\t\"type\": \"");
+			print_type(file, type_infoptr(member->var.type_info));
```
hwchen, line 180: Sorry for the horrible diff.

Basically `print_type` was moved above `emit_type_data`, and then `print_type`(on this line) is now used to print the type of the struct field.

This commit can also be broken out into another PR if not clear.

src/compiler/json_output.c
```diff
@@ -182,6 +159,39 @@ void print_type(FILE *file, TypeInfo *type)
...
+			PRINTF("\t\t\t\t{\n");
+			if (member->name)
+			{
+				PRINTF("\t\t\t\t\t\"name\": \"%s\",\n", member->name);
+			}
+			// TODO, extend this
+			PRINTF("\t\t\t\t\t\"type\": \"");
+			if (member->var.type_info)
+			{
+				print_type(file, type_infoptr(member->var.type_info));
```
hwchen, line 182: Sorry for the horrible diff.

Basically print_type was moved above emit_type_data, and then print_type(on this line) is now used to print the type of the struct field.

This commit (fixing type printing for struct fields) can also be broken out into another PR if not clear.

lerno:
Looks good, thanks!

PR #1552, opened 2024-10-12
fix: segfault in panic when statically compiled

PavelBlinnikov:
In the current state of c3c panic in statically compiled binary leads to segfault.

Test code:
```c
module test;

fn int main(String[] args)
{
    std::core::builtin::panic("test1", "test2", "test3", 31337);
    return 0;
}
```

Compiled with:
```bash
c3c build -z "-static"
```

Reason: segfault happens because dlclose unloads libc.so.6 in static binaries and funcptr given by dlsym is not valid anymore.
https://github.com/c3lang/c3c/blob/e6c9cfed42294849bcfd808ff6cd13348dc33c25/lib/std/os/posix/process.c3#L64-L73

Solution: add RTLD_NODELETE to dlopen which does not allow unloading libc.so.6 in dlclose
```c
void* handle = libc::dlopen("libc.so.6", libc::RTLD_LAZY|libc::RTLD_NODELETE);
```

### DIFF
--- a/lib/std/libc/os/posix.c3
+++ b/lib/std/libc/os/posix.c3
@@ -11,6 +11,7 @@ const int RTLD_LAZY = 0x1;
 const int RTLD_NOW = 0x2;
 const int RTLD_LOCAL = 0x4;
 const int RTLD_GLOBAL = 0x8;
+const int RTLD_NODELETE = 0x1000;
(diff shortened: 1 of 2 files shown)


lerno:
Thank you!

PR #1558, opened 2024-10-14
chore(cmake): improve lld resolution on apple

fernandolguevara:


lerno:
The recent changes in brew for LLVM just pissed me the frack off. If the maintainer is so careless as to do this kind of changes, then we shouldn't ever consider homebrew being a good place to get the LLVM libs from. This is just MINDBLOWINGLY bad. It's not the first time Homebrew messes up the LLVM libs, but this time it really reached the end of my patience. I appreciate you looking into this, but we need a separate CI that just builds LLVM for MacOS and then we only depend on that.

This is not the only problem with the Homebrew install either. The LLVM static libraries *are dependent on Z3 being installed through Homebrew*. Yes, this is just as moronic as it sounds: you can't use LLVM static libraries as static libraries because the LLVM homebrew maintainers decided it was a good idea to hard-code a dynamic library dependency on a Homebrew path!

fernandolguevara:
I understand your position on the situation with LLVM distributed by homebrew... but let's make our lives easier :)

lerno:
Alright!

Issue #1541, opened 2024-10-10
AST JSON: Globals are only qualified names


Caleb-o:
Currently, globals are just qualified names in an array, but don't show anything with the expression for their assignment or their type.

![image](https://github.com/user-attachments/assets/1fe15de8-a400-4f73-8c8f-d783d169eea0)

lerno:
We should rework this anyway. Closing it for a general specification later.

Issue #1540, opened 2024-10-10
AST JSON: Constants are not present


Caleb-o:
Constants seem to not be emitted into the JSON at all, when they should be.

Cardisk:
Hi, new here. I was trying to solve this issue and I also added the type to the globals as well. I will try to add also the init value. Is it ok if I collapse the two issues in only one?
By the way this is my first contribution ever, there are any instructions on how to contribute to this project beside the code of conduct and the codestyle files?

lerno:
Not really, just rebase the commit so that it's a single commit. And add a line in the release notes for the change.

lerno:
Is this fixed?

cardisk:
Yes. There are a few expressions that I really cannot trigger so I wasn't able to output a specific value for them, instead I outputted a `TODO:` with the corresponding type so in the future it can be easy to implement them. At the moment there are no TODOs inside the standard library.
If this issue can be closed I think also the #1541 one should, because inside the AST constants and globals have the same output.

lerno:
Also closing this, as we're reworking what it should contain, see #2349

PR #1564, opened 2024-10-17
Json AST: Adding 'constants' and Improving 'globals' #1540, #1541

cardisk:
Adding constants and improving globals inside the Json AST with information about types and values.

### DIFF
--- a/releasenotes.md
+++ b/releasenotes.md
@@ -16,6 +16,8 @@
 - Improve error messages on expressions like `var $type = int;` #1553.
 - Disallow casting a `void*` to `any` or an interface, unless it is `null`.
 - Defer resolution of declarations when looked up in `def` aliased #1559.
+- Adding constants to the Json AST #1540
+- Adding info to the globals inside Json AST #1541
 
 ### Fixes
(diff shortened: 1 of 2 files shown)


cardisk:
I fixed the problem that CI reported. I was not aware of the size difference of uint64_t on different operating systems.

lerno:
You mean size of long?

cardisk:
Yesterday I roughly fixed the printf problem. The first time, CI failed on linux distros because of the wrong format used (%lld) for uint64_t. I never worked with 64 bits wide types and I was short on free time so I changed the code adding a check on the operating system.
Today, after a bit more research, I found PRIu64 macro to print this type. Is it better to use such macro, or should I leave it like it is now?

lerno:
I think the simplest thing is simply to cast the value to unsigned long long before passing it to printf, and using "%llu"

cardisk:
I've done the change, now it should be ok

lerno:
While I'm not sure it's right to have the expressions in the AST, let's try it out. Thank you for doing this.

Issue #1560, opened 2024-10-15
`$exec` fails when running c3 script with arguments


Book-reader:
in the below example:
```ts
//src/main.c3
module exec_test;
import std::io;

$exec("test.c3", {"hello_string"});

fn int main(String[] args)
{
  io::printn(hello_string);
  return 0;
}

// scrips/test.c3
module exec_test::script;

import std::io;

fn void main(String[] args)
{
  for (int i = 1; i < args.len; i++)
  {
    io::printfn("String %s = \"Hello, World!\";", args[i]);
  }
}
```
running it will result in the error:
```
sh: line 1: hello_string/home/user/code/c3/c3-beta/c3/c3c: No such file or directory
Failed to compile script 'test.c3'.
```
and adding a second argument to the `$exec`:
`$exec("test.c3", {"hello_string", "second_string"})`
results in the error:
```
sh: line 1: hello_string: command not found
Failed to compile script 'test.c3'.
```
running `$exec("test.c3", {})` does not cause an error except for the expected `hello_string not found`

running `$exec("echo", {"'String hello_string = \"Hello, World!\";'"})` also does not cause an error

lerno:
It seems like @zargovv figured this out in #1568, but it was a bit incomplete. I think the change I just pushed addressed that last thing.

Book-reader:
Looks good, thanks!

PR #1568, opened 2024-10-23
Clear `scratch_buffer` in `compile_and_invoke`.

zargovv:
The `scratch_buffer` contains erroneous data in `compile_and_invoke`, therefore should be cleared.

### DIFF
--- a/src/compiler/compiler.c
+++ b/src/compiler/compiler.c
@@ -1517,6 +1517,8 @@ void scratch_buffer_append_native_safe_path(const char *data, int len)
 
 File *compile_and_invoke(const char *file, const char *args, const char *stdin_data)
 {
+  scratch_buffer_clear();
+
 	char *name;
 	if (!file_namesplit(compiler_exe_name, &name, NULL))
 	{


lerno:
This fix would break `execute_scripts()`. This line:

```c
(void) compile_and_invoke(scratch_buffer_to_string(), execs.len ? execs.ptr : "", NULL);
```
Needs to be
```c
(void) compile_and_invoke(scratch_buffer_copy(), execs.len ? execs.ptr : "", NULL);
```

lerno:
This fix addresses #1560 yes, but let's not use "closes" since that would auto-close 1560, which is not desirable. Once it is fixed, we request the reporter to check that the issue is properly fixed instead. This is an extra check.

lerno:
I incorporated it into a fix, together with the additional lines needed. Thanks for tracking it down!

zargovv:
> This fix addresses #1560 yes, but let's not use "closes" since that would auto-close 1560, which is not desirable. Once it is fixed, we request the reporter to check that the issue is properly fixed instead. This is an extra check.

Sorry for the inconvenience caused by the pull request.

I'll keep that in mind for any future submissions. Thanks for the review!

lerno:
No it's fine, no worries. Thank you for finding the issue!

PR #1571, opened 2024-10-25
Update .editorconfig to reflect current style

hwchen:
For `C3` and `C` files, change from spaces to tabs (and remove hardcoded `indent_size` to allow readers to view tabs at preferred width).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,12 +11,10 @@ indent_style = space
 indent_size = 4
 
 [*.{c,cc,h}]
-indent_style = space
-indent_size = 4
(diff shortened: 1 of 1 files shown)


hwchen:
I noticed there's two Python files:

```
msvc_build_libraries.py
test/src/tester.py
```

and each has different indentation: the first is 2 spaces, the second is 4. If you have a preference for one or the other, I can reformat and update the `.editorconfig`.

lerno:
They should probably be tabs, to make it consistent.

lerno:
Thank you!

PR #1578, opened 2024-10-27
Remove requirements for pop_first, cause it already returns fault.

MohMaGen:
```pop``` hasn't this requirements and works fine. But ```pop_first``` break in loops like this:

```c3
while (try current = queue.pop_first())
{
    // do smth.
}
```
So, I just remove this lines and everything works fine.


lerno:
Quite right!

Issue #1579, opened 2024-10-27
Allow current dir as project


ygorpontelo:
This is a small annoyance, but everytime i want to create a project with c3c init i have to pass a name for another directory. Most of the time i already created the directory i want, so would be nice to have an option to just create in the current directory.

I took the liberty to implement this, since it was small enough. Hopefully it's not too bad.

lerno:
Do you feel this can be closed now?

ygorpontelo:
Tested now, seems to be working. It's also validating the directory name, which is nice.

PR #1580, opened 2024-10-27
allow current directory as project

ygorpontelo:
Adressing #1579


lerno:
I changed the code slightly, making the "." more explicit in the code and checking the resulting name.

Issue #1377, opened 2024-08-22
man page


lucas-mior:
A manual page for c3c would be very useful. My solution so far is a clunky shell script to generate a man page from the `--help` output.

lerno:
I guess you could make a pull request then link that pull request here?

PR #1581, opened 2024-10-28
wrote manpage in c3c.1 and added manpage install to CMakeLists.txt

goneal26:
Addressing #1377

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -525,6 +525,17 @@ endif()
 install(TARGETS c3c DESTINATION bin)
 install(DIRECTORY lib/ DESTINATION lib/c3)
 
+# Man page install (OSX/Linux only)
+set(MAN_PAGE_DIR "/usr/local/share/man/man1")
+message(STATUS "installing man page to ${MAN_PAGE_DIR}")
+
+if (NOT WIN32 AND EXISTS "${MAN_PAGE_DIR}")
+    install(FILES c3c.1 DESTINATION ${MAN_PAGE_DIR})
+elseif (NOT WIN32)
+    # won't create MAN_PAGE_DIR if it doesn't already exist
+    message(WARNING "Man page directory ${MAN_PAGE_DIR} does not exist, skipping man page installation")
+endif()
+
 if (C3_WITH_LLVM AND DEFINED sanitizer_runtime_libraries)
     add_custom_command(TARGET c3c POST_BUILD
         COMMAND "${CMAKE_COMMAND}" -E rm -rf -- $<TARGET_FILE_DIR:c3c>/c3c_rt
--- a/c3c.1
+++ b/c3c.1
@@ -1,0 +1,552 @@
+.TH "c3c" "1" "2024-10-27" "C3 Compiler" "User Commands"
+.SH NAME
+c3c \- Compiler for the C3 programming language
+
+.SH SYNOPSIS
+.B c3c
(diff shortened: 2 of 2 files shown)


bpo217:
I think this actually broke the MacOS brew install --head. Copying a man page fails with operation not permitted and homebrew doesn't support sudo installs. I switched to the using brew install --head instead of compiling myself because of lld being moved into a separate package on homebrew and I couldn't find any info about setting an LLD cmake flag akin to the LLVM_DIR for compling on at the latest and it seemed easier but now the whole thing fails because of the man page sudo priviledges.

bpo217:
Maybe I should create a new issue for this and reference?

lerno:
We could place the man page in different pages on MacOS perhaps?

goneal26:
> I think this actually broke the MacOS brew install --head. Copying a man page fails with operation not permitted and homebrew doesn't support sudo installs. I switched to the using brew install --head instead of compiling myself because of lld being moved into a separate package on homebrew and I couldn't find any info about setting an LLD cmake flag akin to the LLVM_DIR for compling on at the latest and it seemed easier but now the whole thing fails because of the man page sudo priviledges.

I had a friend point out a similar issue with MacOS to me on their M2- brew installed their man pages directory to a different location and I started working unofficially on a fix to account for this but I haven't been able to test it yet.

bpo217:
> We could place the man page in different pages on MacOS perhaps?

I agree with this as the first obvious solution and ran with it a bit. For completeness here is the failure I am referring to:
Screenshot 2024-11-03 at 08 50 00. 

Here's homebrew being protective:
Screenshot 2024-11-03 at 09 22 54

Further investigation revealed this file: `/private/etc/manpaths` and it's contents are:
```
/usr/share/man
/usr/local/share/man
```
Another file is `/private/etc/man.conf`. This is a portion of it:
```
...
# Every automatically generated MANPATH includes these fields
#
MANPATH	/usr/share/man
MANPATH	/usr/local/share/man
MANPATH	/usr/X11/man
MANPATH	/Library/Apple/usr/share/man
#
# Uncomment if you want to include one of these by default
#
# MANPATH	/opt/*/man
# MANPATH	/usr/lib/*/man
# MANPATH	/usr/share/*/man
# MANPATH	/usr/kerberos/man
...
```

I think where it's going is the right place. One just needs sudo priviledges... which fails with homebrew... Which is confusing because surely some other tool in homebrew installs a man page? Apparently not? I mean all these dirs require sudo priviledges. I switched back to compile from source and then it's just a matter of copying the man page file really. So this will need to be some kind of informative message after installing c3c for the man page. The brew script would need to install the man page along side c3c in it's own opt folder and then display a message that in order to use the man page you have to copy or symlink it or something. What do we think about this?

bpo217:
Actually that X11 man page is probably not standard maybe brew installed that. I'll check it out.

lerno:
Does it work now after the patch?

bpo217:
> Does it work now after the patch?

If you mean: does it work after this particular merge I am commenting on... The answer is I made the comment because the merge caused the break. The latest `brew install c3c --HEAD` fails. Unless you are referring to a newer one.

bpo217:
Looks like there was an patch which fixed it today! Coolio.

lerno:
Were you doing this from a script @goneal26 ? If so that would be useful to have as well.

lerno:
C3? 😄

PR #1590, opened 2024-11-01
Relative DESTINATION path when installing man page

rexim:
This makes the `install()` function install the file with respect to `CMAKE_INSTALL_PREFIX` which may not always be `/usr/local/` especially when some people install the compiler locally in their `$HOME` folder.

https://cmake.org/cmake/help/latest/command/install.html

I personally like to set my `CMAKE_INSTALL_PREFIX` to `$HOME/opt/c3`, so for me this results in this

2024-11-01-183528_1022x309_scrot

This PR makes it respect the installation prefix

2024-11-01-184954_820x259_scrot

cc: @goneal26


lerno:
Thanks!

PR #1594, opened 2024-11-01
Fix compile with gcc 14.2.1

Waqar144:
Fixes -Werror=maybe-uninitialized warnings


Waqar144:
There are also `-Werror=format-overflow` warnings in file_utils.c `file_append_path_temp()`:

```C
sprintf(path_buffer, "%s/%s", path, name);
...
sprintf(path_buffer, "%s%s", path, name);
```

I just locally added `-Wno-format-overflow` to avoid them for now.

lerno:
Could we instead set these to an invalid value? So rather than `null`, perhaps (void*)-1, and the len similarly. The best would be to add a `#define` like `#define NO_INIT_PTR (void*)0xAAAAAAAA` or something.

lerno:
Thank you!

PR #1596, opened 2024-11-02
encoding: implement RFC4648 base32 encoding

konimarti:
Implement base32 encoding and decoding according to RFC 4648 with the standard and extended hex alphabets. Add unit tests.

Link: https://www.rfc-editor.org/rfc/rfc4648

### DIFF
--- a/lib/std/encoding/base32.c3
+++ b/lib/std/encoding/base32.c3
@@ -52,10 +52,13 @@ fn void! Base32Encoder.init(&self, Alphabet encoder = STD_ALPHABET, int padding 
 fn usz Base32Encoder.encode_len(&self, usz n)
 {
 	// A character is encoded into 8 x 5-bit blocks.
-	if (self.padding >= 0) {
+	if (self.padding >= 0)
+	{
 		// with padding
 		return (n + 4) / 5 * 8;
-	} else {
+	}
+	else
+	{
 		// no padding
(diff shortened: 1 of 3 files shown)


lerno:
Looks good, can you also put a not in the release notes for it, plus change to Allman brace style, which is the style used in the stdlib elsewhere, so rather than this:
```c
 	if (self.padding >= 0) {
 		// with padding
 		return (n / 8) * 5;
 	} else {
 		// no padding
 		usz trailing = n % 8;
 		return n / 8 * 5 + (trailing * 5 ) / 8;
 	}
```
Use:
```c
 	if (self.padding >= 0) 
 	{
 		// with padding
 		return (n / 8) * 5;
 	} 
 	else 
 	{
 		// no padding
 		usz trailing = n % 8;
 		return n / 8 * 5 + (trailing * 5 ) / 8;
 	}
```

konimarti:
Thanks @lerno! I've addressed your comments above with two fixup commits for easier review. Let me know if you want me to squash them.

lerno:
I can squash it, so that's fine. Thank you for the contribution!

lerno:
One thing that would be nice would be to have Base64 and Base32 streams. It would be a much nicer interface to output things.

lerno:
Yeah, it's not super hard to, and you get some nice features to copy one stream to another.

Issue #1582, opened 2024-10-30
`@tag` attribute is allowed on macros but value cannot be retrieved with tagof


Book-reader:
trying to retrieve a tag from a macro results in the error `There is no member or method 'tagof' on 'void'`.

As far as I am aware, macros have no properties like nameof and tagof, so maybe macros should be disallowed from having the `@tag` attribute.

lerno:
This should be fixed now, please try it

Book-reader:
works correctly now, thanks!

PR #1597, opened 2024-11-03
Make tag not be allowed on macros

StandUp2001:
I have made it so when you try to add a `@tag` by a macro it will give an error.
This was because of this issue: #1582


lerno:
Actually, it *should* be allowed, the fix was slightly complicated but I've fixed it now. But thank you for submitting a patch.

PR #1600, opened 2024-11-04
Fix: Unify termios types for actions & flags

tomaskallup:
The type for `Tcflags` was used instead of the CInt (now `Tcactions`) and vice versa.

This required some explicit casts when working with termios:
```
termios.c_lflag &= (Tcflags)~(termios::ICANON | termios::ECHO);
termios.setAttr(self.fd, (int)termios::TCSANOW);
```

With this PR, it can just be:
```
termios.c_lflag &= ~(termios::ICANON | termios::ECHO);
termios.setAttr(self.fd, termios::TCSANOW);
```

It should also reflect what the C implementation has.


lerno:
Looks correct. Can you add a test too?

lerno:
Nevermind, it would only be run on some platforms.

lerno:
Thank you!

PR #1602, opened 2024-11-06
Fix Formatter.floatformat and Object.to_format

konimarti:
Fix two minor issues in the stdlib.


lerno:
Good catches, thank you!

PR #1614, opened 2024-11-11
Add support for nix flakes

vssukharev:
Added support for nix flakes. 

Flake provides the following outputs:
- `packages.${system}.default` - refers to `c3c`
- `packages.${system}.c3c` - c3c derivation
- `packages.${system}.c3c-debug` - debug version (also includes compile_commands.json in derivation path)
- `packages.${system}.c3c-nochecks` - same as `c3c-debug` but with omitted checks
- `devShells.${system}.default` - dev shell for lsp

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,3 +75,6 @@ TAGS
 /.cache/
 /compile_commands.json
 
+# 'nix build' resulting symlink
+result
+
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,61 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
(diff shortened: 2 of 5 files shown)


tomaskallup:
I can compile & use c3c via `nix build`. The devShell is... sketchy, due to the LLVM & LLD being separeted, cmake won't work as it doesn't find `LLD`. Sadly that's not really something we can handle, unless we do a shellHook which does the replacing, but then it polutes users git state with modified `CMakeLists.txt`

vssukharev:
I'made dev shell only for LSP working correctly. So that you need to use `nix build .#c3c-nochecks` to make project built. Unfortunately, as you said, there are no appropriate workarounds how to do it in dev shell directly.

lerno:
Thank you!

Issue #1520, opened 2024-10-04
c3c init example does not add test/** to "sources" in project.json


joshring:
```bash
c3c init example
```
Does not add `test/**` to "sources" in project.json

 By default, running `c3c test` does not work for tests unless the test is inside of `src/**`, let's add this change to fix it to project.json

```json
"sources": [ "src/**", "test/**" ],
```

lerno:
There should be an additional `"test-sources": [ "test/**"]` which should be added when invoking "test" to avoid compiling them otherwise. This is not implemented.

joshring:
Could test directory and project.json entry added, on `c3c init` and only compiled on `c3c test` ?

cardisk:
@lerno is this feature still incomplete in some way?
By simply adding `"test-sources": [ "test/**"]` to `project.json` the compiler works as I expect, except for a warning about an unknown entry within the json file.
I fixed the warning and added the entry to the `init` command. 
Is there anything else that can be done?

lerno:
On top of this, the these test sources should be added to what is compiled *only* when the the test flag is added. @cardisk

cardisk:
I tested a bit, it seems to be so

lerno:
Great! Then we can close this one.

PR #1615, opened 2024-11-12
Init command will now add `test-sources` to `project.json` #1520

cardisk:
I recreated the pull request because I broke the previous one accidentally. Now it has the fix under 0.6.5


cardisk:
No problem! Thanks for understanding

PR #1616, opened 2024-11-12
Fix WriteBuffer.write_byte

hwchen:
Fixes implementation of `WriteBuffer.write_byte`.

### DIFF
--- a/lib/std/io/stream/buffer.c3
+++ b/lib/std/io/stream/buffer.c3
@@ -120,10 +120,13 @@ fn usz! WriteBuffer.write(&self, char[] bytes) @dynamic
 
 fn void! WriteBuffer.write_byte(&self, char c) @dynamic
 {
-	usz n = self.bytes.len - self.index;
-	if (n == 0) self.write_pending()!;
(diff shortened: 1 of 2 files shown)


hwchen:
Sorry, I've only verified this in my own program. I can write a unit test later, but am a bit busy right now.

lib/std/io/stream/buffer.c3
```diff
@@ -120,14 +120,17 @@ fn usz! WriteBuffer.write(&self, char[] bytes) @dynamic
...
 fn void! WriteBuffer.write_byte(&self, char c) @dynamic
 {
-	usz n = self.bytes.len - self.index;
-	if (n == 0) self.write_pending()!;
-	self.bytes[0] = c;
-	self.index = 1;
+	usz n = self.bytes.len - self.index - 1;
+	if (n == 0) {
+		self.write_pending()!;
+		self.index = 0;
```
lerno, line 126: This row is redundant I think?

lerno:
Thank you!

PR #1625, opened 2024-11-14
Fix WriteBuffer.write_bytes off-by-one

hwchen:
Was randomly thinking about my previous PR https://github.com/c3lang/c3c/pull/1616 which fixed WriteBuffer.write_bytes and realized I was off-by-one because I didn't think about the post-increment. Basically, the buffer was flushed one step too early.

With the bug, no data was lost, it's just that the WriteBuffer buffer would effectively be shortened by one. Now, the entire length of the buffer is used.


hwchen:
I also added a test, which is a bit low-level. Can easily be reverted, as it's in a separate commit.

lerno:
It's all good. Thank you!

PR #1626, opened 2024-11-15
Add io stream primitives

konimarti:
Add io stream primitives for the InStream and OutStream interfaces:
- MultiReader
- MultiWriter
- TeeReader

Inspired by Golang's io package.


konimarti:
Ugh, sorry. All tests failed. After a cursory look it's not obvious to me why it failed. I'll have to look into this.

lib/std/io/io.c3
```diff
@@ -42,6 +42,7 @@ fault IoError
 	UNKNOWN_ERROR,
 	UNSUPPORTED_OPERATION,
 	WOULD_BLOCK,
+	SHORT_WRITE,
```
lerno, line 45: What is the difference between this one an "INCOMPLETE_WRITE"?

lib/std/io/stream/multireader.c3
```diff
@@ -0,0 +1,71 @@
...
+<*
+ @param [&inout] self
+ @param [&inout] allocator
+ @require self.readers.len == 0 "Init may not run on already initialized data"
+ @ensure self.index == 0
+*>
+fn MultiReader* MultiReader.new_init(&self, InStream... readers, Allocator allocator = allocator::heap())
+{
+	usz n = readers.len;
+	InStream *ptr =  allocator::new_array(allocator, InStream, n);
```
lerno, line 24: `allocator::new_array` already returns a slice, so you can do:
```c
InStream[] copy = allocator::new_array(allocator, InStream, readers.len);
copy[..] = readers[..];
*self = { .readers = copy, .allocator = allocator };
return self;
```

lib/std/io/stream/multireader.c3
```diff
@@ -0,0 +1,71 @@
...
+*>
+fn MultiReader* MultiReader.temp_init(&self, InStream... readers)
+{
+	return self.new_init(...readers, allocator: allocator::temp());
+}
+
+fn void MultiReader.free(&self)
+{
+	if (!self.allocator) return;
+	if (void* ptr = self.readers.ptr) allocator::free(self.allocator, ptr);
```
lerno, line 43: This:
```c
if (void* ptr = self.readers.ptr) allocator::free(self.allocator, ptr);
```
Can be replaced by:
```c
allocator::free(self.allocator, self.readers);
```

lib/std/io/stream/multiwriter.c3
```diff
@@ -0,0 +1,60 @@
...
+<*
+ @param [&inout] self
+ @param [&inout] allocator
+ @require writers.len > 0
+ @require self.writers.len == 0 "Init may not run on already initialized data"
+*>
+fn MultiWriter* MultiWriter.new_init(&self, OutStream... writers, Allocator allocator = allocator::heap())
+{
+	usz n = writers.len;
+	OutStream *ptr =  allocator::new_array(allocator, OutStream, n);
```
lerno, line 21: Same note as for the Reader

lib/std/io/stream/multiwriter.c3
```diff
@@ -0,0 +1,60 @@
...
+	*self = {};
+}
+
+fn usz! MultiWriter.write(&self, char[] bytes) @dynamic
+{
+	usz n;
+	foreach (w : self.writers) 
+	{
+		n = w.write(bytes)!;
+		if (n != bytes.len) return IoError.SHORT_WRITE?;
```
lerno, line 50: INCOMPLETE_WRITE maybe?

lib/std/io/stream/teereader.c3
```diff
@@ -0,0 +1,42 @@
...
+	OutStream w;
+}
+
+<* Returns a reader that implements InStream and that will write any data read
+ from the wrapped reader r to the writer w. There is no internal buffering.
+ 
+ @param [&inout] r "Stream r to read from."
+ @param [&inout] w "Stream w to write to what it reads from r."
+*>
+fn TeeReader tee_reader(InStream r, OutStream w) => { r, w };
```
lerno, line 15: This could possibly be a macro.

lerno:
Probably the added IO fault added the line number by one in a bunch of places and the test scripts are sensitive to that. I can fix it if you want. Did you need the "SHORT_WRITE" when there was INCOMPLETE_WRITE? If you remove SHORT_WRITE then it will probably just work for now.

lib/std/io/io.c3
konimarti: I must have overlooked INCOMPLETE_WRITE. That expresses the indent of short write error. I'll adjust this and hope that the build error goes away.

lerno:
Thank you!

PR #1629, opened 2024-11-18
Fix error when HashMap.remove on uninitialized HashMap

hwchen:
Bug: when `HashMap` is uninitialized, calling `remove` will result in an error like

```
ERROR: 'Array index out of bounds (array had size 0, index was 3456971336)'
```

This is because a non-zero hash for the array idx is calculated even when the array is len 0.

Other methods handle this by short-circuiting before attempting array access when map is uninitialized. This PR applies the same fix.


hwchen:
I did a quick browse of the hashmap implementation, didn't see other methods that missed checking for uninitialized.

PR #1634, opened 2024-11-20
Add new AMX and other feature flags

sandervdbosch:
I added AMX and other feature flags that the compiler did not handle on my machine.
I did this following the pattern from #1482.
Since I am not sure if I added the flags correctly to the [x86_features_add_feature](https://github.com/sandervdbosch/x86-AMX-features/blob/7a5039f4480f7911b32eb053b99cb9b44521cb4d/src/compiler/target.c#L528) function in target.c so I would recommend someone to look into that. Also I was unsure if I needed to add the flag to [x86_initialize_cpu_features](https://github.com/sandervdbosch/x86-AMX-features/blob/7a5039f4480f7911b32eb053b99cb9b44521cb4d/lib/std/core/private/cpu_detect.c3#L149C9-L149C36), so input on that part is also very welcome.

PS. Sorry for some seemingly unnecessary indentation changes, I was not aware my editor was messing with that.


lerno:
Eventually we want to make sure that all of these new flags are handled completely. Esp if we add new CPU baseline types, but for now this is fine. Thank you!

PR #1640, opened 2024-11-23
add check for to_string in is_struct_with_default_print

ellipse12:
The function wasn't checking if the struct had the method `to_string` which caused a segfault when trying to override the default to_string on a struct


lerno:
Thank you!

PR #1641, opened 2024-11-23
fix: base64 decoding

konimarti:
Fix the base64 decoding. If there's an 'A' character in the encoded text, the base64 decode function returns an INVALID_PADDING error. The reason lies in the way Base64Decoder.init tries to find a suitable invalid character. Fix this by defining the invalid character as 0xff (which is already the case for a decoding without padding).

This error has not been caught by the test harness, because no test contains an 'A' character in the the encoded text yet. Add a new test.


lerno:
Thank you!

PR #1643, opened 2024-11-24
encoding: add hex encoding (base16)

konimarti:
Implement hex encoding and decoding (base16) according to RFC 4648. Add unit tests.

Link: https://www.rfc-editor.org/rfc/rfc4648


lerno:
Thank you, you can probably have a look at the stdlib and see if there are any hex conversions there that can be replaced.

PR #1647, opened 2024-11-28
Add char_at method in DString

neokeld:
It is not intuitive to do `&payload_string.str_view()[index]` to get a char from the DString (data method is private and so I assume this is the good thing to do).
I propose here a char_at method to do &payload_string.char_at(index)
Let me know if this stupid, I'm new to C3 😃

### DIFF
--- a/lib/std/core/dstring.c3
+++ b/lib/std/core/dstring.c3
@@ -132,7 +132,7 @@ fn usz DString.capacity(self)
 	return self.data().capacity;
 }
 
-fn usz DString.len(&self) @dynamic
+fn usz DString.len(&self) @dynamic @operator(len)
 {
 	if (!*self) return 0;
 	return self.data().len;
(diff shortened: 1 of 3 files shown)


lib/std/core/dstring.c3
```diff
@@ -154,6 +154,14 @@ fn String DString.str_view(self)
 	return (String)data.chars[:data.len];
 }
 
+<*
+ @require index < self.len()
+ @require self.data() "Empty string"
+*>
+fn char DString.char_at(self, usz index) {
```
lerno, line 161: Can you fix the formatting? All the other stdlib code uses Allman brace, so this should be brought in style with everything else. The other thing you might want to consider is to add `@operator([])` to enable `dstring[2]`. In this case the len() function should also get `@operator(len)` to allow foreach and reverse indexing. Similarly you might also consider adding support for []= and &[] as well.

lerno:
On top of the changes I requested, can you also add a note in the releasenotes?

lib/std/core/dstring.c3
neokeld: Of course, thank you! I'm also adding a few tests. I haven't added &[] yet. I'll pick this up again on the occasion.

lerno:
Thank you!

PR #1652, opened 2024-12-01
fix: prevent infinite read-loop by updating `left_to_read` after write

NexushasTaken:
Fixed a bug where `left_to_read` was not updated after writing, causing the loop to continue indefinitely, resulting in over-read and write random data to the file.

### DIFF
--- a/src/build/libraries.c
+++ b/src/build/libraries.c
@@ -224,7 +224,7 @@ static inline JSONObject *resolve_zip_library(BuildTarget *build_target, const c
 		zip_check_err(lib, zip_dir_iterator_next(&iterator, &file));
 		if (file.uncompressed_size == 0 || file.name[0] == '.') continue;
 		// Copy file.
-		zip_file_write(f, &file, dir, false);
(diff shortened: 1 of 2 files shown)


src/build/libraries.c
```diff
@@ -224,7 +224,7 @@ static inline JSONObject *resolve_zip_library(BuildTarget *build_target, const c
 		zip_check_err(lib, zip_dir_iterator_next(&iterator, &file));
 		if (file.uncompressed_size == 0 || file.name[0] == '.') continue;
 		// Copy file.
-		zip_file_write(f, &file, dir, false);
+		ASSERT0(zip_file_write(f, &file, dir, false) == NULL);
```
lerno, line 227: This is not right, since ASSERT0 is removed in release builds. Better test and do an error_exit if it fails.

NexushasTaken: ohh okay, i initially thought it meant to be use to notice the user to file a bug report; my bad.

ill just remove it.

lerno:
Thank you.

PR #1660, opened 2024-12-05
sort: add is_sorted

konimarti:
Add `is_sorted` function to check whether a list is sorted or not. Sort order (ascending or descending) will be detected by looking at the data. Add tests.

### DIFF
--- a/lib/std/sort/sorted.c3
+++ b/lib/std/sort/sorted.c3
@@ -1,0 +1,69 @@
+module std::sort;
+import std::sort::sorted;
+
+<*
+Returns true if list is sorted in either ascending or descending order.
+ @require @is_sortable(list) "The list must be indexable and support .len or .len()"
+ @require @is_valid_cmp_fn(cmp, list, context) "Expected a comparison function which compares values"
+ @require @is_valid_context(cmp, context) "Expected a valid context"
+*>
+macro bool is_sorted(list, cmp = EMPTY_MACRO_SLOT, context = EMPTY_MACRO_SLOT) @builtin
+{
+	usz len = sort::@len_from_list(list);
+	if (len <= 1) return true;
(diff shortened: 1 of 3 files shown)


lerno:
I can accept this as-is, but note that if you want a generic function you can also implement it as a lambda inside of a macro.

lerno:
If you have this:
```c
macro bool is_sorted(list, cmp = EMPTY_MACRO_SLOT, context = EMPTY_MACRO_SLOT) @builtin
 {
  usz len = sort::@len_from_list(list);
  if (len <= 1) return true;
   return sorted::is_sorted(<$typeof(list), $typeof(cmp), $typeof(context)>)(list, 0, (isz)len - 1, cmp, context);
 }
 ```
 You can use a lambda:
 ```c
 macro bool is_sorted(list, cmp = EMPTY_MACRO_SLOT, context = EMPTY_MACRO_SLOT) @builtin
 {
  var $Type = $typeof(list);
  usz len = sort::@len_from_list(list);
  if (len <= 1) return true;
  var sort = fn bool($Type list, usz start, usz end, $typeof(cmp) cmp, $typeof(context) context) 
  {
 	   ... inline the generic function here ...
  };
  return sort(list, 0, len - 1, cmp, context);
 }
 ```
 Or do you envision people using the generic is_sorted outside of the macro?

konimarti:
Ah, I see. That makes sense, too.

I copied the structure over from the quicksort implementation and figured that the macro is a convenient function for most users but that they can have direct access to the generic is_sorted if need (e.g. when they want to check if only a subsection of the list is sorted, but not sure how relevant that is).

Also, I forgot to add a release note..

lerno:
You could also pull it into a second macro that holds the lambda alone.

lerno:
Thank you!

PR #1667, opened 2024-12-07
Fix fnv a hashes

konimarti:
Order of the multiplication and XOR operations should be reversed for FNV-1a hashes.

Link: https://datatracker.ietf.org/doc/draft-eastlake-fnv/
Link: https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function
Link: https://md5calc.com/hash/fnv132/Hello+world

### DIFF
--- a/lib/std/hash/fnv64a.c3
+++ b/lib/std/hash/fnv64a.c3
@@ -8,7 +8,7 @@ distinct Fnv64a = ulong;
 const FNV64A_START @private = 0xcbf29ce484222325;
 const FNV64A_MUL @private = 0x00000100000001b3;
 
-macro void @update(ulong* &h, char x) @private => *h = (*h * FNV64A_MUL) ^ x;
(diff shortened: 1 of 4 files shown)


lib/std/hash/fnv64a.c3
```diff
@@ -27,7 +27,9 @@ fn void Fnv64a.update(&self, char[] data)
 
 macro void Fnv64a.update_char(&self, char c)
 {
-	@update(*self, x);
+	ulong h = (ulong)*self;
```
lerno, line 30: Why did you make this change?

konimarti: There's no `x`only a char `c` in the macro. The rest is to make the call to @update work (maybe there's a more compact way).

lerno:
I simplified it a little. Thanks for finding this.

PR #1668, opened 2024-12-08
math: update complex numbers

konimarti:
Add `inverse`, `conjugate`, and `equals` functions to the Complex numbers. Add an `IMAGINARY` constant to represent the imaginary unit. Add unit tests and test for different types.


lerno:
Thank you!

PR #1671, opened 2024-12-10
math: add gcd and lcm

konimarti:
Add gcd and lcm functions to calculate the greatest common divisor (gcd) and the least common multiple (lcm) to the math module. This will also work for BigInts that implements its own gcd/lcm.


lerno:
Thank you!

PR #1675, opened 2024-12-12
Fix weekday conversion from `tm.tm_wday`

radekm:
Currently

```c3
import std;

fn void main()
{
    DateTime dt = datetime::now();
    io::printfn("%s %s %s %s %s %s %s", dt.year, dt.month, dt.day, dt.hour, dt.min, dt.sec, dt.weekday);
}
```

prints `2024 DECEMBER 12 15 33 30 SATURDAY` but it should print `2024 DECEMBER 12 15 33 30 THURSDAY`.


lerno:
Thank you!

PR #1676, opened 2024-12-13
math: fix adjoint of Matrix2

konimarti:
Fix the adjoint of the Matrix2x2 implementation in the math module. This also fixes the calculation of the inverse which depends on the adjoint.


lerno:
Can you add it to the release notes too?

lerno:
Thank you!

PR #1682, opened 2024-12-15
Support detection of shared MinGW-w64 LLVM libraries

brechtsanders:
Support detection of shared MinGW-w64 LLVM libraries by also looking for liblld*.dll.a files.


lerno:
Thank you!

PR #1691, opened 2024-12-17
in new_struct_to_str, fix uaf

hwchen:
In `new_struct_to_str`, a temporary DString (initialized within the scope of `@stack_mem`) is used for appending writes, before copying to the final result using the allocator passed in to `new_struct_to_str`. However, the final copy happens after the temp string is outside the scope of `@stack_mem`, resulting in use-after-free.


lerno:
Thank you!

Issue #1695, opened 2024-12-17
HashMap `temp_init_from_map` skips over some elements


tomaskallup:
Example code:
```
import std::io;
import std::collections::map;

def Vector2 = int[<2>];

fn uint Vector2.hash(self) => self.x * 31 + self.y;

HashMap(<Vector2, Vector2>) visited_positions;

fn void main() {
  visited_positions.set({1,10}, {0,-1});
  visited_positions.set({1,5}, {0,-1});
  visited_positions.set({1,6}, {0,-1});
  visited_positions.set({1,7}, {0,-1});
  visited_positions.set({3,5}, {0,1});

  visited_positions.@each(;Vector2 pos, Vector2 dir) {
    io::printfn("%s: %s", pos, dir);
  };

  io::printn("Copy:");

  HashMap(<Vector2, Vector2>) visited_positions_copy;
  visited_positions_copy.temp_init_from_map(&visited_positions);
  visited_positions_copy.@each(;Vector2 pos, Vector2 dir) {
    io::printfn("%s: %s", pos, dir);
  };
}
```

Running this with `c3c compile-run test.c3` will result in following output:
```
Program linked to executable 'test'.
Launching ./test
[<3, 5>]: [<0, 1>]
[<1, 7>]: [<0, -1>]
[<1, 5>]: [<0, -1>]
[<1, 6>]: [<0, -1>]
[<1, 10>]: [<0, -1>]
Copy:
[<3, 5>]: [<0, 1>]
[<1, 5>]: [<0, -1>]
[<1, 6>]: [<0, -1>]
[<1, 10>]: [<0, -1>]
```
As you can see, `[<1, 7>]: [<0, -1>]` is missing in the copied HashMap.

tomaskallup:
Managed to get even simpler repro:
```
import std::io;
import std::collections::map;

def Map = HashMap(<String, usz>);

fn void main() {
  Map map;
  map.temp_init();

  map.set("aa", 1);
  map.set("b", 2);
  map.set("bb", 1);

  Map map_copy;
  map_copy.temp_init_from_map(&map);
  map_copy.@each(;String key, usz val) {
    io::printfn("%s: %s", key, val);
  };
}
```

PR #1696, opened 2024-12-17
Fix hashmap put all for create + test case

tomaskallup:
I based these changes on the implementation of `@each`, which does an inner `while` loop, checking for `entry.next`.

Closes: #1695

### DIFF
--- a/lib/std/collections/hashmap.c3
+++ b/lib/std/collections/hashmap.c3
@@ -455,8 +455,11 @@ fn void HashMap.put_all_for_create(&map, HashMap* other_map) @private
 	if (!other_map.count) return;
 	foreach (Entry *e : other_map.table)
 	{
-		if (!e) continue;
-		map.put_for_create(e.key, e.value);
+		while (e)
+		{
+			map.put_for_create(e.key, e.value);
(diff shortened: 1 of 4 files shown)


lib/std/collections/hashmap.c3
```diff
@@ -455,8 +455,10 @@ fn void HashMap.put_all_for_create(&map, HashMap* other_map) @private
 	if (!other_map.count) return;
 	foreach (Entry *e : other_map.table)
 	{
-		if (!e) continue;
-		map.put_for_create(e.key, e.value);
+		while (e) {
```
lerno, line 458: Can you fix it to match the other coding style, so Allman braces?

lerno:
Can you also add mention in the release notes?

lerno:
And the same bug is presumably in map.c3 as well, can you fix it?

tomaskallup:
Fixed it up, I'll try to also add unit tests for the regular `Map`.

lerno:
Yes, they are different. There was a question on what to go for.

lerno:
Thank you for the fix!

PR #1699, opened 2024-12-20
Adding trigonometric and hyperbolic trigonometric tests 

MendocinoWhiteDeer:
Changes were primarily made to the math_tests module. The main goal is to add tests to test/unit/stdlib/math/math.c3.

Rewrote the test_atan() function to be a bit more compact and organized while adding to the net number of assertions overall. 

Next, I added a number of trigonometric and hyperbolic trigonometric tests. I didn't write tests for sinh, cosh, and tanh because the macros use the exponential macro and not LLVM intrinsics (for now). However, there are inverse hyperbolic trig tests.

When setting up the cosine test I found and fixed a bug where the cosine macro wouldn't accept an integer input.

### DIFF
--- a/lib/std/math/math.c3
+++ b/lib/std/math/math.c3
@@ -290,7 +290,7 @@ macro copysign(mag, sgn) => $$copysign(values::promote_int_same(mag, sgn), ($typ
 <*
  @require values::@is_promotable_to_floatlike(x) `The input must be a number value or float vector`
 *>
-macro cos(x) => $$cos(x);
+macro cos(x) => $$cos(values::promote_int(x));
 
 <*
  @require values::@is_promotable_to_floatlike(x) `The input must be a number value or float vector`
--- a/lib/std/math/math.c3
+++ b/lib/std/math/math.c3
@@ -132,6 +132,28 @@ macro deg_to_rad(x) {
 macro abs(x) => $$abs(x);
 
 <*
+ @require values::@is_int(x) || values::@is_float(x) "Expected an integer or floating point value"
+ @require values::@is_int(y) || values::@is_float(y) "Expected an integer or floating point value"
+*>
+macro is_approx(x, y, eps)
+{
+	if (x == y) return true;
+	if (is_nan(x) || is_nan(y)) return false;
+	return abs(x-y) <= eps; 
+}
+
+<*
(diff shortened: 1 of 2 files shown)


MendocinoWhiteDeer:
I'm sorry for all the force pushes. I rewrote all the commits after accidentally adding a binary file and then had to rebase to fix commit typos. As of now, I won't make any further edits on my own.

lerno:
The tests seem to fail on some platforms, I am guessing they're not identical but just less than an epsilon apart?

MendocinoWhiteDeer:
Yeah that seems to be the problem. Direct comparison worked fine on my Linux distribution but it was naive of me to assume checking floats exactly would work cross-platform for so many tests. I could add two macros for floating point comparison into the math module and check for an acceptable tolerance in the test file. I was looking at some of my older C code and I think I could translate it into something like this

```
<*
 @require values::@is_int(x) || values::@is_float(x) "Expected an integer or floating point value"
 @require values::@is_int(y) || values::@is_float(y) "Expected an integer or floating point value"
*>
macro is_approx(x, y, eps)
{
	$if x == y:
	    return true;
	$endif
	$if is_nan(x) || is_nan(y):
	    return false;
	$endif
	return abs(x-y) <= eps; 
}

<*
 @require values::@is_int(x) || values::@is_float(x) "Expected an integer or floating point value"
 @require values::@is_int(y) || values::@is_float(y) "Expected an integer or floating point value"
*>
macro is_approx_rel(x, y, eps)
{
	$if x == y:
	    return true;
	$endif
	$if is_nan(x) || is_nan(y):
	    return false;
	$endif
	return abs(x-y) <= eps * max(abs(x), abs(y)); 
}
```

The first one is for comparing small values near zero and should take tolerances roughly on the order of the floating point precision and the second should take tolerances roughly on the order of the square root of the floating point precision. I'll be conservative with the tests and mostly use the latter macro and forgiving tolerances.

MendocinoWhiteDeer:
This time I think the tests should work cross-platform. A relative error of 1e-6 for floats and 1e-12 for doubles should be lenient enough to account for the error. Like before, all the tests were passed on my system. The vast majority of the checks are for relative error, not absolute error.

lerno:
Yes, all good now!

PR #1706, opened 2024-12-21
Prevent trailing _ for all numbers

lith-x:
Just doing some basic QA as I was picking up the language, noticed the compiler gave an error for `int x = 0x01_;` but binary/octal numerals like `int x = 0b01_;` compiled without fail, just moved the underscore check in `lexer.c` from `scan_hex` and `scan_dec` to a single check in `scan_number_suffix`, which is called from: `scan_hex`, `scan_oct`, `scan_binary`, and `scan_dec`.

Thought about copy-pasting the check throughout the various `scan_[numtype]` funcs, but it seems more effective and maintainable here, as it's the only spot a trailing underscore would occur.


lerno:
Can you add the update to the release notes too please?

lith-x:
Noted "#" was being let through on `--list-operators`, after going through the grammar it's being used only in ways that `$` and `@` are used, which are culled, so I added `#` to the check as well. All changes reflected in `releasenotes.md` :)

lerno:
Thank you!

PR #1709, opened 2024-12-22
C3_MATH feature

MendocinoWhiteDeer:
This feature allows the usage of nolibc math files even when libc is in use. The motivation is so that `compile-tests` may be used to test these functions without any hassle. If a nolibc definition exists, it will be used in place of libc, otherwise, it will default to libc.

To run math tests, for example,

```
c3c compile-test -D C3_MATH test/unit/stdlib/math/math.c3
```

The change involves swapping `module std::math::nolibc @if(env::NO_LIBC);` at the start of each nolibc math file with `module std::math::nolibc @if(env::NO_LIBC || $feature(C3_MATH));`.

### DIFF
--- a/lib/std/math/math_nolibc/atan.c3
+++ b/lib/std/math/math_nolibc/atan.c3
@@ -1,5 +1,17 @@
 module std::math::nolibc @if(env::NO_LIBC || $feature(C3_MATH));
 
+/* origin: FreeBSD /usr/src/lib/msun/src/s_atan.c */
+/*
+ * ====================================================
+ * Copyright (C) 1993 by Sun Microsystems, Inc. All rights reserved.
+ *
+ * Developed at SunPro, a Sun Microsystems, Inc. business.
+ * Permission to use, copy, modify, and distribute this
+ * software is freely granted, provided that this notice
+ * is preserved.
+ * ====================================================
+ */
+
 const double[*] ATANHI @private = {
   4.63647609000806093515e-01, /* atan(0.5)hi 0x3FDDAC67, 0x0561BB4F */
   7.85398163397448278999e-01, /* atan(1.0)hi 0x3FE921FB, 0x54442D18 */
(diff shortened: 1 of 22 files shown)


MendocinoWhiteDeer:
I entered an extra commit. Some of the functions in the same file come from different files in musl (which in turn were lifted from FreeBSD). So an appropriate filepath to the original source code was provided for each function.

lib/std/math/math_nolibc/atan.c3
```diff
@@ -175,6 +202,19 @@ fn float _atanf(float x) @weak @extern("atanf") @nostrip
...
+/*
+ * ====================================================
+ * Copyright (C) 1993 by Sun Microsystems, Inc. All rights reserved.
+ *
+ * Developed at SunSoft, a Sun Microsystems, Inc. business.
+ * Permission to use, copy, modify, and distribute this
+ * software is freely granted, provided that this notice
+ * is preserved.
+ * ====================================================
+ *
```
lerno, line 215: This looks like just the comment was added, and no code? So either both or neither? ;)

MendocinoWhiteDeer: I removed it so the code can merge but I'd appreciate it if you read what I wrote below. 
 
In atan.c3, I assumed multiple functions were taken from musl (which in turn were taken from FreeBSD). In the other nolibc files you had a different MIT license notice for each function (as they each came from different files in the original source code).  

Basically I added the comments because I wanted it to match the other files, and I assumed that having those comments there was the original intent (but I didn't write the code so I realize I shouldn't assume that without asking haha).

I found the source code links in those notices helpful as well (the original source provides a bit more explanation for the algorithm which is omitted from the slightly more terse c3 implementation).

lerno: Yeah, I missed that part of your explanation when I looked at the code. To be honest, they are from various sources, some come credited from MUSL in which case I kept the comments. When there are no comments it's because there were no comments in that source. It's a bit of a mixed bag 😥

Accidental wrong attribution seems like a very bad thing in itself 😰

From the code can you be sure it is from that source then we should definitely add it. If all code matches up (I tried to keep names etc) then we can add them.

Also, I have encountered the same code with different attributions, in which case I've tried to pick the one that seemed more likely to be the correct one.

MendocinoWhiteDeer: Yes I'm very sure all the atan functions can be traced to musl which attribute FreeBSD. The comments I copied over were just 4 MIT License notices that each contained a link to the original FreeBSD file.

lerno:
Thank you!

PR #1712, opened 2024-12-24
math::nolibc: acos and asin

MendocinoWhiteDeer:
There are now nolibc definitions for the inverse cosine and inverse sine. These definitions, like many others in nolibc, originate from FreeBSD (see [here](https://cgit.freebsd.org/src/tree/lib/msun/src )) or you can see them in use in musl (see [here](https://cgit.freebsd.org/src/tree/lib/msun/src)). 

If anybody would like to verify that these nolibc functions are correct, you can do so in a libc environment with the feature `C3_MATH`. This replaces libc definitions with nolibc definitions when they exist and defaults to libc when they don't. If anybody wants to test the authenticity of these functions run

`c3c compile-test -D C3_MATH test/unit/stdlib/math/math.c3`

One can even ruin the test results in different branches of the functions and cause the tests to fail for particular points. 

More test points were added for acos, asin, and atan in the math_tests module. This was done because the nolibc inverse trigonometric functions have various branching conditions depending on the provided input value. Tests failed to send inputs to certain branches in these functions previously.

### DIFF
--- a/lib/std/math/math_nolibc/acos.c3
+++ b/lib/std/math/math_nolibc/acos.c3
@@ -1,0 +1,146 @@
+module std::math::nolibc @if(env::NO_LIBC || $feature(C3_MATH));
+
+/* origin: FreeBSD /usr/src/lib/msun/src/e_acos.c */
+/*
+ * ====================================================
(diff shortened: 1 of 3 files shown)


lerno:
Thank you!

PR #1714, opened 2024-12-25
Fixed -P JSON syntax, string conversion.

lith-x:
- Changed incorrect `[` `]` to `{` `}`
- Implemented string conversions for `CONST_BYTES` that will now translate to `x\"01 23 4F F0\"`
- `CONST_STRING` is now "JSON legal", i.e. all escapes that are necessary (per [JSON RFC](https://datatracker.ietf.org/doc/html/rfc8259#section-7), double quotes, backslashes, ASCII control characters) will happen, such that it won't break formatting.

### DIFF
--- a/src/compiler/json_output.c
+++ b/src/compiler/json_output.c
@@ -408,7 +408,7 @@ void print_var_expr(FILE *file, Expr *expr)
                             fputs("\\\"", file);
                             break;
                         }
-                        const unsigned char *ptr = expr->const_expr.bytes.ptr;
+                        const unsigned char *ptr = (unsigned char *)expr->const_expr.bytes.ptr;
                         char *res = malloc_string((size_t)(len * 3 + 1));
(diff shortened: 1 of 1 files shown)


lerno:
It seems like there is some tiny issue with the code when compiling with the default warnings-as-errors:
```
/home/runner/work/c3c/c3c/src/compiler/json_output.c:411:46: error: initializing 'const unsigned char *' with an expression of type 'const char *' converts between pointers to integer types where one is of the unique plain 'char' type and the other is not [-Werror,-Wpointer-sign]
  411 |                         const unsigned char *ptr = expr->const_expr.bytes.ptr;
      |                                              ^     ~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```
Could you please fix that? I think it could possibly just be an explicit cast to fix.

lerno:
Thank you!

PR #1723, opened 2024-12-27
`std::net::tcp::accept()` error handling makes it impossible to create non-blocking TCP server

rexim:
I don't really know what was the justification behind introducing `NetError.ACCEPT_FAILED`, but it obscures the actual error thrown by underlying `accept()` call which carries a lot of meaningful information. One of the important errors it throws is `EWOULDBLOCK` when you set the socket non-blocking which is crucial for asynchronous networking. Another important error `accept()` throws is `EINTR` in case of being interrupted.

All these errors are important for the programmer to distinguish. I propose to just propagate the `os::socket_error()` like `std::net::tcp::listen_to()` does.

Please let me know if this is not how you envisioned the design of the standard library. Thank you!

## How to reproduce the problem

Consider file:

```rust
// file main.c3
import std::io;
import std::net::tcp;

fn void main() {
    TcpServerSocket server = tcp::listen("localhost", 6969, 69, REUSEADDR)!!;
    server.sock.set_non_blocking(true)!!;
    while ACCEPT: (true){
        TcpSocket! client = tcp::accept(&server);
        if (catch nani = client) {
            case IoError.WOULD_BLOCK:
                io::printfn("OK");
                break ACCEPT;
            default: nani?!!;
        }
    }
}
```

Observed behavior:

```console
$ c3c compile-run main.c3
Program linked to executable 'main'.
Launching ./main

ERROR: 'Unexpected fault 'ACCEPT_FAILED' was unwrapped!'
  in std.core.builtin.default_panic (/home/rexim/opt/c3/lib/c3/std/core/builtin.c3:99) [./main]
  in std.core.builtin.panicf (/home/rexim/opt/c3/lib/c3/std/core/builtin.c3:153) [./main]
  in main.main (/home/rexim/Programming/probe/c3-tcp-accept-bug/main.c3:13) [./main]
  in @main_to_void_main (/home/rexim/opt/c3/lib/c3/std/core/private/main_stub.c3:18) [./main] [inline]
  in main (/home/rexim/Programming/probe/c3-tcp-accept-bug/main.c3:4) [./main]
  in __libc_start_call_main (./csu/../sysdeps/x86/libc-start.c:58) [/usr/lib/libc.so.6]
  in __libc_start_main_impl (./csu/../csu/libc-start.c:360) [/usr/lib/libc.so.6]
  in _start (/builddir/glibc-2.39/csu/../sysdeps/x86_64/start.S:115) [./main]
Program interrupted by signal 4.
```

Expected behavior:
```console
Program linked to executable 'main'.
Launching ./main
OK
```

### DIFF
--- a/lib/std/net/tcp.c3
+++ b/lib/std/net/tcp.c3
@@ -45,7 +45,7 @@ fn TcpSocket! accept(TcpServerSocket* server_socket)
 	TcpSocket socket;
 	socket.ai_addrlen = socket.ai_addr_storage.len;
 	socket.sock = os::accept(server_socket.sock, (SockAddrPtr)&socket.ai_addr_storage, &socket.ai_addrlen);
-	if (!socket.sock.is_valid()) return NetError.ACCEPT_FAILED?;
+	if (!socket.sock.is_valid()) return os::socket_error()?;
 	return socket;
 }
 


rexim:
As I was writing this report I realized that the intended usage of accept() is probably something like this

```rust
import std::io;
import std::net::tcp;
import std::net::os;

fn void main() {
    TcpServerSocket server = tcp::listen("localhost", 6969, 69, REUSEADDR)!!;
    server.sock.set_non_blocking(true)!!;
    while ACCEPT: (true){
        TcpSocket! client = tcp::accept(&server);
        if (catch client) {
            switch (os::socket_error()) {
                case IoError.WOULD_BLOCK:
                    io::printfn("OK");
                    break ACCEPT;
                default: unreachable();
            }
        }
    }
}
```

Since the specific error is OS-dependent you as the author of the stdlib want to make it the responsibility of the user to care about its cross-platformability. Alright, fair, enough. Sorry for wasting your time. :D

rexim:
On the other hand, why `listen_to` doesn't do the same thing? And after looking around I noticed that IoError.WOULD_BLOCK is actually fairly cross-platform. Win32 throws it to... Anyway, never mind. I at least have a workaround now.

lerno:
I think this API needs quite a bit of love. I seem to recall basing it off some C library, but with limited time to test it.

lerno:
I was debugging something and was also annoyed exactly at the detailed errors being eaten by the former layers. They shouldn't be.

lerno:
I think I *didn't* fix it then because I was just fixing a bug and wasn't doing anything serious, and I remember thinking "I'll get back to this when I have a proper use-case or I'll just make the API worse in some different way". Feel free to improve and suggest a PR.

PR #1729, opened 2024-12-27
Fix Segfault and UX of the `c3c project add-target` subcommand

rexim:
My version:

```console
c3c --version
C3 Compiler Version:       0.6.6 (Pre-release, Dec 28 2024 03:58:06)
Installed directory:       /home/rexim/opt/c3/bin/
Git Hash:                  291b26f2300dfaaf8d6c3d8890a3c1294cd39cde
```

Consider the following workflow in a completely empty folder:

```console
$ cat > main.c3 <<END
import std::io;

fn void main() {
    io::printfn("Hello, World");
}
END
$ c3c project add-target main executable
Could not open file "project.json".

$ touch project.json
$ c3c project add-target main executable
Expected a map of project information in 'project.json'.

$ echo {} > project.json
$ c3c project add-target main executable
Segmentation fault
$ Uhm... EXCUSE ME?!!
```

First of all, we have an obvious NULL-dereferencing bug. But that's not the biggest problem. The whole process is absolutely and needlessly excruciating. Why do I have to create the file myself? Why do I have to initialize it? Why add-target can't just do what I mean and create everything for me?

In this patch we

- Fix segfault on `project.json` containing empty map `{}`
- Enable `add-target` to operate on non-existing project files
- Extend `add-target` syntax to accept source through CLI args

So the whole workflow turns into

```console
$ cat > main.c3 <<END
import std::io;

fn void main() {
    io::printfn("Hello, World");
}
END
$ c3c project add-target main executable main.c3
$ cat project.json
{
  "targets": {
    "main": {
      "type": "executable",
      "sources":  [ "main.c3" ]
    }
  }
}
$ c3c run
Hello, World
```

Bam! Very smooth, feels great! Minimum changes required. Please let me know if you see any problems with this patch.

### DIFF
--- a/src/build/build.h
+++ b/src/build/build.h
@@ -453,6 +453,7 @@ typedef struct BuildOptions_
 		ProjectSubcommand command;
 		const char *target_name;
 		TargetType target_type;
+		const char **sources;
 	} project_options;
 	CompileOption compile_option;
 	TrustLevel trust_level;
--- a/src/build/build_options.c
+++ b/src/build/build_options.c
@@ -222,7 +222,8 @@ static void parse_project_subcommand(BuildOptions *options)
 		if (at_end() || next_is_opt()) error_exit("Expected a target type like 'executable' or 'static-lib'");
 		options->project_options.target_type = (TargetType)get_valid_enum_from_string(next_arg(), "type", targets, ELEMENTLEN(targets), "a target type like 'executable' or 'static-lib'");
 
-		while (!at_end() && !next_is_opt()) {
+		while (!at_end() && !next_is_opt())
+		{
 			vec_add(options->project_options.sources, next_arg());
 		}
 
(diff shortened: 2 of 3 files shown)


lerno:
So kind of for people who don't want to use `c3c init <project-name>`? 🤔 I never thought of the project update to work as a custom project init, but if you're finding this work flow then others will as well 😄

rexim:
> So kind of for people who don't want to use `c3c init <project-name>`? 🤔 I never thought of the project update to work as a custom project init, but if you're finding this work flow then others will as well 😄

Yeah, I'm really sorry. I personally find `c3c init` output too overwhelming. I'd prefer to increase the complexity gradually as needed. :D

lerno:
Only one thing: the rest of the code uses Allman brace style, so could you fix your additions to be consistent with that? 🙏

lerno:
> Yeah, I'm really sorry. I personally find c3c init output too overwhelming. I'd prefer to increase the complexity gradually as needed. :D

No it's interesting! This is evolving the product to fit different styles, and that is a good thing.

rexim:
> Only one thing: the rest of the code uses Allman brace style, so could you fix your additions to be consistent with that? 🙏

Done

lerno:
NICE Thank you!

PR #1730, opened 2024-12-28
math::nolibc: atanh 

MendocinoWhiteDeer:
Added `atanh` nolibc definitions (not to be confused with `atan`!) and more `atanh` test points in the math_tests module. `atanh` relies on `log1p`, which was also defined. These definitions were translated from FreeBSD (see [here](https://cgit.freebsd.org/src/tree/lib/msun/src)) and musl (see [here](https://cgit.freebsd.org/src/tree/lib/msun/src)). 

If anybody would like to verify that these nolibc functions are correct, you can do so in a libc environment with the feature `C3_MATH`. This replaces libc definitions with nolibc definitions when they exist and defaults to libc when they don't. If anybody wants to test the authenticity of these functions run

`c3c compile-test -D C3_MATH test/unit/stdlib/math/math.c3`


lerno:
Thank you!

PR #1737, opened 2024-12-29
Several nix fixes

vssukharev:
Recent development shell didn't really provide the working one so that it wasn't possible to build c3c manually. The commit is fixing this.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -8,12 +8,17 @@
 
   outputs = { self, ... } @ inputs: inputs.flake-utils.lib.eachDefaultSystem 
   (system: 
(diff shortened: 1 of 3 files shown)


tomaskallup:
I can build it just fine using the shell, what I'm not sure about is the comment about usage for `C3_CMAKE_FLAGS`, the way I build the binary is following:
```sh
$ mkdir build
$ cd build
$ cmake .. $C3_CMAKE_FLAGS
$ cmake --build .
```

vssukharev:
I have no idea how to do it in other way. We should declare additional information about LLD, cause cmake can't find it. Hopefully, we don't need patches: just additional environment variable.

lerno:
Thank you!

PR #1741, opened 2024-12-30
add std::net::url for parsing/generating URLs

louis77:
Tests are inferred from https://datatracker.ietf.org/doc/html/rfc3986

### DIFF
--- a/lib/std/net/url.c3
+++ b/lib/std/net/url.c3
@@ -1,0 +1,202 @@
+module std::net::url;
+
+import std::io;
+import std::collections::map;
+import std::collections::list;
+
+def ListStr = List(<String>);
+def Values = map::HashMap(<String, ListStr>);
+
+struct Url(Printable) {
+    String scheme;
+    String host;
+    uint   port;
+    String username;
+    String password;
+    String path;
+    String query;
+    String fragment;
+}
+
+fn usz! String.index_of_chars(String s, char[] needle)
+{
+    foreach (i, c : s)
+    {
+        foreach (j, pin : needle)
+        {
+            if (c == pin) return i;
+        }
+    }
+
+    return SearchResult.MISSING?;
+}
+
+fn Url! parse(String url_string) {
+    Url url = Url{};
+    url_string = url_string.trim();
+    if (url_string.len == 0) {
+        return url;
+    }
+
+    // Parse scheme
+    if (try pos = url_string.index_of("://")) {
+        url.scheme = url_string[0 .. pos-1];
+        url_string = url_string[url.scheme.len + 3..];
+    }
+    else if (url_string.contains(":")) {
+        // Handle schemes without authority like 'mailto:'
+        url.scheme = url_string[0 .. url_string.index_of(":")!-1];
+        url_string = url_string[url.scheme.len+1..];
+        url.path = url_string;
+
+        return url;
+    }
+
+    // Parse host, port
+    if (url.scheme != "urn") {
+        usz! authorityEnd = url_string.index_of_chars("/?#");
+        if (catch authorityEnd) {
+            authorityEnd = url_string.len;
+        }
+
+        String authority = url_string[0 ..  authorityEnd-1]!;
+
+        if (try usz userInfoEnd = url_string.index_of_char('@')) {
+            String userinfo = authority[..userInfoEnd-1];
+            String[] userpass = userinfo.split(":");
+            defer free(userpass);
+            url.username = userpass[0];
+            if (userpass.len > 1) {
+                url.password = userpass[1];
+            }
+            authority = authority[userInfoEnd+1..];
+        }
+
+        // Check for IPv6 address in square brackets
+        if (authority.starts_with("[") && authority.contains("]")) {
+            usz ipv6End = authority.index_of("]")!;
+            url.host = authority[0..ipv6End];  // Includes closing bracket
+            if ((ipv6End + 1) < authority.len && authority[..ipv6End] == ":") {
+                url.port = authority[..ipv6End + 1].to_uint()!;
+            }
+        } else {
+            String[] hostPort = authority.split(":");
+            defer mem::free(hostPort);
+            if (hostPort.len > 1) {
+                url.host = hostPort[0];
+                url.port = hostPort[1].to_uint()!;
+            } else {
+                url.host = authority;
+            }
+        }
+        url_string = url_string[authorityEnd ..]!;
+    }
+
+    // Parse path
+    long queryIndex = (long)url_string.index_of_char('?') ?? -1;
+    long fragmentIndex = (long)url_string.index_of_char('#') ?? -1;
+
+    if (queryIndex != -1 || fragmentIndex != -1) {
+        long pathEnd = min(queryIndex == -1 ? url_string.len : queryIndex,
+                           fragmentIndex == -1 ? url_string.len : fragmentIndex,
+                           url_string.len);
+        url.path = url_string[0 .. (pathEnd-1)];
+    } else {
+        url.path = url_string;
+    }
+
+    // Remove the path part from url for further parsing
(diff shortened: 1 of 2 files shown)


lib/std/core/string.c3
```diff
@@ -362,6 +362,29 @@ fn usz! String.index_of_char(s, char needle)
 	return SearchResult.MISSING?;
 }
 
+<*
+ Find the index of the first incidence of a one of the chars.
+
+ @param [in] s
+ @param needle "The characters to look for"
```
lerno, line 369: needle can also be [in] since it is not mutated.

lib/std/net/url.c3
```diff
@@ -0,0 +1,189 @@
+module std::net::url;
+
+import std::io;
+import std::collections::map;
+import std::collections::list;
+
+def ListStr = List(<String>);
+def Values = map::HashMap(<String, ListStr>);
+
+struct Url(Printable) {
```
lerno, line 10: The stdlib uses Allman brace style, so please change it to make it consistent.

lib/std/net/url.c3
```diff
@@ -0,0 +1,189 @@
+module std::net::url;
+
+import std::io;
+import std::collections::map;
+import std::collections::list;
+
+def ListStr = List(<String>);
+def Values = map::HashMap(<String, ListStr>);
```
lerno, line 8: `map::HashMap` can be replaced by simply `HashMap`

lib/std/net/url.c3
```diff
@@ -0,0 +1,189 @@
+module std::net::url;
+
+import std::io;
+import std::collections::map;
+import std::collections::list;
+
+def ListStr = List(<String>);
```
lerno, line 7: It is more common in the stdlib to swap these, so `StrList` or `StringList`

lib/std/net/url.c3
```diff
@@ -0,0 +1,189 @@
+module std::net::url;
+
+import std::io;
+import std::collections::map;
+import std::collections::list;
```
lerno, line 5: These three can be collected in a single import:
`import std::io, std::collections::map, std::collections::list;`

lib/std/net/url.c3
```diff
@@ -0,0 +1,189 @@
...
+    String password;
+    String path;
+    String query;
+    String fragment;
+}
+
+fn Url! parse(String url_string) {
+    Url url = Url{};
+    url_string = url_string.trim();
+    if (url_string.len == 0) {
```
lerno, line 24: More idiomatic `if (!url_string) return url;`

lib/std/net/url.c3
```diff
@@ -0,0 +1,189 @@
...
+        return url;
+    }
+
+    // Parse scheme
+    if (try pos = url_string.index_of("://")) {
+        url.scheme = url_string[0 .. pos-1];
```
lerno, line 30: More idiomatic: `url.scheme = url_string[:pos];`

lib/std/net/url.c3
```diff
@@ -0,0 +1,189 @@
...
+        url_string = url_string[url.scheme.len + 3..];
+    }
+    else if (url_string.contains(":")) {
+        // Handle schemes without authority like 'mailto:'
+        url.scheme = url_string[0 .. url_string.index_of(":")!-1];
```
lerno, line 35: Similar here (and later): if you find yourself writing `a[0..y - 1]` you can instead write this as `a[:y]`

lib/std/net/url.c3
```diff
@@ -0,0 +1,189 @@
...
+        url_string = url_string[url.scheme.len+1..];
```
lerno, line 36: Spacing around the operators to match stdlib code style.

lib/std/net/url.c3
```diff
@@ -0,0 +1,189 @@
...
+            defer free(userpass);
+            url.username = userpass[0];
+            if (userpass.len > 1) {
+                url.password = userpass[1];
+            }
+            authority = authority[userInfoEnd+1..];
+        }
+
+        // Check for IPv6 address in square brackets
+        if (authority.starts_with("[") && authority.contains("]")) {
```
lerno, line 63: Consider replacing if / else with a switch

lib/std/net/url.c3
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    return url;
+}
+
+fn String Url.to_string(Url* self, Allocator allocator = allocator::temp()) @dynamic
+{
+    @pool(allocator)
+	{
+        DString builder = dstring::new("", allocator);
```
lerno, line 120: Use the temp allocator for the dstring instead?

lib/std/net/url.c3
```diff
@@ -0,0 +1,189 @@
...
+            builder.append_chars(self.query);
+        }
+
+        // Add fragment if it exists
+        if (self.fragment != "") {
+            builder.append_char('#');
+            builder.append_chars(self.fragment);
+        }
+
+        return builder.str_view();
```
lerno, line 164: This is returning a str_view from a freed builder and is a use after free.
Instead, make the dstring a temp dstring using temp_init. Then 
`return builder.copy_str(allocator);`

lib/std/net/url.c3
```diff
@@ -0,0 +1,189 @@
...
+    };
+}
+
+fn Values Url.query_values(Url* self, Allocator allocator = allocator::heap()) {
+    Values vals = {};
+
+    String[] raw_vals = self.query.split("&");
```
lerno, line 171: Consider using Splitter using `self.query.splitter("&")` instead. You might choose .tokenizer if you want to skip empty parts (eg `&&`) from being picked up.

lib/std/net/url.c3
```diff
@@ -0,0 +1,189 @@
...
+            builder.append_char('#');
+            builder.append_chars(self.fragment);
+        }
+
+        return builder.str_view();
+    };
+}
+
+fn Values Url.query_values(Url* self, Allocator allocator = allocator::heap()) {
+    Values vals = {};
```
lerno, line 169: You need to initialize the hash map here with the allocator, so use:
```c
Values vals;
vals.init(allocator);
```

lib/std/core/string.c3
louis77: Done.

lib/std/net/url.c3
louis77: Done.

lib/std/net/url.c3
louis77: Done.

lib/std/net/url.c3
louis77: Done.

lib/std/net/url.c3
louis77: Done.

lib/std/net/url.c3
louis77: Done.

lib/std/net/url.c3
louis77: Done.

lib/std/net/url.c3
louis77: Done.

lib/std/net/url.c3
louis77: Done.

lib/std/net/url.c3
louis77: Done.

lib/std/net/url.c3
louis77: Done, I used the tokenizer.

lib/std/net/url.c3
louis77: Done.

lib/std/net/url.c3
louis77: As discussed, for future consideration.

PR #1748, opened 2024-12-30
add std::net::url - with fixes

louis77:


lib/std/net/url.c3
```diff
@@ -0,0 +1,240 @@
...
+		url.path = url_string;
+
+		return url;
+	}
+
+	// Parse host, port
+	if (url.scheme != "urn")
+	{
+		usz! authorityEnd = url_string.index_of_chars("/?#");
+		if (catch authorityEnd)
```
lerno, line 56: One more style issue. The stdlib uses snake_case for variables and methods. So `authority_end` would be correct. (And similar changes in the rest of the code)

louis77: Done.

lerno:
Have you tested that there is no leak now? You can use `mem::report_heap_allocs_in_scope() { .. }:` to do so.

lerno:
Looks, fine, just one thing. The query values... to free those currently you first would need to take each value in the list and free it, then free "Values". I think it would be better to wrap this in a distinct type, which has a `.free()` which will free everything. And then it needs to be new_query_values() because the convention is to add `new` when the function is by default uses the heap allocator.

Essentially there are three possible conventions to follow when a function returns something that needs to be freed:
1. `copy_foo` this one should have an allocator, which defaults to the heap, it can be paired with `tcopy_foo` which uses the temp allocator.
2. `new_foo` (optionally also `temp_foo`), new_foo uses the heap allocator by default as a default argument.
3. `foo` `new_foo` `temp_foo`, this triad has `foo` explicitly taking the allocator (no default), then new_foo and temp_foo taking the heap allocator and temp allocator respectively.

Let me know if you want me to fix this part.

PR #1749, opened 2024-12-30
Refactor Posix NativeThread to be able to free allocated memory

louis77:
As discussed, the NativeThread implementation for Posix allocates memory on create(), which is passed to the pthread_create() function. However, since no reference to it was retained, the memory leaked after the thread finished. 

See: https://github.com/c3lang/c3c/blob/ad9cfcdcc7bf7fd8ac27803e81e0e596147e3194/lib/std/threads/os/thread_posix.c3#L161

According to my research, there's no way to get a reference to this memory block back from pthread, so I changed NativeThread so that a reference to this memory can be stored and added a `destroy()` method which ultimately can free the memory.

For the Win32 implementation, there is no special allocation. `destroy()` is therefore a no-op for the moment.

In my tests the leak disappeared and threads still worked.

This PR should just serve as a proposal, since I'm not so sure if I'm fully aware of the potential implications of this change. Please feel free to dismiss or ignore.

### DIFF
--- a/lib/std/threads/os/thread_posix.c3
+++ b/lib/std/threads/os/thread_posix.c3
@@ -8,8 +8,11 @@ struct NativeMutex
 }
 
 def NativeConditionVariable = Pthread_cond_t;
-def NativeThread = Pthread_t;
 def NativeOnceFlag = Pthread_once_t;
+struct NativeThread {
+	Pthread_t pthread;
+	PosixThreadData* data;
+}
 
 <*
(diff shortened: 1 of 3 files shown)


lerno:
This changes semantics and size of the thread object. I am not sure why my proposed solutions would not work.

lerno:
I solved this a bit differently. You can see that I store the data in a thread local, then destroy the memory on thread exit.

louis77:
Your solution is much better and my tests show no leaks, thank you! 🚀

PR #1752, opened 2025-01-01
Fix `native_fputc`

radekm:
Since C `libc::fputc` returns code of character which was written the original code wasn't working properly when writing 0 character.


lerno:
Thank you for the fix!

PR #1762, opened 2025-01-03
socket.c3: optimize poll() duration wrapping

ecki:
I saw you fixed it in response to @rexim Video, but I think switching the wrapper would make it a bit more efficient since it avoids a time::ms()/Duration::to_ms() roundtrip for These who prefer the raw long. I can understand to use the more expressive types for the actual implementation, but in the specific case Duration isnt even the best fit.

I also made sure POLL_FOREVER and the Clint.max Champing is documented.

BTW do you also need to clamp for underflows or maybe have a (timeout_ms<0)-1:(timeout_ms<CInt.max)?(CInt)timeout_ms:CInt.max) instead?

this is only a draft PR for discussion

### DIFF
--- a/lib/std/net/socket.c3
+++ b/lib/std/net/socket.c3
@@ -56,9 +56,9 @@ struct Poll
  @param [inout] polls
  @param timeout "duration to poll (clamped to CInt.max ms), or POLL_FOREVER."
 *>
-fn ulong! poll_ms(Poll[] polls, long timeout_ms)
+fn ulong! poll_ms(Poll[] polls, Duration timeout)
 {
-	return poll(polls, time::ms(timeout_ms)) @inline;
+	return poll(polls, timeout.to_ms()) @inline;
 }
 
 <*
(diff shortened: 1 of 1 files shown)


ecki:
If you dont Like the Parameter Reading I can also use `CInt time_ms =` instead, that also remove more casts

lerno:
Hmm.. I think you changed the function signature, in a non-backwards compatible way and then none of the CI passed.

I am not sure why you changed `poll_ms` to take a Duration and `poll` to take a long in being the time in ms? That seems backwards?

ecki:
The signatures stayed the same, they just switched. (At least they should, but I might be missing some c3 specific Details). But let me get it test compile locally.

Die the CI run on the initial Patch or the fixed one, it was not complete on the first try.

lerno:
Yes, I just don't understand why `poll_ms` would take a Duration and `poll` would take a long with ms?

ecki:
Yes I fixed the problem now. I can squash the commits when CI succeeds.

what do you think about clamping negative values, too?

lerno:
Might be overkill.

lerno:
Thank you!

PR #1763, opened 2025-01-03
build: add /usr/lib to LLVM_LIB search paths

ThomasAdam:
It seems that on Alpine Linux (Edge), the LLVM_LIBRARY_DIRS cmake variable doesn't look in /usr/lib, which is where the relevant files are on Alpine.

Only do this for !WIN32 systems.


lerno:
Looks like this worked fine in CI, so let's add it.

lerno:
Thank you!

PR #1767, opened 2025-01-03
Additional convenient functions and enums

rtc11:
Added some missing API calls and enums when working with the Objective-C Runtime.

These are useful functions and enums when doing trivial stuff like managing Window, System tray,  Menus or creating additional methods on classes.


lerno:
I think some typo happened, see the CI

rtc11:
Another typo, my bad. Will fix it when the kids are not hanging around my neck!

lerno:
Looking good now. Thank you!

PR #1770, opened 2025-01-04
libc: add isatty()

ThomasAdam:
This a trivial PR (perhaps a selfish one?) to add the libc isatty() call as I'm needing it for a different module I'm intending to write.

### DIFF
--- a/lib/std/libc/libc.c3
+++ b/lib/std/libc/libc.c3
@@ -109,6 +109,7 @@ extern fn ZString getenv(ZString name);
 extern fn ZString gets(char* buffer);
 extern fn Tm* gmtime(Time_t* timer);
 extern fn Tm* gmtime_r(Time_t *timer, Tm* buf) @if(!env::WIN32);
+extern fn CInt isatty(Fd fd) @if(!env::WIN32);
 extern fn CLong labs(CLong x);
 extern fn LongDivResult ldiv(CLong number, CLong denom);
 extern fn Tm* localtime(Time_t* timer);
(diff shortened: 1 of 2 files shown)


ThomasAdam:
@lerno,

This does raise the question that there's a slew of potential libc::* exports we could add.  You'll note I've not added ttyname() or ttyslot(), even though I could have -- but because I'm personally not using them, I didn't bother.

Is there some policy around this, or are libc:: additions done on an as-and-when basis?

Also, I've not added an os-native version of isatty() -- I could, as it's just using ioctl() under-the-hood -- when would you typically want to see os-native versions of libc syscalls?

lerno:
I think the more we can get into libc the better service it is for people who use the language in the beginning. Some care has to be taken as to make sure it is convenient to use cross platform. So that you added it for win32 as well is great.

So the short answer is: yes, adding more is better. I haven't done more because adding something means also checking that it's right, so I've been adding things that the are needed for the stdlib.

Which kind of raises a question whether not `libc::istty` should be added to std::io.

Something like `io::stdout().is_tty()` + a convenience function `io::is_tty()` which checks that stdout and stdin is tty

ThomasAdam:
> Which kind of raises a question whether not `libc::istty` should be added to std::io.
> 
> Something like `io::stdout().is_tty()` + a convenience function `io::is_tty()` which checks that stdout and stdin is tty

Interesting.  We could look at this after-the-fact.  Usually, such wrapper functions would only care to ensure the output isn't being sent to something which wouldn't understand various escape sequences, or is looking for interactive input when there isn't one, etc.  I guess we could add this to std::io -- it depends whether that's useful for consumers of a isatty() check, versus doing that themselves.

lerno:
Thank you for the addition!

PR #1773, opened 2025-01-05
Add String.trim_left() / right()

louis77:
Proposal, I hope that's alright.

### DIFF
--- a/lib/std/core/string.c3
+++ b/lib/std/core/string.c3
@@ -177,7 +177,7 @@ fn String String.trim_left(string, String to_trim = "\t\n\r ")
 fn String String.trim_right(string, String to_trim = "\t\n\r ")
 {
 	usz len = string.len;
-	while (len > 0 && char_in_set(string[len-1], to_trim)) len--;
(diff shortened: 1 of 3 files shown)


lib/std/core/string.c3
```diff
@@ -145,14 +145,40 @@ fn String join_new(String[] s, String joiner, Allocator allocator = allocator::h
...
+
+ @param [in] string `The string to trim`
+ @param [in] to_trim `The set of characters to trim, defaults to whitespace`
+ @pure
+ @return `a substring of the string passed in`
+*>
+fn String String.trim_right(string, String to_trim = "\t\n\r ")
+{
+	usz len = string.len;
+	while (len > 0 && char_in_set(string[len-1], to_trim)) len--;
```
lerno, line 180: Just the spacing here to `[len - 1]`. Yes, a bit anal, but consistency is important in the stdlib.

lerno:
Looks fine! Just that small style fix...

lib/std/core/string.c3
louis77: Done

lerno:
Thank you!

PR #1775, opened 2025-01-05
Fix 'clean' command on Windows MinGW

KillerxDBr:
'c3c clean' command try to use 'rm' on Windows MinGW compiled executable, since it does not define "_MSC_VER", but define "_WIN32"
There are other uses of "_MSC_VER" in the same file, some can probably be replaced by "_WIN32" too.


lerno:
Thank you!

Issue #1632, opened 2024-11-18
math_nolibc is missing implementations for: exp, log, pow


chris-rcn:
This has implications for WASM targets, since that doesn't use libc.

lerno:
These should now have been added @chris-rcn, can you verify?

PR #1781, opened 2025-01-05
Feat: Added exp, log and pow functions as requested in #1632

Adversing:


lerno:
When running without libc, it looks like some things doesn't compile. Try --use-libc=no and see what happens

lerno:
Do you need help getting this to work?

lerno:
When compiling this it seems to be missing essential functions for the implementation. frexpr for example.

MendocinoWhiteDeer:
I am looking through the errors. In addition to missing functions mentioned by @lerno you are using NAN and INFINITY which the compiler doesn't like.
You should replace it with `float.nan`, `float.inf`, `double.nan`, `double.inf` depending on whether you need the float or double.

Adversing:
@MendocinoWhiteDeer the constants were already replaced by @lerno in one of his commits on this PR. I'm currently working on the implementation of the missing functions.

MendocinoWhiteDeer:
> @MendocinoWhiteDeer the constants were already replaced by @lerno in one of his commits on this PR. I'm currently working on the implementation of the missing functions.

Oops for some reason I was looking at the old build. Thought I was reading the recent one!!

lerno:
Thank you!

PR #1782, opened 2025-01-06
Date/Time formatters

louis77:
There are multiple ways to approach this:

1. Rely on libc's `strftime` and fix what it can't do to implement necessary formats
2. Use Go's approach of a domain-specific layout language (i.e. "Mon, 02 Jan 2006 15:04:05 GMT"), which gives a nice user experience and very flexible formatting options, but involves of string manipulation
3. Just define a set of pre-defined popular formats (mostly RFCs) that are used for stuff like HTTP servers, etc. and implement it in the most efficient way, without relying on libc and too much string manipulation.

This is an initial implementation of (3).

I have a working implemention of (2) also available. While I finalized it, I felt like it was too much of a departure from C (kind of just "because we can"), and since the user always can revert to `strftime`, it is basically a reinvention of what already exists.

The formats implemented are:
```
- ANSIC,	// "Mon Jan _2 15:04:05 2006"
- UNIXDATE,	// "Mon Jan _2 15:04:05 GMT 2006"
- RUBYDATE,	// "Mon Jan 02 15:04:05 -0700 2006"
- RFC822,	// "02 Jan 06 15:04 GMT"
- RFC822Z,	// "02 Jan 06 15:04 -0700"
- RFC850,	// "Monday, 02-Jan-06 15:04:05 GMT"
- RFC1123,	// "Mon, 02 Jan 2006 15:04:05 GMT"
- RFC1123Z,	// "Mon, 02 Jan 2006 15:04:05 -0700"
- RFC3339,	// "2006-01-02T15:04:05Z"
- RFC3339Z,	// "2006-01-02T15:04:05+07:00"
- RFC3339MS,	// "2006-01-02T15:04:05.999999Z"
- RFC3339ZMS,	// "2006-01-02T15:04:05.999999+07:00"
- DATETIME,	// "2006-01-02 15:04:05"
- DATEONLY,	// "2006-01-02"
- TIMEONLY,	// "15:04:05"
```

These are the same formats that Go's std lib implements and cover most use cases.

Additional notes:

- the `format()` method is implemented on the Enum `FormatSpec` itself, and `DateTime.format()`, `TzDateTime.format()` redirect to it
- a `tformat()` variant is also available
- when the user has a `DateTime` object without a time zone, it is automatically assumed to be GMT-00:00
- formats which use a timezone abbreviation (GMT, Z) are always moved to offset 0 - because most RFCs require the timezone to be GMT/Z and not any arbitratry one
- when the offset itself is part of the format, it is not moved. In any case, the date/time stays factually correct

While I wait for some feedback, I plan to add some more tests. If another approach is preferred, I have those in my archive and could easily switch to them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,3 +79,5 @@ TAGS
 # 'nix build' resulting symlink
 result
 
+# macOS
+.DS_Store
--- a/lib/std/time/datetime.c3
+++ b/lib/std/time/datetime.c3
@@ -11,7 +11,7 @@ fn DateTime now()
  @require hour >= 0 && hour < 24
  @require min >= 0 && min < 60
  @require sec >= 0 && sec < 60
- @require us >= 0 && us < 999_999
+ @require us >= 0 && us <= 999_999
 *>
 fn DateTime from_date(int year, Month month = JANUARY, int day = 1, int hour = 0, int min = 0, int sec = 0, int us = 0)
 {
@@ -25,7 +25,7 @@ fn DateTime from_date(int year, Month month = JANUARY, int day = 1, int hour = 0
  @require hour >= 0 && hour < 24
  @require min >= 0 && min < 60
  @require sec >= 0 && sec < 60
- @require us >= 0 && us < 999_999
+ @require us >= 0 && us <= 999_999
  @require gmt_offset >= -12 * 3600 && gmt_offset <= 14 * 3600
 *>
 fn TzDateTime from_date_tz(int year, Month month = JANUARY, int day = 1, int hour = 0, int min = 0, int sec = 0, int us = 0, int gmt_offset = 0)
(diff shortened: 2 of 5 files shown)


lerno:
Something broke I think?

louis77:
I see that `to_gmt_offset` relies on libc, which is not present on WASM. This probably answers also my question about approach 1 (relying on `strftime`) - with a clear *no*. I'll continue to work on it and try to remove dependency on libc.

lerno:
Is this still a draft?

lerno:
So.. ready?

louis77:
Ready.

lerno:
We might complement this with adding YYYY mm and such formatting to the formatter, but I think this is a good start.

I did some refactoring, for example it could be useful to have the names as associated values instead of as a separate method. I added days in each month as a convenience for later.

PR #1791, opened 2025-01-06
Add Nix build and checks to CI/CD

vssukharev:
The following workflow doesn't create any binaries and just tests build under Nix.


lerno:
Thanks!

PR #1792, opened 2025-01-06
mem: add macro to assert on memory leak in scope

konimarti:
Implement `mem::@assert_leak` to assert on a memory leak in the scope of the macro body. Memory report for the leak can be disabled by setting the boolean argument to false.


lerno:
You probably don't want to run this unless it's in env::DEBUG == true, so my suggestion is to check that and only if env::DEBUG is true OR $feature(MEMORY_ASSERTS) is true this tracking is done.

alexveden:
I accidentally found that `--sanitize=address` also checks memory leaks! Also, it checks out of bounds access, which is cool!

image

PR #1795, opened 2025-01-07
net/url: implement url encoding (RFC 3986)

konimarti:
Implement url percent-encoding and -decoding functions according to RFC 3986.

Link: https://datatracker.ietf.org/doc/html/rfc3986

An Url can be parsed from a String with `new_parse()` or `temp_parse()`. The parsed fields are decoded. The only field that is not decoded is `raw_query`. To access the decoded query values, use `Url.query_values()`.

`Url.to_string()` will re-assemble the fields into a valid Url string with proper percent-encoded values.

If the Url struct fields are filled in manually, use the actual (un-encoded) values. To create a raw query string, initialize an `UrlQueryValues` map, use `UrlQueryValues.add()` to add the query parameters and, finally, call `UrlQueryValues.to_string()`.

CC: @louis77 as discussed on discord and to keep you in the loop.


lerno:
Hmm.. this becomes quite a bit larger, I have to take a longer look at this one. But for now just a small thing: `new_parse` and `temp_parse` would be the proper names to conform with the others.

konimarti:
Yes, please take your time. I'll change the names in the meantime.

lerno:
It's quite the pity that the non-allocating solution suddenly has so much allocation happening. The whole thing with the query values being separate is because Url wasn't allocating anything too. So now it makes more sense to pull that into URL

konimarti:
I'll close this PR to think more about a better memory model for this approach.

lerno:
No need to close it. We can discuss improvements.

Issue #1713, opened 2024-12-24
Fix build tools detection for Windows


SunP04:
Hello.

Currently, for the windows build, the compiler is trying to find the [vswhere](https://github.com/microsoft/vswhere) tool, which could be seen [here](https://github.com/c3lang/c3c/blob/master/src/utils/find_msvc.c#L38).

However that breaks when the user installs only the VS Build Tools (not the editor), which `vswhere` cannot detect, as it's not a complete installation of Visual Studio. Which forces most people to have a second installation just for c3, which is just a bit dumb (sry for the wording, but there isn't a better word there).

An easy fix would be just to try to find `vcvarsall.bat` instead of `vswhere`, which is installed by default at `{{Program Files}}\Microsoft Visual Studio\2022\Community\VC\Auxiliary\Build\vcvarsall.bat` (Program Files there could be x86 or not). That is also installed by the editor and it's the "default" way other compilers try to know if the user has the Build Tools installed (Examples with [Odin](https://github.com/odin-lang/Odin/blob/master/misc/shell.bat#L8), [Rust](https://github.com/rust-lang/cc-rs/blob/main/src/windows/find_tools.rs#L431)).

This also means the python script could be better optimized to install only the necessary build tools, not the whole editor, but that's another "issue" (not really an issue).

Thanks in advance, Sun.

lerno:
The python script does not install VS, nor does it install the full build tools. It just installs enough to build with Windows.

pekochan069:
msvc build tools can be installed using https://visualstudio.microsoft.com/visual-cpp-build-tools/ which doesn't include full vs so it is preferred when only build tools are needed.
But when I install build tools this way, c3c doesn't detect msvc toolchain

lerno:
@pekochan069 until someone submits a PR to help that detection, you can use `--win-vs-dirs` to point to msvc buildtools.

lerno:
I think this is better now, please check.

PR #1796, opened 2025-01-07
windows: detect MSVC cpp build tools

pekochan069:
I'm not familiar with C so the code is pretty messy. But it solves problem addressed in #1713.
I have tested on my local windows 11 machine with msvc version 17.13.35617.110 and on a clean install of Windows Sandbox with latest version of https://visualstudio.microsoft.com/visual-cpp-build-tools/


lerno:
If you can add that, then sure.

lerno:
This looks like it worked fine in CI 🎉 So let's add it!

lerno:
Thank you!

Issue #1774, opened 2025-01-05
What about making benchmark and test runner functions accept arguments?


alexveden:
As I understood the code generation, the unit test and benchmark runners are converted to `bool main()` functions when `testrun` program is built.

Is it possible to refactor default signature to `bool testfn(String[] args)` in order to extend passing custom and **arbitrary** arguments to test runners. So that test runners may act as CLI with more flexibility.

Default, signature looks like this:
https://github.com/c3lang/c3c/blob/07c59e6a6c6ccde946b81369cd38780631905214/lib/std/core/runtime.c3#L242

Eventually, I'd like to be able to add some customizable behavior for test runners, at least these two options: https://github.com/c3lang/c3c/issues/1753#issuecomment-2568720639

lerno:
Yes, that would certainly be possible.

lerno:
How would these arguments be passed into the test runner from the compiler?

alexveden:
> How would these arguments be passed into the test runner from the compiler?

Technically speaking, the test runner is a separate program, in `build/testrun` so it can accept any arguments into main after it was compiled.  As I understand code, the `c3c` just calls `build/testrun` program via system call, so `c3c` may filter its arguments and include only `--test-` and pass them directly to the `testrun`.

However, more interesting possibilities may be opened when using side test runners. The test runner may use `c3c` CLI to build (say using --testfn parameter to override default test runner logic) and pass it's custom arguments to the `build/testrun`. So then using `--testfn` the c3c compiler essentially replaces test runner logic (main() function), and it would be cool to pass any arbitrary set of arguments to test runners.

lerno:
This is now implemented, but it's lacking `--` because I don't know if this is the right choice or if we should have some arguments in c3c that are dedicated to the tests.

In any case, a test will produce a binary, and the arguments to that binary is now available to the test runner.

alexveden:
@lerno Awesome!

> This is now implemented, but it's lacking -- because I don't know if this is the right choice or if we should have some arguments in c3c that are dedicated to the tests.

I think `--` is necessary because, it harmonizes behavior of arguments when we run tests by `c3c test -- --opt -h args` and directly `./build/testrun --opt -h args`

I opened new PR if I may.

PR #1803, opened 2025-01-10
allowing `c3c test -- <args1>`

alexveden:
A small improvement for #1774

Allowing custom and default test/bench runners receiving arguments:
```zig
fn bool default_test_runner(String[] args)
{
	io::printfn("default_test_runner: %s", args);
	return false;
}
```

```
> c3c test --testfn "std::core::test::default_test_runner" -- foo --bar

Program linked to executable 'build/testrun'.
Launching ./build/testrun foo --bar
default_test_runner: [./build/testrun, foo, --bar]
```

The same behavior is consistent when we do direct run of compiled test runner:
```
> ./build/testrun foo --bar

default_test_runner: [./build/testrun, foo, --bar]
```

We might be getting help via:
```
c3c test -- --help
```

This will decouple @tests/@benchmark runner logic completely from c3c compiler space into c3 (stdlib space) or custom runner space.

### DIFF
--- a/releasenotes.md
+++ b/releasenotes.md
@@ -27,6 +27,7 @@
 - Allow test runners to take String[] arguments.
 - Added `--lsp` output.
 - Improve the error message when running out of memory.
+- Allowed passing arguments to @test / @benchmark runners via `c3c test[benchmark] -- -o --opt1 <arg1>`
 
 ### Fixes
 - Fix case trying to initialize a `char[*]*` from a String.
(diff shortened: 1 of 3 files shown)


alexveden:
A show case how it can work:
```
~/code/c3test ➜ c3c test --testfn "std::core::test::default_test_runner" -- --help

Program linked to executable 'build/testrun'.
Launching ./build/testrun --help

Usage:
testrun [options]

c3c default test runner

Test options
    -h, --help            show this help message and exit
    -f, --filter=<str>    filter tests by name, e.g. some::test, or some [default: ]
    -b, --breakpoint      emits breakpoint if any assert() is failed [default: false]
    --sort                sort all tests alphabetically by name, --no-sort - preserve natural order [default: true]
    -c, --color           test output with colors [default: false]
Program completed with exit code 1.

```

lerno:
Looks good, although I assume we'd prefer direct arguments sent to the test runner in the long run.

Issue #1753, opened 2025-01-01
Improving unit test developer experience


alexveden:
Hi @lerno, thank you for a such great language, and I really appreciate the clarity and simplicity of c3 code. I'd like to discuss direction of unit testing part of the language and how to improve developer productivity using c3 test driven development approach. 

I'd like to improve the unit testing experience of the c3, and I wanted to know your vision of the problem I'd like to solve, or if you consider them as problems at all.

## List of improvements I think would be useful in unit test experience
(The ideas below are the subject of discussion, let me know what you think)

### Setup / teardown test functions
It would be great to have setup/teardown functions for test modules (i.e. marked as `module test_something @test;`). Because in production code, sometimes we need to setup environment (DB, socket, mocks) for each test case in a suite and make some cleanup afterward.
Example:
![image](https://github.com/user-attachments/assets/9c3ca14d-c38b-4f02-95c0-c13c91bb234a)

### Special test asserts
Using built-in `assert` maybe is not optimal for productive test driven development.  For example, failure of this assert `assert(6 == bar::mul(2,3));`  produces an error:
```
 Error: Assert violation
  - in test_sub test_foo.c3:19.
```
However, it's not clear what value was returned by `bar::mul(2,3))`, to figure it out we need to add formatting to assert, like this: `assert(6 == bar::mul(2,3), "failed: %s", bar::mul(2,3));`. In my experience, maintaining of this approach quickly becomes a hell. So my proposal is to introduce special test asserts which include expected/actual value printing.  

Conceptually, it could be a special macro for this, i.e. `tassert_eq(a, b);` or maybe test suite struct. Concept code which compiles and works below:
![image](https://github.com/user-attachments/assets/67fd8214-1ce8-4a59-a8ee-351ffc2d4781)

Having this value printouts on failures, significantly improving development and debugging velocity. And the developer gets ideas of what's going wrong much faster, with this error message: 
```
 Error: a[3] != b[2]
  - in test_sub test_foo.c3:12.
```

### Code asserts should produce backtrace
Debugging assert failure inside test function might be trivial, but when assert (or `@require`/`@ensure`) was triggered somewhere deep in the guts of the project, it's not trivial what caused the failure. 
Simple example:
![image](https://github.com/user-attachments/assets/f9ebf922-5872-4f59-8d2c-d77d1c9d5c1d)

test_add fails with this obscure error:
```
Testing test_foo::test_add .. [error]

 Error: Assert violation
  - in add foo.c3:5.
```

[...]

PR #1808, opened 2025-01-11
Test runner

alexveden:
This is a huge PR which covers #1753, and more.

What was done:
- New `std::core::test` module for unit testing machinery
- Completely refactored unit test default runner, added extra optional arguments (colors, test case filters, breakpoints, etc..., see `c3c test -- --help`)
- Added stdout/stderr output hijacking, it will be printed out only when test fails
- Added new `std::os::argparse::` module for CLI and arguments parsing, with sub-commands support
- Allowed passing arguments to @test / @benchmark runners via `c3c compile-test -- -o --opt1 <arg1>`


lerno:
Sorry for the late review. I started looking at this but found the output looks different from your mockups:
```
---------------------------------------- TESTS -----------------------------------------

ERROR: 'Assert violation'
  in std.core.builtin.print_backtrace (/Users/lerno/Projects/c3c/lib/std/core/builtin.c3:70) [testrun]
  in std.core.runtime.test_panic (/Users/lerno/Projects/c3c/lib/std/core/runtime_test.c3:60) [testrun]
  in any_tests.any_compare (/Users/lerno/Projects/c3c/test/unit/regression/any.c3:9) [testrun]
  in std.core.runtime.run_tests (/Users/lerno/Projects/c3c/lib/std/core/runtime_test.c3:227) [testrun]
  in std.core.runtime.default_test_runner (/Users/lerno/Projects/c3c/lib/std/core/runtime_test.c3:274) [testrun]
  in main (/Users/lerno/Projects/c3c/lib/std/core/runtime_test.c3:269) [testrun]
  in start (source unavailable) [/usr/lib/dyld]
[FAIL] any_tests::any_compare ( any.c3:9 ) Assert violation
[PASS] any_tests::test_aliasing
[PASS] arraytests::concat
[PASS] arraytests::find
[PASS] arraytests::find_subarray
[PASS] atomic::add
[PASS] atomic::div
```

In this case I'm also assuming that you are writing the `[FAIL]` and `[PASS]` after the test has run. This is something I took great pains *to avoid*. The reason for this is if you have a test that is for some reason deadlocking or something similar in that case you want to se:

```
Running arraytests::find [OK]
Running sometest::blurb ...
```
So that it's obvious which test is breaking.

So ideally like this:
1. Print something like "Running foo ..."
2. If there is an output, print it *below* but similar to what you did, so `> output:`
3. When the test ends we have four cases
4. The test succeeded, and it didn't output anything. just write [OK] or [PASS] or something.
5. If the test succeeded but with output, there are two options: (1) we delete the output and then do like (4)
  *or* we copy the text (1) one more time and add an OK to that line
6. If the test failed but didn't output anything, write [FAIL] and print information about it (stack trace etc)
7. If the test failed and output something, preferrably write [FAIL] keep the output and print the information below the output.


So to show some examples:
```
(4)

Running foo... [OK]

(5)
Running foo...
> Output:  Hello world
Running foo... [OK]

(6)
Running foo... [FAIL] (any.c3:9) Assert violation
 Backtrace:
  in std.core.builtin.print_backtrace (/Users/lerno/Projects/c3c/lib/std/core/builtin.c3:70) [testrun]
  in std.core.runtime.test_panic (/Users/lerno/Projects/c3c/lib/std/core/runtime_test.c3:60) [testrun]
  in any_tests.any_compare (/Users/lerno/Projects/c3c/test/unit/regression/any.c3:9) [testrun]
  in std.core.runtime.run_tests (/Users/lerno/Projects/c3c/lib/std/core/runtime_test.c3:227) [testrun]
  in std.core.runtime.default_test_runner (/Users/lerno/Projects/c3c/lib/std/core/runtime_test.c3:274) [testrun]
  in main (/Users/lerno/Projects/c3c/lib/std/core/runtime_test.c3:269) [testrun]
  in start (source unavailable) [/usr/lib/dyld]

(7)
Running foo... [FAIL] (any.c3:9) Assert violation
> Output:  Hello world
 Backtrace:
  in std.core.builtin.print_backtrace (/Users/lerno/Projects/c3c/lib/std/core/builtin.c3:70) [testrun]
  in std.core.runtime.test_panic (/Users/lerno/Projects/c3c/lib/std/core/runtime_test.c3:60) [testrun]
  in any_tests.any_compare (/Users/lerno/Projects/c3c/test/unit/regression/any.c3:9) [testrun]
  in std.core.runtime.run_tests (/Users/lerno/Projects/c3c/lib/std/core/runtime_test.c3:227) [testrun]
  in std.core.runtime.default_test_runner (/Users/lerno/Projects/c3c/lib/std/core/runtime_test.c3:274) [testrun]
  in main (/Users/lerno/Projects/c3c/lib/std/core/runtime_test.c3:269) [testrun]
  in start (source unavailable) [/usr/lib/dyld]
```

lerno:
I realize that this will only be possible when a TTY is available, so there needs to be two modes essentially.

lerno:
I think the non TTY mode can look different and more rudimentary, like:
```c
Running foo ----
Result: OK
```

alexveden:
> In this case I'm also assuming that you are writing the [FAIL] and [PASS] after the test has run. This is something I took great pains to avoid. The reason for this is if you have a test that is for some reason deadlocking or something similar in that case you want to se:

Actually, it's a great point. The only reason I decided to print test result past the output, because everything messes when the test writes stuff into stdout. Currently, with buffer output hijacking, it's not the case. I think I can stick to the original `running <test> .....` model. I'll make changes and send an update.

alexveden:
New look and feel:
image


image

alexveden:
> Can we make the [FAIL] move to the to the same place as [PASS]?

It's conflicting with stdout/stderr hijacking, and lead to this.

image

alexveden:
I've figured out
image

pushing the updates

lerno:
Is this ready for me to have a look at again now?

alexveden:
Yes, everything is ready. Please, go ahead.

alexveden:
look good now, ready to review

lib/std/core/runtime_test.c3
```diff
@@ -44,82 +41,170 @@ fn int cmp_test_unit(TestUnit a, TestUnit b)
 	return (int)(an - bn);
 }
 
-TestContext* test_context @private;
+fn bool terminal_has_ansi_codes() @local {
+	// TODO: implement terminal ansi code support check, currently colors can be enabled manually
```
lerno, line 45: Please implement this, it's bad to have a TODO in the code here.

lib/std/io/io.c3
```diff
@@ -227,7 +227,14 @@ fn void! out_putstream_fn(void* data, char c) @private
 
 fn void! out_putchar_fn(void* data @unused, char c) @private
 {
-	libc::putchar(c);
+	$if env::TESTING:
```
lerno, line 230: Instead of this, change stdout and stderr to a suitable file.

lib/std/os/argparse.c3
```diff
@@ -0,0 +1,536 @@
+<*
+Console arguments parsing module
```
lerno, line 2: Make this one a separate PR

lib/std/io/io.c3
alexveden: To do this, `io.c3` has to access internals of `std::core::test`, so we could make a cross dependency here. In `env::TESTING` we have `stdout` file replaced by hijacked temporary file under the hood. This is a nature of the hack in essence.

lib/std/core/runtime_test.c3
alexveden: ok, but I don't have access to Windows, it has specific implementation. Any thought how to do this right?

lib/std/os/argparse.c3
alexveden: `test.c3` depends on `argparce.c3`, are you going to merge it into master ASAP to let this one work?

lerno:
I've now merged this (without argparse) from a different branch I made (#1908). The argument parser is on a different pull request and we can discuss any potential changes there. Please look at the pull request for more information about changes made, especially regarding argument parsing.

So I'll close this one and we'll head over to argparse to discuss the API a little?

PR #1814, opened 2025-01-11
use WORKING_DIRECTORY to ensure stability of finding the git hash

jeffamstutz:
Without passing `WORKING_DIRECTORY` to `execute_process()`, the build directory _must_ be in the source tree for the `git` command to succeed.


lerno:
Thank you!

PR #1838, opened 2025-01-14
fix the link order to support LLVM_20 change

Colibrow:
@lerno Does the current PR look good to you?


lerno:
Yes, this correctly fixes things! Thank you.

PR #1842, opened 2025-01-14
math_tests: pow test

MendocinoWhiteDeer:
This PR was made after looking at the recent PR #1781. It seeks to add testing or improve testing for the relevant math functions. Of course, even though the PR mentioned involves nolibc implementations of these functions, these math tests should work in both a libc or nolibc environment.

1. Added a test for `pow()`.
2. Added more test points to `exp()` and `log()` to look at various possible inputs. Both lacked floating-point inputs that are less than 1.  The `log()` test lacked inputs that produced negative outputs. 
3. Replaced nolibc code with word macros. For example, `atan.c3` has an unnecessary macro defined called `extract_words()` when it can use already defined macros.


Adversing:
Thanks!

lerno:
Thank you!

PR #1843, opened 2025-01-14
[FEAT] add golang like channel

Snikimonkd:
this is a go like channel struct

### DIFF
--- a/lib/std/threads/buffered_channel.c3
+++ b/lib/std/threads/buffered_channel.c3
@@ -1,0 +1,162 @@
+module std::thread::channel(<Type>);
+
+distinct BufferedChannel = void*;
+
+struct BufferedChannelImpl @private
+{
+	Allocator allocator;
+	Mutex mu;
+	bool closed;
+	usz size;
+	usz elems;
+
+	usz sendx;
+	usz send_waiting;
+	ConditionVariable send_cond;
+
+	usz readx;
+	usz read_waiting;
+	ConditionVariable read_cond;
+
+	Type[*] buf;
+}
+
+fn void! BufferedChannel.new_init(&self, usz size = 1, Allocator allocator = allocator::heap())
+{
+	BufferedChannelImpl* channel = allocator::new_with_padding(allocator, BufferedChannelImpl, Type.sizeof * size)!;
+
+	channel.allocator = allocator;
+	channel.size = size;
+	channel.elems = 0;
+	channel.sendx = 0;
+	channel.send_waiting = 0;
+	channel.readx = 0;
+	channel.read_waiting = 0;
+
+	channel.mu.init()!;
+	defer catch (void)channel.mu.destroy();
+	channel.send_cond.init()!;
+	defer catch (void)channel.send_cond.destroy();
+	channel.read_cond.init()!;
+	defer catch (void)channel.read_cond.destroy();
+
+	*self = (BufferedChannel)channel;
+}
+
+fn void! BufferedChannel.destroy(&self)
+{
+	BufferedChannelImpl* channel = (BufferedChannelImpl*)(*self);
+
+	anyfault err = @catch(channel.mu.destroy());
+	err = @catch(channel.send_cond.destroy()) ?: err;
+	err = @catch(channel.read_cond.destroy()) ?: err;
+	allocator::free(channel.allocator, channel);
+
+	*self = null;
+
+	if (err) return err?;
+}
+
+fn void! BufferedChannel.push(self, Type val)
+{
+	BufferedChannelImpl* channel = (BufferedChannelImpl*)self;
+
+	channel.mu.lock()!;
+	defer catch (void)channel.mu.unlock();
+
+	// if channel is full -> wait
+	while (channel.elems == channel.size && !channel.closed)
+	{
+		channel.send_waiting++;
+		channel.send_cond.wait(&channel.mu)!;
+		channel.send_waiting--;
+	}
+
+	// check if channel is closed
+	if (channel.closed)
+	{
+		return ThreadFault.CHANNEL_CLOSED?;
+	}
+
+	// save value to buf
+	channel.buf[channel.sendx] = val;
+
+	// move pointer
+	channel.sendx++;
+	if (channel.sendx == channel.size)
+	{
+		channel.sendx = 0;
+	}
+
+	// channelge elems counter
+	channel.elems++;
+
+	// if someone is waiting -> awake him
+	if (channel.read_waiting > 0)
+	{
+		channel.read_cond.signal()!;
+	}
+
+	channel.mu.unlock()!;
+}
+
+fn Type! BufferedChannel.pop(self)
+{
+	BufferedChannelImpl* channel = (BufferedChannelImpl*)self;
+
+	channel.mu.lock()!;
+	defer catch (void)channel.mu.unlock();
+
+	// if chan is empty -> wait for sender
+	while (channel.elems == 0 && !channel.closed)
+	{
+		channel.read_waiting++;
+		channel.read_cond.wait(&channel.mu)!;
+		channel.read_waiting--;
+	}
+
+	// check if chan is closed and empty
+	if (channel.closed && channel.elems == 0)
+	{
+		return ThreadFault.CHANNEL_CLOSED?;
+	}
+
+	// read from buf
+	Type ret = channel.buf[channel.readx];
+
+	// move pointer
+	channel.readx++;
+	if (channel.readx == channel.size)
+	{
+		channel.readx = 0;
+	}
+
+	// change elems counter
+	channel.elems--;
+
+	// if someone is waiting -> awake him
+	if (channel.send_waiting > 0)
+	{
+		channel.send_cond.signal()!;
+	}
+
+	channel.mu.unlock()!;
+
+	return ret;
+}
+
+fn void! BufferedChannel.close(self)
+{
+	BufferedChannelImpl* channel = (BufferedChannelImpl*)self;
+
+	anyfault err = @catch(channel.mu.lock());
+
+	channel.closed = true;
+
+	err = @catch(channel.read_cond.broadcast()) ?: err;
+	err = @catch(channel.send_cond.broadcast()) ?: err;
+	err = @catch(channel.mu.unlock()) ?: err;
+
+	if (err) return err?;
+}
+
--- a/lib/std/threads/buffered_channel.c3
+++ b/lib/std/threads/buffered_channel.c3
@@ -23,7 +23,13 @@ struct BufferedChannelImpl @private
 
 fn void! BufferedChannel.new_init(&self, usz size = 1, Allocator allocator = allocator::heap())
 {
+	return self.init(size, allocator::heap());
+}
+
+fn void! BufferedChannel.init(&self, usz size = 1, Allocator allocator)
+{
 	BufferedChannelImpl* channel = allocator::new_with_padding(allocator, BufferedChannelImpl, Type.sizeof * size)!;
+	defer catch allocator::free(allocator, channel);
 
 	channel.allocator = allocator;
 	channel.size = size;
@@ -73,10 +79,7 @@ fn void! BufferedChannel.push(self, Type val)
 	}
 
 	// check if channel is closed
-	if (channel.closed)
-	{
-		return ThreadFault.CHANNEL_CLOSED?;
-	}
+	if (channel.closed) return ThreadFault.CHANNEL_CLOSED?;
 
 	// save value to buf
 	channel.buf[channel.sendx] = val;
--- a/test/unit/stdlib/threads/channel.c3
+++ b/test/unit/stdlib/threads/channel.c3
@@ -1,0 +1,337 @@
+module thread_test;
+
+import std::thread::channel;
+import std::thread;
+import std::time;
+import std::io;
+
+fn void init_destroy_buffered() @test
+{
+	for (usz i = 0; i < 20; i++)
+	{
+		BufferedChannel(<int>) c;
+		c.new_init(1)!!;
+		defer c.destroy()!!;
+	}
+}
+
+fn void init_destroy_unbuffered() @test
+{
+	for (usz i = 0; i < 20; i++)
+	{
+		UnbufferedChannel(<int>) c;
+		c.new_init()!!;
+		defer c.destroy()!!;
+	}
+}
+
+fn void push_to_buffered_channel_no_lock() @test
+{
+	BufferedChannel(<int>) c;
+	c.new_init(1)!!;
+	defer c.destroy()!!;
+
+	c.push(1)!!;
+}
+
+fn void push_pop_buffered_no_locks() @test
+{
+	BufferedChannel(<int>) c;
+	c.new_init(1)!!;
+	defer c.destroy()!!;
+
+	c.push(123)!!;
+	int got = c.pop()!!;
+	assert(got == 123);
+}
+
+fn void push_pop_unbuffered_with_locks() @test
+{
+	UnbufferedChannel(<int>) c;
+	c.new_init()!!;
+	defer c.destroy()!!;
+
+	Thread thread;
+	defer thread.join()!!;
+
+	thread.create(fn int(void* arg)
+	{
+		UnbufferedChannel(<int>) c = (UnbufferedChannel(<int>))arg;
+		c.push(123)!!;
+		c.push(321)!!;
+		return 0;
+	}, (void*)c)!!;
+
+	int got = c.pop()!!;
+	assert(got == 123);
+	got = c.pop()!!;
+	assert(got == 321);
+}
+
+fn void sending_to_closed_unbuffered_chan_is_forbidden() @test
+{
+	UnbufferedChannel(<int>) c;
+	c.new_init()!!;
+	defer c.destroy()!!;
+
+	c.close()!!;
+
+	if (catch err = c.push(123))
+	{
+		assert(err == ThreadFault.CHANNEL_CLOSED);
+		return;
+	}
+	assert(false);
+}
+
+fn void sending_to_closed_buffered_chan_is_forbidden() @test
+{
+	BufferedChannel(<int>) c;
+	c.new_init(1)!!;
+	defer c.destroy()!!;
+
+	c.close()!!;
+
+	if (catch err = c.push(123))
+	{
+		assert(err == ThreadFault.CHANNEL_CLOSED);
+		return;
+	}
+	assert(false);
+}
+
+fn void reading_from_empty_closed_unbuffered_chan_is_forbidden() @test
+{
+	UnbufferedChannel(<int>) c;
+	c.new_init()!!;
+	defer c.destroy()!!;
+
+	c.close()!!;
+
+	if (catch err = c.pop())
+	{
+		assert(err == ThreadFault.CHANNEL_CLOSED);
+		return;
+	}
+	assert(false);
+}
+
+fn void reading_from_empty_closed_buffered_chan_is_forbidden() @test
+{
+	BufferedChannel(<int>) c;
+	c.new_init(1)!!;
+	defer c.destroy()!!;
+
+	c.close()!!;
+
+	if (catch err = c.pop())
+	{
+		assert(err == ThreadFault.CHANNEL_CLOSED);
+		return;
+	}
+	assert(false);
+}
+
+fn void reading_from_non_empty_closed_buffered_chan_is_ok() @test
+{
+	BufferedChannel(<int>) c;
+	c.new_init(3)!!;
+	defer c.destroy()!!;
+
+	c.push(1)!!;
+	c.push(2)!!;
+	c.push(3)!!;
+
+	c.close()!!;
+
+	int got = c.pop()!!;
+	assert(got == 1);
+	got = c.pop()!!;
+	assert(got == 2);
+	got = c.pop()!!;
+	assert(got == 3);
+
+	int! got_err = c.pop();
+	if (catch err = got_err)
+	{
+		assert(err == ThreadFault.CHANNEL_CLOSED);
+		return;
+	}
+
+	assert(false);
+}
+
+fn void reading_from_empty_buffered_chan_aborted_by_close() @test
+{
+	BufferedChannel(<int>) c;
+	c.new_init(3)!!;
+	defer c.destroy()!!;
+
+	Thread thread;
+	defer thread.join()!!;
+
+	thread.create(fn int(void* arg)
+	{
+		BufferedChannel(<int>) c = (BufferedChannel(<int>))arg;
+		c.close()!!;
+		return 0;
+	}, (void*)c)!!;
+
+	int! res = c.pop();
+	if (catch err = res)
+	{
+		assert(err == ThreadFault.CHANNEL_CLOSED);
+		return;
+	}
+
+	assert(false);
+}
+
+fn void reading_from_unbuffered_chan_aborted_by_close() @test
+{
+	UnbufferedChannel(<int>) c;
+	c.new_init()!!;
+	defer c.destroy()!!;
+
+	Thread thread;
+	defer thread.join()!!;
+
+	thread.create(fn int(void* arg)
+	{
+		UnbufferedChannel(<int>) c = (UnbufferedChannel(<int>))arg;
+		c.close()!!;
+		return 0;
+	}, (void*)c)!!;
+
+	int! res = c.pop();
+	if (catch err = res)
+	{
+		assert(err == ThreadFault.CHANNEL_CLOSED);
+		return;
+	}
+
+	assert(false);
+}
+
+fn void sending_to_full_buffered_chan_aborted_by_close() @test
+{
+	BufferedChannel(<int>) c;
+	c.new_init(1)!!;
+	defer c.destroy()!!;
+
+	c.push(1)!!;
+
+	Thread thread;
+	defer thread.join()!!;
+
+	thread.create(fn int(void* arg)
+	{
+		BufferedChannel(<int>) c = (BufferedChannel(<int>))arg;
+		c.close()!!;
+		return 0;
+	}, (void*)c)!!;
+
+	anyfault err = @catch(c.push(1));
+	if (err)
+	{
+		assert(err == ThreadFault.CHANNEL_CLOSED);
+		return;
+	}
+
+	assert(false);
+}
+
+fn void sending_to_unbuffered_chan_aborted_by_close() @test
+{
+	UnbufferedChannel(<int>) c;
+	c.new_init()!!;
+	defer c.destroy()!!;
+
+	Thread thread;
+	defer thread.join()!!;
+
+	thread.create(fn int(void* arg)
+	{
+		UnbufferedChannel(<int>) c = (UnbufferedChannel(<int>))arg;
+		c.close()!!;
+		return 0;
+	}, (void*)c)!!;
+
+	anyfault err = @catch(c.push(1));
+	if (err)
+	{
+		assert(err == ThreadFault.CHANNEL_CLOSED);
+		return;
+	}
+
+	assert(false);
+}
+
+fn void multiple_actions_unbuffered() @test
+{
+	UnbufferedChannel(<int>) c;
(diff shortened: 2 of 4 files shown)


Adversing:
I took a look at your code and it seems like both the buffered and unbuffered channels set `self.closed = true;` in `close()` but never call `send_cond.signal()`, `read_cond.signal()`, or a broadcast to wake up any threads that might be stuck in wait() calls. 

Example in `BufferedChan.close()`:
```c3
fn void BufferedChan.close(&self) @dynamic {
    self.mu.lock()!!;
    defer self.mu.unlock()!!;

    self.closed = true;
    // self.send_cond.broadcast()!!;
    // self.read_cond.broadcast()!!;
}
```

Example in `UnbufferedChan.close()`:
```c3
fn void UnbufferedChan.close(&self) @dynamic {
    self.mu.lock()!!;
    defer self.mu.unlock()!!;

    self.closed = true;
    // self.send_cond.broadcast()!!;
    // self.read_cond.broadcast()!!;
}
```

So, since neither version signals/broadcasts on close, if a thread is blocked in `push()` (or `pop()`) waiting on `self.send_cond.wait(...)`/`self.read_cond.wait(...)`, there is nothing to wake it up once `closed = true;` is set.

lib/std/threads/chan.c3
```diff
@@ -0,0 +1,257 @@
...
+	self.mu.destroy()!!;
+	self.send_mu.destroy()!!;
+    self.send_cond.destroy()!!;
+	self.read_mu.destroy()!!;
+    self.read_cond.destroy()!!;
+    allocator::free(self.allocator, self);
+}
+
+fn void! UnbufferedChan.push(&self, Type val) @dynamic {
+    self.mu.lock()!!;
```
radekm, line 188: Why an error can happen here?

Snikimonkd: i don't know, i just follow the instructions of the compiler ))0)

Snikimonkd: lib/std/threads/thread.c3:42

```
macro void! Mutex.lock(&mutex) => NativeMutex.lock((NativeMutex*)mutex);
```

Snikimonkd: it seems like this is very os in depth errors, which probably noone wants to handle

Snikimonkd: on the other hand in `lib/std/threads/fixed_pool.c3` i can see that some errors are rethrown and some errors are paniced

lib/std/threads/buffered_chan.c3
```diff
@@ -0,0 +1,136 @@
...
+
+fn Chan(<Type>)! BufferedChan.init(&self) @dynamic {
+	self.mu.init()!;
+	defer catch (void)self.mu.destroy();
+	self.send_cond.init()!;
+	defer catch (void)self.send_cond.destroy();
+	self.read_cond.init()!;
+	defer catch (void)self.read_cond.destroy();
+
+	self.size = SIZE;
```
Book-reader, line 29: if the size is constant, you don't need `this.size` and can just use `SIZE`.

lib/std/threads/buffered_chan.c3
```diff
@@ -0,0 +1,136 @@
...
+struct BufferedChan (Chan(<Type>)) {
+	Mutex mu;
+	Type[SIZE] buf;
+	usz size;
+	bool closed;
+	usz elems;
+
+	usz sendx;
+	usz send_waiting;
+	thread::ConditionVariable send_cond;
```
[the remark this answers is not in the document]
Snikimonkd: fixed

lib/std/threads/buffered_chan.c3
Snikimonkd: fixed

lib/std/threads/buffered_chan.c3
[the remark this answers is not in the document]
Book-reader: Great! When you have resolved a review, there should be an option to 'mark as resolved', which collapses the section of the review to make it easier to tell what has been resolved :)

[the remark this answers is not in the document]
Snikimonkd: Ok, thx. At the place where i work we have an agreement that only the the author of the review/issue can resovle it, so we can validate that the issue was not closed without fix

lib/std/threads/buffered_chan.c3
```diff
@@ -0,0 +1,134 @@
+module std::thread::buffered_chan(<Type, SIZE>);
+
+import std::thread::chan;
+
+struct BufferedChan (Chan(<Type>)) {
+	Mutex mu;
+	Type[SIZE] buf;
```
Book-reader, line 7: this is another brain dump, so be warned.

Have you considered dynamic buffers for `BufferedChan`?
I am aware that `List(<>).remove_first()` like would be needed for this is a lot slower than the current method with the static array, but there is the `LinkedList(<>)` that is not quite as fast as the current method of using an array as a queue but is a lot faster than `List(<>)`  and allows a buffer of effectively unlimited size.

I understand that there may be a usecase for limiting the number of values in the channel so maybe giving a size of `0` or `-i` could use some `$if` and `@if` to enable a dynamic buffer or always use a dynamic buffer but check if there are too many messages and handle it somehow.

If implemented with both dynamic and static buffers, It might be difficult for a user to realize that setting the capacity to `0`/`-1` actually makes it dynamic, so maybe this could be a usecase for default generic arguments or maybe using some variation https://github.com/c3lang/c3c/issues/1856#issuecomment-2597173213 to have a `Chan ch = unbuffered_chan::new_unlimited(<int>)() or unbuffered_chan::new_unlimited(int)` to not have to pass the second argument and automatically pass the magic value that makes it unlimited, but this is reaching c3 feature request levels.

there could also be two seperate buffered channel structs one with a static buffer and one with a dynamic buffer.

having just a dynamic buffer could slightly things as instead of doing the array stuff you could just `buf.append()` and `buf.pop_front()`

I understand that this is a lot of stuff, please feel free to ignore it.

If you have any questions about my ramblings, feel free to ask :)

lib/std/threads/buffered_chan.c3
```diff
@@ -0,0 +1,134 @@
...
+	anyfault err = @catch(self.mu.destroy());
+	err = @catch(self.send_cond.destroy()) ?: err;
+	err = @catch(self.read_cond.destroy()) ?: err;
+
+	if (err) return err?;
+}
+
+fn void! BufferedChan.push(&self, Type val) @dynamic {
+	self.mu.lock()!;
+	defer catch (void)self.mu.unlock();
```
Book-reader, line 46: if you cast this to `void` you don't need to `catch` it, and if you `catch` it you don't need to cast it to `void` (just `defer catch self.mu.unlock()!!` instead, even though it is `!!` it won't actually crash because of the `catch`, I have just tested this).

this is the same for all the other times you do `defer catch (void)...`

there should maybe be an error for `defer catch`ing a value that is not optional? I am not sure about this

lib/std/threads/buffered_chan.c3
```diff
@@ -0,0 +1,134 @@
...
+
+	// change elems counter
+	self.elems++;
+
+	// if someone is waiting -> awake him
+	if (self.read_waiting > 0) {
+		self.read_cond.signal()!;
+	}
+
+	self.mu.unlock()!;
```
Book-reader, line 78: you `defer catch (void)self.mu.unlock();` at the start of the function, one of them is not needed (the second one will throw `UNLOCK_FAILED` but because the error is caught the program won't crash) so probably remove this one.

this is the same for `BufferedChan.pop()` and `UnbufferedChan`

lib/std/threads/buffered_chan.c3
```diff
@@ -0,0 +1,134 @@
...
+	// if chan is full -> wait
+	while (self.elems == SIZE && !self.closed) {
+		self.send_waiting++;
+		self.send_cond.wait(&self.mu)!;
+		self.send_waiting--;
+	}
+
+	// check if chan is closed
+	if (self.closed) {
+		self.mu.unlock()!;
```
Book-reader, line 57: you don't need to do this as the above `defer unlock()` will do it
same for `pop()` and all `UnbufferedChan` functions

lib/std/threads/buffered_chan.c3
Snikimonkd: it seems that if you want to use unlimited size - it is easier to just use LinkidList with mutex

Snikimonkd: i think that one of the ideas behind channels is blocking on emptiness/fullness

lib/std/threads/buffered_chan.c3
Snikimonkd: `defer catch` triggers only if the fucntion returns optional

here i need to unlock mutex at the end of function - it's meant for the happy path without errors

if case any of cond.wait/cond.signal throws an error defer catch will try to release the lock

Book-reader: I see, that makes sense then.

lib/std/threads/buffered_chan.c3
Book-reader: > there should maybe be a compiler error for `defer catch`ing a value that is not optional? I am not sure if this is a bug.

I just didn't understand `defer catch`, this can be dismissed

lib/std/threads/buffered_chan.c3
Book-reader: That makes sense, this can be marked as resolved then

lib/std/threads/buffered_chan.c3
Snikimonkd: fixed

lib/std/threads/buffered_chan.c3
```diff
@@ -0,0 +1,136 @@
...
+	usz sendx;
+	usz send_waiting;
+	thread::ConditionVariable send_cond;
+
+	usz readx;
+	usz read_waiting;
+	thread::ConditionVariable read_cond;
+}
+
+fn Chan(<Type>)! BufferedChan.init(&self) @dynamic {
```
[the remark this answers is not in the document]
Snikimonkd: The reasons of the constructor being implemented like this are the following:
1) channel must be passed to the thread by reference - that's why i hide the channel implementation behind the Interface
2) (imho) - if we can create something without allocations - its better to do it like this
but i don't mind to redo it)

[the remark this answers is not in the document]
Snikimonkd: BufferedChan signature `(<Type, SIZE>)` is fully stolen from ring buffer (in reality i just reimplemented ring buffer with some quirks specific to channel)

lib/std/threads/unbuffered_chan.c3
```diff
@@ -0,0 +1,121 @@
...
+	err = @catch(self.send_cond.destroy()) ?: err;
+	err = @catch(self.read_mu.destroy()) ?: err;
+	err = @catch(self.read_cond.destroy()) ?: err;
+
+	if (err) return err?;
+}
+
+fn void! UnbufferedChan.push(&self, Type val) @dynamic {
+	self.mu.lock()!;
+	defer catch (void)self.mu.unlock();
```
radekm, line 49: Is the `catch` necessary when the excuse is ignored?

radekm: Oh I see, it's only when existing with error. But doesn't the error mean that the lock wasn't locked? So it's not necessary to unlock it?

lib/std/threads/unbuffered_chan.c3
```diff
@@ -0,0 +1,121 @@
...
+fn Type! UnbufferedChan.pop(&self) @dynamic {
+	self.mu.lock()!;
+	defer catch (void)self.mu.unlock();
+	self.read_mu.lock()!;
+	defer catch (void)self.read_mu.unlock();
+
+	// if no one is waiting, then there is nothing in the buffer
+	while (!self.closed && self.send_waiting == 0) {
+		self.read_waiting++;
+		self.read_cond.wait(&self.mu)!;
```
radekm, line 89: When waiting both `self.mu` and `self.read_mu` are locked? Is that correct?

lib/std/threads/buffered_chan.c3
```diff
@@ -0,0 +1,132 @@
...
+	if (self.read_waiting > 0) {
+		self.read_cond.signal()!;
+	}
+
+	self.mu.unlock()!;
+
+	return;
+}
+
+fn Type! BufferedChan.pop(&self) @dynamic {
```
radekm, line 82: Is there a fairness mechanism?

lib/std/threads/unbuffered_chan.c3
Book-reader: There are error returns in other places, like on line 54 there is `return ThreadFault.CHANNEL_CLOSED?` where this also applies

lib/std/threads/buffered_chan.c3
Snikimonkd: you mean if the thread came and locked first - then it is getting the value first? - no

lib/std/threads/unbuffered_chan.c3
Snikimonkd: `cond.wait()` releases the mutex, so `self.mu` will not be locked

Snikimonkd: and when someone `signals` - it means that the thread which was waiting owns the mutex

lib/std/threads/buffered_chan.c3
radekm: Why do we want to unlock the lock only in case of error?

lib/std/threads/buffered_chan.c3
Snikimonkd: to make it work we need some sort of mechanism to `awake/send to sleep` specific thread

lib/std/threads/unbuffered_chan.c3
radekm: Sorry. But still why do we unlock only in case when failure happens? Why not unlock always?

lib/std/threads/buffered_chan.c3
Snikimonkd: if we have this mechanism, then instead of `send_waiting/read_waiting` we can use `linkedlist` wich stores threads

lib/std/threads/unbuffered_chan.c3
Snikimonkd: so on all successful exits from function i have `self.send_mu.unlock()!;`
this defer is meant when `self.send_cond.wait(&self.mu)!;` or `self.read_cond.signal()!;` or any other function throws an error and exits imideatly

Snikimonkd: for example if `self.read_cond.signal()!;` on line 64 throws an error - this defer will try to release the lock

Snikimonkd: and if everything goes good i will unlock the at the end of function on lines 74 and 75

lib/std/threads/buffered_chan.c3
radekm: Something like that. Or that each thread will eventually pop a value from the channel, provided there are sufficient values available. It is guaranteed that no thread will be starved (i.e., some threads consuming all the values while others never get a chance to pop any value).

lib/std/threads/unbuffered_chan.c3
radekm: Thanks!

lib/std/threads/unbuffered_chan.c3
radekm: And why not remove `catch` from `defer` and always unlock in `defer` and also remove lines 74 and 75? Would it work?

radekm: And another question, isn't the code unlocking twice when `self.closed` is `true` on line 70?

lib/std/threads/buffered_chan.c3
Snikimonkd: actually i have an idea how to implement it - we can create a cond for every thread and keep this conds in LiknedList - this will guarantee us the order of awaking of threads is the same as the order of them going to wait

lib/std/threads/unbuffered_chan.c3
radekm: What if there's a failure in `self.send_cond.wait(&self.mu)!` after unlocking `self.mu`. Wouldn't the code then unlock twice?

lib/std/threads/buffered_chan.c3
Snikimonkd: but i think it's gonna be very inefficient

lib/std/threads/unbuffered_chan.c3
Snikimonkd: so to be honest i don't know how `wait` fails. Will it unlock the mutex if it fails - i can't find anything about it in docs (i used pthread docs, cause c3 threads for linux/macos are baased on pthread)? In any case we try to fix as mach as we can on exit - if mutex is already unlocked -ok, we will get an error and hide it with `(void)` cast, if it is not - then we will unlock it

lib/std/threads/unbuffered_chan.c3
Snikimonkd: self.read_mu is locked, while self.mu will not be locked

lib/std/threads/unbuffered_chan.c3
Snikimonkd: i try to clean as mach as i can when exiting - thats why i try to unlock everything

Book-reader (commented):
I don't have enough time to do a full review rn so I have just reviewed the first things I noticed

Book-reader (commented):
Did you think about implementing the idea that @lerno said on stream with
```
struct BufferedChannel
{
  ...
  Type[*] buf
}

fn BufferedChannel* new_buffered(usz buf_size, Allocator allocator = heap allocator)
{
  BufChan new = allocator::new_w_padding(allocator, BufChan, Type.sizeof*buf_size)
 ...
 return new;
}
```
This would allow removing the second parameter from the generic module and allow buffers sized at runtime, as well as removing some possible confusion because the function would specifically say if it is unbuffered or not.

A little module frickery could be done to allow this:
`BufChan(<int>) a = channel::new_buffered(int, 123)` instead of `BufChan(<int>) a = channel::new_buffered(<int>)(123)`

I also think that the buffered and unbuffered should be seperate to make it more obvious than just the size being 0 (especially with the above where size would not be part of the module):
```
distinct BufChan = void*;
struct BufChanImpl @private
{
  ...
  Type[*] buf;
}

fn void! BufChan.push(self, ...)
{
  BufChanImpl* chan = (BufChanImpl)self;
  chan.buf[idx] = whatever;
}
```
Written on my phone so may be formatting issues and shortend words.

Snikimonkd:
> ability to pass both a buffered and unbuffered channel to the same function as the same parameter

Thats actually a very good point which i agree with - it should be impossible to pass any (buffered/unbuffered) channel as a parameter. 

P.S. I came from go where the difference between buffered and unbuffered channel is only how it is created. This is a correct code which compiles and works fine:
```
package main

func testChan(ch chan int) {
	_ = ch
}

func main() {
	c1 := make(chan int)    // unubuffered
	c2 := make(chan int, 1) // buffered

	testChan(c1)
	testChan(c2)
}
```
I don't say it's good, it's just something i'am used to))

lib/std/threads/buffered_channel.c3
```diff
@@ -0,0 +1,184 @@
...
+	BufferedChannelImpl* channel = allocator::new_with_padding(allocator, BufferedChannelImpl, Type.sizeof * size)!;
+	channel.init(size, allocator)!;
+	*self = (BufferedChannel)channel;
+	return;
+}
+
+fn void! BufferedChannel.destroy(&self)
+{
+	BufferedChannelImpl* channel = (BufferedChannelImpl*)(*self);
+	channel.destroy()!;
```
Book-reader, line 16: since you only call the `ChannelImpl` methods from inside the `Channel` type, it may be good to remove the methods from `ChannelImpl` and inline them in `Channel` to be consistent with `Map`

lib/std/threads/buffered_channel.c3
```diff
@@ -0,0 +1,184 @@
...
+	// if chan is empty -> wait for sender
+	while (self.elems == 0 && !self.closed)
+	{
+		self.read_waiting++;
+		self.read_cond.wait(&self.mu)!;
+		self.read_waiting--;
+	}
+
+	// check if chan is closed and empty
+	if (self.closed && self.elems == 0)
```
Book-reader, line 142: so the `pop` method only returns `CHANNEL_CLOSED` if both the channel is closed and the buffer is empty, but above you wait for the sender to send something so the buffer will never be empty here and I don't see how this code will ever execute

lib/std/threads/buffered_channel.c3
```diff
@@ -0,0 +1,184 @@
...
+	self.allocator = allocator;
+	self.size = size;
+	self.elems = 0;
+	self.sendx = 0;
+	self.send_waiting = 0;
+	self.readx = 0;
+	self.read_waiting = 0;
+
+	self.mu.init()!;
+	defer catch (void)self.mu.destroy();
```
Book-reader, line 68: all of the `defer catch` lines need to be above the line they are trying to catch otherwise they will never execute: https://c3-lang.org/language-common/defer/#pitfalls-with-defer-and-defer-catch

Book-reader (commented):
this is looking really good, so a lot of what I have now are just nitpicks

lib/std/threads/buffered_channel.c3
```diff
@@ -0,0 +1,184 @@
...
+	return channel.push(val);
+}
+
+fn Type! BufferedChannel.pop(&self)
+{
+	BufferedChannelImpl* channel = (BufferedChannelImpl*)(*self);
+	return channel.pop();
+}
+
+fn void! BufferedChannel.close(&self)
```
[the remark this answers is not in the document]
Snikimonkd: usually you have a producer and consumer, producers pushes some values to channel while consumer can be chilling or reading slowly - then producer closes the channel when he has finished his job, but there is still values inside, consumer slowly reads them and when the buffer is empty consumer gets CHAN_CLOSED error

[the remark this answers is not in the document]
Book-reader: I see, that makes sense. this can be resolved then

lib/std/threads/buffered_channel.c3
Snikimonkd: for example the channel is empty and consumer blocks on line 137

then channel is closed `cond.broadcast()` will awake the consumer, `self.closed == ture` -> `while loop` breaks

you got to this point)

lib/std/threads/buffered_channel.c3
Snikimonkd: so i understand how `inline` works with structs

but how it works with functions?
whats the difference between this two:
```
fn usz Map.len(map) @inline
{
  return map ? ((MapImpl*)map).count : 0;
}
fn Value! Map.get(map, Key key) @operator([])
{
  return *map.get_ref(key) @inline;
}
```

lib/std/threads/buffered_channel.c3
Snikimonkd: it seems like there is no need to `destroy` a mutext if it was not `inited`

lib/std/threads/buffered_channel.c3
```diff
@@ -0,0 +1,184 @@
...
+
+fn void! BufferedChannel.new_init(&self, usz size = 1, Allocator allocator = allocator::heap())
+{
+	BufferedChannelImpl* channel = allocator::new_with_padding(allocator, BufferedChannelImpl, Type.sizeof * size)!;
+	channel.init(size, allocator)!;
+	*self = (BufferedChannel)channel;
+	return;
+}
+
+fn void! BufferedChannel.destroy(&self)
```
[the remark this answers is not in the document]
Snikimonkd: `new_init` also needs to take pointer

lib/std/threads/buffered_channel.c3
Snikimonkd: so it seems like the first one is inline function and the second one is inline call

Snikimonkd: but i still don't know why i should use one over another

Book-reader: when I said inline I meant something like this:
```
distinct Foo = void*;
struct FooImpl
{
  int x;
}

fn void Foo.increment(self)
{
  FooImpl* foo = (FooImpl*)self;
  foo.x++;
}

```

instead of 
```
distinct Foo = void*;
struct FooImpl
{
  int x;
}

fn void Foo.increment(self)
{
  FooImpl* foo = (FooImpl*)self;
  foo.increment();
}

fn void FooImpl.increment(&self)
{
  self.x++;
}
```
I should have clarified

Book-reader: this is how it is done in the `Map`:
```
fn Value*! Map.get_ref(self, Key key)
{
	MapImpl *map = (MapImpl*)self;
	if (!map || !map.count) return SearchResult.MISSING?;
	uint hash = rehash(key.hash());
	for (Entry *e = map.table[index_for(hash, map.table.len)]; e != null; e = e.next)
	{
		if (e.hash == hash && equals(key, e.key)) return &e.value;
	}
	return SearchResult.MISSING?;
}
```

Snikimonkd: will this produce the same compiled result as your first example?
```
fn void Foo.increment(self)
{
  FooImpl* foo = (FooImpl*)self;
  foo.increment() @inline;
}
```

Book-reader: it might, but my review is less about the compiled result and more the actual code. There are two function definitions with the same parameters, the first function calls the second function without modifying the parameters at all, and the first function it is the only place the second function is ever called. it just seems like an unnecessary extra function.

Snikimonkd: makes sense

lib/std/threads/buffered_channel.c3
Book-reader: yep, that makes sense
It is always these sorts of logic errors that get me, sorry for bothering you with so many of these misguided reviews

Snikimonkd: it's ok

Snikimonkd: also - i tried it

somewhere deep inside mutext implementation it panics if you try to `deinit` not inited mutex :)

lib/std/threads/buffered_channel.c3
```diff
@@ -0,0 +1,166 @@
...
+	channel.mu.init()!;
+	defer catch (void)channel.mu.destroy();
+	channel.send_cond.init()!;
+	defer catch (void)channel.send_cond.destroy();
+	channel.read_cond.init()!;
+	defer catch (void)channel.read_cond.destroy();
+
+	*self = (BufferedChannel)channel;
+
+	return;
```
Book-reader, line 45: the only nitpick I have left is that you don't need a `return`

test/unit/stdlib/threads/channel.c3
```diff
@@ -0,0 +1,340 @@
+module thread_test;
+
+import std::thread::channel;
+import std::thread;
+import std::time;
+import std::io;
+
+// just an example how you can use it :)
+def IntBufChan = BufferedChannel(<int>);
```
Book-reader, line 9: this alias is defined, but only used once. you should decide if you want to use this alias and create one for an unbuffered channel as well or remove it.

Snikimonkd: at first i thought it's a good way to show newcomers how you can use alliases with generic modules, but then i realised that the amount of people who will ever look inside std lib tests is slightly more then 0

Book-reader (approved):
This looks great! If there are any more issues then I can't see them.

Thank you for putting up with my many incorrect and misguided reviews, as you might be able to tell I don't do this often.

lerno:
Thank you!

Issue #1845, opened 2025-01-15
Using `sort::quicksort` on a fixed size array can cause stalling


gaogao-qwq:
When compile & run the code below, `sort::quicksort(arr_2)` will stall forever.
```cpp
import std::io;
import std::sort;

fn int main() {
	int[]  arr_1 = {10,9,8,1,2,3,4,5,5,6,19};
	int[*] arr_2 = {10,9,8,1,2,3,4,5,5,6,19};
	int[]  arr_3 = {10,9,8,1,2,3,4,5,5,6,19};
	int[*] arr_4 = {10,9,8,1,2,3,4,5,5,6,19};

	std::io::print("Is arr_1 sortable: ");
	std::io::printn(std::sort::@is_sortable(arr_1));
	std::io::print("Is arr_2 sortable: ");
	std::io::printn(std::sort::@is_sortable(arr_2));
	std::io::print("Is arr_3 sortable: ");
	std::io::printn(std::sort::@is_sortable(arr_3));
	std::io::print("Is arr_4 sortable: ");
	std::io::printn(std::sort::@is_sortable(arr_4));

	// won't stall
	sort::quicksort(arr_1);
	sort::insertionsort(arr_3);
	sort::insertionsort(arr_4);
	std::io::printn("Won't stall until here.");

	// will stall
	sort::quicksort(arr_2);
	std::io::printn("Would never be able to reach here.");
	return 0;
}
```

output:
```
Is arr_1 sortable: true
Is arr_2 sortable: true
Is arr_3 sortable: true
Is arr_4 sortable: true
Won't stall until here.
```

lerno:
There were some errors here:
1. Passing an array by value was permitted.
2. Passing an array by ref failed

This should now be fixed.

gaogao-qwq:
Tested on my machine and can conform that this issue is fixed.

PR #1846, opened 2025-01-15
fix quicksort stalling

konimarti:
Fix quicksort stalling by passing `#list` instead of `list` to the `@partition` function. The stalling occurs when a fixed-sized array is passed to quicksort. Fixed-sized arrays parameters will be passed by value (and not by references); hence, a fixed-sized array will not be sorted but it should also not stall the algorithm.

As additional measure (or alternative), we could use a contract to avoid passing fixed-size arrays or vectors to the quicksort function, i.e.
```
@require values::@typekind(list) != ARRAY
@require values::@typekind(list) != VECTOR
```

Fixes: #1845

### DIFF
--- a/lib/std/sort/quicksort.c3
+++ b/lib/std/sort/quicksort.c3
@@ -106,40 +106,40 @@ fn ElementType! qselect(Type list, isz low, isz high, isz k, CmpFn cmp, Context 
 	return SearchResult.MISSING?;
 }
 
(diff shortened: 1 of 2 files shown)


konimarti:
Thanks! I felt obliged to propose a solution since it seemed like a regression I introduced with the partition function. 

BTW: @lerno did you see the failure of one of the unit tests? This looks unrelated to the PR and seems to come from the threads unit tests.

PR #1863, opened 2025-01-18
[FEAT] Add CLI flags to filter 'c3c project view' results 

BWindey:
This PR adds flags to `c3c project view` to filter the shown project properties.
This is useful for external tooling, like (bash) completion scripts, who can provide better completions without the need to parse the project properties from the full `c3c project view` output.

The original `project view` stayed the same, but now has a `--help / -h` flag to display the options. Note that I did not provide a `-hh` for long help, as I did not see the need for this subcommand.

```
$ c3c project view --help
Usage: c3c [<options>] project view [<options>]

View the content of project.json in a more readable format.
Without flags, this prints out all the project properties.

Flags can be used to filter for specific properties.
When one or more flags are used, the other properties do not get shown

The verbosity of the output can be controlled with -v and -vv
Verbosity has no effect without flags. When -vv is used with flags,
the results will be printed out like they are in the full view.
Otherwise the "<header>: " is left out.

With flags on, each selected property will be seperated by an empty
line, and properties with multiple values (like --authors) will have
their values printed each on a new line.

View options:
  -h --help                Show this help
  -v -vv                   Normal or very verbose output
  --authors                List of authors
  --version                Project version
  --language-revision      Project language revision
  --warnings-used          List of enabled compiler-warnings
  --c3l-lib-search-paths   List of C3 linker library search paths
  --c3l-lib-dependencies   List of C3 linker library dependencies
  --source-paths           List of C3 source file paths
  --output-location        Output directory
  --default-optimization   Default optimization level
  --targets                Project targets (!= compilation-targets)
  ```

Currently I did not implement this for every possible target, as there are a lot of them, and the current code could use some refactoring to more easily support this kind of actions.
I would suggest that those others can either be implemented in the same "verbose" way as these currently implemented in this PR, or to first rethink the  way project properties are stored and displayed, which could be a different PR.

This is my first real PR, so I'm not sure I used the correct code-style, and if there are things I missed.

### CODE REVIEW

**Comment by BWindey** [comment]
Maybe add a reference to #1311 , as it's on the same realm.

**Comment by BWindey** [comment]
Added one last thing: `c3c project --help / -h` will now show the `project_usage()` without the following error message:
```
Error: Cannot process the unknown subcommand "--help".
```

lerno:
Thank you!

PR #1865, opened 2025-01-19
Fix `linux-crt` and `linux-crtbegin` not getting recognized as a project parameter

cd-n0:
The project parameters `linux-crt` and `linux-crtbegin` are implemented but not recognized by the compiler, leading to erroneous warning messages. This PR adds them to the `project_default_keys` and `project_target_keys` arrays, along with their descriptions, to suppress the warnings.


lerno:
Thank you!

PR #1868, opened 2025-01-20
Fix 'strtul' to 'strtoul'

BWindey:
There was a small typo in the libc bindings. `strtul` does not exist, but `strtoul` does. That function does not have a binding yet, so I'm confident that `strtoul` was originally intended.


lerno:
Thank you!

PR #1869, opened 2025-01-20
[feat] add test tmp files to gitignore

Snikimonkd:
Every test run generates two files, which are visible for git:
- first one is the test binary itself `test/testrun`
- second one is from read/write test
This pr adds them to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,3 +81,7 @@ result
 
 # macOS
 .DS_Store
+
+# tests
+/test/tmp/*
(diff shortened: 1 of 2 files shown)


lerno:
Looks like it fails on the tmp files.

Snikimonkd:
hehe, tmp was empty folder, so git ignored it. Added `.gitkeep` this must fix it

lerno:
Yes that worked, thank you!

PR #1885, opened 2025-01-23
[FIX] Let `c3c project` subcommands use same logic as others to get `project.json`

### LINKED ISSUE #1884: `c3c project` subcommands fail to find `project.json`, unlike `build`, `dist`, `directives`, `run`, ...
The commands `c3c build/run/...` all try to locate `project.json` in parent-directories of the current working directory if `project.json` is not found in the current working directory. 
For some reason, `c3c project ...` subcommands do not look upward. This is a strange disparity, given the fact that all those commands should parse `project.json` to get their info.

Looking at the code, I found that the `build/run/...` subcommands use `init_build_target()`, which locates the top-level-directory and then calls `project_load()`, which in turns reads `project.json`, parses it into JSON and puts everything in the `Project` struct, which it returns.
The `c3c project` subcommands however rely on `read_project()`, which reads `project.json` in the current working directory and parses it into JSON, and returns that JSON. Then that JSON is directly used withing the subcommands.
An interesting note however, is that `read_project()` has a boolean flag `assume_empty_if_not_exists`. This creates a strange behaviour when `true`: if you are in `myC3project/src/` and use `c3c project add-target targetX executable`, the command will create `myC3project/src/project.json`. I don't really know why that flag is there, because why would you want to add a target when there is not project to begin with? Maybe there is a reason I don't know.

So, what can be done?
There currently are 2 methods:
- locate-top-level-dir -> load-project -> parse-JSON & load in Project struct
- parse-JSON

I would suggest a single workflow:
- `load_project()` calls
  - `load_project_JSON()` calls
    - `find_top_level_dir()`
  - `project_add_targets()`       (loads JSON in Project struct)

Then `c3c project` subcommands can call into `load_project_JSON()`, while the other commands call into `load_project()`.
This would ensure that all subcommands get the same info, and that there is a layering of abstractions.

I would volunteer to implement this in this way or in a different way you suggest. (If I do this, I would like to know what to do with `assume_empty_if_not_exists`.)

BWindey:
Implemented this as suggested in #1884 .

Closes #1884 .

### DIFF
--- a/src/build/build_internal.h
+++ b/src/build/build_internal.h
@@ -125,6 +125,7 @@ static const char *sanitize_modes[4] = {
 	[SANITIZE_THREAD] = "thread",
 };
 
+JSONObject *project_JSON_load(const char **filename_ref);
 Project *project_load(const char **filename_ref);
 BuildTarget *project_select_target(const char *filename, Project *project, const char *optional_target);
 
--- a/src/build/build_internal.h
+++ b/src/build/build_internal.h
@@ -125,7 +125,7 @@ static const char *sanitize_modes[4] = {
 	[SANITIZE_THREAD] = "thread",
 };
 
-JSONObject *project_JSON_load(const char **filename_ref);
+JSONObject *project_json_load(const char **filename_ref);
 Project *project_load(const char **filename_ref);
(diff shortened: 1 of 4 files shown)


lerno:
Thank you!

PR #1897, opened 2025-01-27
std::os::argparse module

alexveden:
std::os::argsparse module


hwchen:
I don't know if the API for std argparse has already been discussed (It's not obvious from a quick search of issues or looking at the [test runner pr](https://github.com/c3lang/c3c/pull/1808). If not, I've got opinions :smile: and code I'd be willing to donate. But if this has already been decided I don't want to derail.

alexveden:
There are a bunch of tests of argparse there in the test runner PR. So you may try to get a sense of it API. Anyway, I'm open for ideas.

lerno:
@hwchen did you have some feedback?

hwchen:
Just want to be clear that I'm not really commenting on the the current implementation. I'm more interested in whether there's a certain type of API we're looking for in an argparse module.

I come from Rust, and not C, so I'll explain in terms of those libraries.
- [Clap](https://docs.rs/clap/latest/clap/) is very full featured. Help text generation, deriving parser using struct attributes (tags), explicit subcommands, validation, built-in API for parsing common types.
- [lexopt](https://docs.rs/lexopt/latest/lexopt/) is very minimal, it only provides a stream of values/options.

The ripgrep crate [moved away](https://github.com/BurntSushi/ripgrep/pull/2626) from clap to lexopt, in part to reduce dependencies, and also because lexopt would end up providing more control over arg parsing (at the cost of having to implement more boilerplate).

I feel that the current PR API sits between the two (more towards simplicity). I think for stdlib, I'd prefer either extreme; if it's simpler, more complex parsers can be built on it, and if it has more features it can be used easily as-is for more scenarios. Odin ended up with something more [comprehensive](https://github.com/odin-lang/Odin/pull/3700) (can defined opts using a struct with tags).

Also, I believe that wherever we want to sit on the spectrum, it's good to be explicit about it.

---

As for my own biases, I've written an [arg parsing library for c3](https://github.com/hwchen/opter-c3) which follows the general structure of lexopt's API. I might prefer something like it in the std library, but I can also see the appeal of other approaches. And seeing as everybody ends up writing their own argparse, there's probably a lot of other opinions out there too :)

tomaskallup:
I have a bit of feedback on this.

I feel like the API is fine, it's exactly what it says that it is, argument parser. If something more like a full blown CLI app API would be needed (to have 0 hassle subcommands and what not) I could be in another module, which would utilize argparse under the hood.

What I currently don't see is a way to provide an array option, since from the implementation it would seem that providing a single option multiple times would result in an error of "duplicated option". The value of the option could be handled by the `callback` function from the looks of it.

The only other thing that came to mind was a bit more "hackability", for example if I wanted to somehow implement validation of a parameter, I would have to do it myself after the parsing and I would also have to write the extra help info (if it was for example an enum). But again, this could be solved by the wrapping module, which would hold the users hand a bit more. Altough my view is similar to hwchens' above, I feel like the current implementation here is good enough and if one wants to opt-out of some of the features, they still can (for example the help option is opt-in).

alexveden:
> What I currently don't see is a way to provide an array option, since from the implementation it would seem that providing a single option multiple times would result in an error of "duplicated option". The value of the option could be handled by the callback function from the looks of it.

This is a kind a thing I was thinking about. I think it's common for CLI to have accumulated values, e.g. `-vvvv` for verbosity levels. I didn't implement arrays, because I wanted to have argparse non-allocating. But I think it may be a good idea to add multiple values, at least make it possible to do it with callbacks.

So by design, the callback mechanism is the way to extend the argparse to whatever is needed. I can refine callbacks and arrays of arguments after PR approval.

> The only other thing that came to mind was a bit more "hackability", for example if I wanted to somehow implement validation of a parameter, I would have to do it myself after the parsing and I would also have to write the extra help info (if it was for example an enum).

All hackability is implemented via callbacks, or explicit param validation after parsing in the main (or other function). `argparse` module still does simple validation, so if you expect `int` type in the option value and given a string, it will raise validation error. More complex cases, should be handled by the program via callback of argparse, or after parse completes in regular code.

tomaskallup:
So for the arrays, just a simple flag `multiple` would be needed for the arg? Also requiring you to use the callback.

Since now it would call the callback once and then error. I'm fine with arrays not being available by default and requiring custom implementation.

alexveden:
FYI, I found array args impractical in most cases, I barely can remember anything I used with array args except maybe gcc :). For simple use, it's possible to use `--flag` + `array of arguments`

tomaskallup:
That's what most tools do, single flag with values separated by some character. But sometimes you might want those values to be arbitrary strings and there might not be a feasible separator, like when specifying ENV variables for docker etc.

lerno:
I am sorry this one isn't looked at yet. It's half past midnight and I don't have the time this lib deserves to check it. I'll need to push it to the weekend.

lerno:
Maybe I'm not the kind of audience who is using something like this, but for me it's more natural with a simpler design, as you might have guessed from the way build_options.c work.

It is quite simple: have a switch which looks at each arg, then if the arg starts with `-` it instead runs through the switch with `-` opts, and if it finds another `-` then that's a long opt and will be checked with the longopts.

This way checking is trivially stateful, which can be useful.

So the useful functionality is not parsing the arguments but rather:

1. Skip an argument
2. Check if a string (argument) is a vaild file or directory
3. Check if a string (argument) is an int
4. Check if a string (argument) is one in a list of values, and return that index.

What are your thoughts?

hwchen:
My [library](https://github.com/hwchen/opter-c3/blob/main/examples/basic.c3) is also basically a big switch, but there's some additional lexing it does which I think is an improvement over just checking for `-` or `--`.

- it's easy to group the long and short options as the same case
- separator boilerplate: allows either `' '` or `=` (or unseparated short) for opt-value separator
- handles short `-abc` for short flags with no values
- lexing is fairly well tested

I had also added some convenience methods for parsing ints and paths, although I don't think they're really necessary.

I might change up the error handling a bit; I want some decent canned error messages, but I also want it to be easy to silence when the user wants custom error messages.

Example switch in loop:
```
struct Opts {
	Maybe(<String>) thing;
	uint number;
	bool shout;
}

fn void! parse_cli(String[] args, Opts* opts) {
	Opter opter;
	opter.init(args);

	while (true) {
		Arg arg = opter.next()!;
		if (arg.type == EOF) break;
		switch {
			case arg.is(SHORT, 'n'):
			case arg.is(LONG, "number"):
				opts.number = opter.value()!.as_int(uint)!;

			case arg.is(LONG, "shout"):
				opts.shout = true;

			case arg.is(VALUE):
				if (!opts.thing.has_value) {
					opts.thing = { .value = arg.value.as_str(), .has_value = true };
				}

			case arg.is(LONG, "help"):
				io::printn("Usage: hello [-n|--number=NUM] [--shout] THING");

			default:
				return opter.err_unexpected_arg();
		}
	}
}
```

lerno:
Just a comment there, this:
```c
while (true) {
		Arg arg = opter.next()!;
		if (arg.type == EOF) break;
		switch {
```
Should be possible to rewrite as:
```c
while (try arg = opter.next())
{
  switch { ... }
}
```

lerno:
But I think the one benefit of a module is to create some uniform way of presenting the help. I think the actual parsing might be something of a red herring.

```c
enum Options : (String short_opt, String long_opt, String description)
{
   UNKNOWN = { "", "", "" }, // Mandatory
   VALUE = { "", "", "The value" },
   N = { "n=NUM", "number=NUM", "The number to use" },
   SHOUT = { "", "shout", "Shout out" }   
}
```
In the above then, we can imagine:
```c
@parse_opt("foo.exe [options]", strings, Options; @body(OptParser* parser, OptArg arg)
{
    switch (arg.type)
    {
        case UNKNOWN:
            // Handle anything else here
            // Example using another argument
            OptArg! next = parser.next();
            if (catch next_thing = parser.next())
            {
                return report_error("Expected an argument after %s.", arg.string);
            }
            io::printfn("The argument after was %s.", next.string);
        case VALUE: 
            // Value arguments are handled in UNKNOWN
            unreachable();
        case N:
            ...
        case SHOUT:
            ...            
    }
}    
```
Becoming:
```
Usage: foo.exe <VALUE> [options]
  <VALUE>                 The value
  -n, --number=NUM        The number to use
  --shout                 Shout out
  -h, --help              Show this help
```
But then we can imagine essentially a mini DSL defining things instead:
```c
enum Options : (String short_opt, String long_opt, String description)
{
   UNKNOWN = { "EMPTY" }, 
   VALUE = { "The value" },
   N = { "n=;number=;NUM;The number to use" },
   DEBUG = { "g;debug;Use debug" }, // -g, --debug
   SHOUT = { "shout;Shout out" }   
}
```
It all depends on how much is in the DSL and how much is in the language. It's nice if the description and the commands are in sync, and that's the basic service provided. Then providing those utility methods for matching multiple options and so on.

alexveden:
Let me explain and reason about my design decisions for `argparse`. First, it's designed with [Python's argparse](https://docs.python.org/3/library/argparse.html) library.  Which I used in many projects and find handy. 

C3 argparse is designed with the following requirements in mind:
- It abstracts grunt work of matching argument type, parsing it, and making sanity checks (e.g. int is "123" but not "asd")
- it doesn't allocate memory, and `ArgParse` structure is self containing, data is stored in `.values` and the program resposible for its lifetime
- It automatically checks if option is required
- It doesn't require creating extra type dedicated only for parsing
- It generates help printout under the hood
- It supports default values (so the value of C3 variable which passed to argparse settings structs is the default)
- It detects option type by the type of `.value=` item, e.g. `.value = &foo,` if `foo`  is bool it's a flag, if int - it's a numerical, String and other (which customizes validation method without any extra configuration from user side)
- it refines non-optionable arguments and store them as separate array of strings
- it supports `--` which in POSIX makes all following arguments (even with `--...` prefix) arguments
- it supports both `-s` and `--long` option syntax
- it supports short stacking `-sev` is equal to `-s -e -v`
- supports `=" in  `--long-opt=`  or `--long-opt`
- supports subcommands vie early stopping, for example `main.exe --foo command --bar -b -f`, with a special flag can stop at `command --bar -b -f`, so you can pass this array of arguments into another instance of argparse which support command. Whereas `--foo` option was passed to main argparse.
- for extensibility you would be able to pass callback function to argparse if you need to do some validation or fancy argument parsing 

@hwchen alternative library looks more like syntactic sugar to me, which requires working through each argument separately, and making printout manually.

In order to have fair comparison, we need to implement the same functionality using alternative argparse libraries. And judge which one provides cleaner or shorter or more flexible code. It has to be different use case, different options and argument types, maybe sub commands.

hwchen:
Not sure how I feel about it, but I implemented @lerno 's suggested API on top of my library (without help generation yet; I don't think that's the difficult part).

Implementation: https://github.com/hwchen/opter-c3/blob/691e5b19ba5cc8e45418f7e717c1351072f083f3/fancy.c3
Example: https://github.com/hwchen/opter-c3/blob/691e5b19ba5cc8e45418f7e717c1351072f083f3/examples/fancy.c3

I think this demonstrates one perspective on my library: it's the minimum amount of lexing/parsing code that's still actually helpful. With the basic lexing/parsing out of the way, it's easy to focus on adding features if building a more advanced cli library, or handling some really twisted config logic in an application. To me, this gives a feeling of "hackability", which I associate with C3.

So if we end up taking this general approach of just a switch statement (and using some of my code), I think it would be nice if the "lower-level" was still exposed, even if there was a higher-level API.

alexveden:
Example of using callbacks, 2 flavors of callbacks:
```zig
fn void test_custom_type_callback_unknown_type()
{

    char val = '\0';
    ArgParseCallbackFn cbf = fn void! (ArgOpt* opt, String value) {
        io::printfn("flt--callback");
        test::eq(value, "bar");
        *anycast(opt.value, char)! = value[0];
    };
    // NOTE: pretends app struct
    ArgParse my_app_state = {};

    ArgParse agp = {
        .options = {
            {
                .short_name = 'f',
                .long_name = "flt",
                .value = &val,
                .callback = cbf
            },
            {
                .short_name = 'o',
                .long_name = "other",
                .value = &my_app_state,
                .callback = fn void! (ArgOpt* opt, String value) {
                    ArgParse* ctx = anycast(opt.value, ArgParse)!;
                    io::printfn("other--callback");
                    // NOTE: pretends to update important app struct
                    ctx.usage = value;
                }
            }
        }
    };

    String[] args = { "testprog", "--flt=bar", "--other=my_callback" };
    test::eq(val, '\0');
    test::eq(agp.options[0]._is_present, false);

    agp.parse(args)!!;

    test::eq(val, 'b');
    test::eq(my_app_state.usage, "my_callback");
    test::eq(agp.options[0]._is_present, true);
}
```

alexveden:
@tomaskallup With small amendments I managed to add multiple options parsing via callbacks

```zig
fn void test_custom_type_callback_int_accumulator()
{

    List(<int>) numbers;
    numbers.new_init(); 
    defer numbers.free();

    argparse::ArgParse agp = {
        .options = {
            {
                .short_name = 'n',
                .long_name = "num",
                .value = &numbers,
                .callback = fn void! (ArgOpt* opt, String value) {
                    io::printfn("value: %s", value);
                    List(<int>) * ctx = anycast(opt.value, List(<int>))!;
                    int val = value.to_integer(int)!;
                    ctx.push(val);
                }
            },
        }
    };

    String[] args = { "testprog", "--num=1", "-n", "5", "--num", "2" };
    test::eq(numbers.len(), 0);
    test::eq(agp.options[0]._is_present, false);

    agp.parse(args)!!;

    io::printfn("%s", numbers);
    test::eq(numbers.len(), 2);
    test::eq(numbers[0], 1);
    test::eq(numbers[1], 5);
    test::eq(numbers[2], 2);
    test::eq(agp.options[0]._is_present, true);
}

```

```Output

value: 1
value: 5
value: 2
[1, 5, 2]

```

alexveden:
So currently I evolve `argparse` module as side project, if you greenlight this, I'll add updated version + unit tests.

hwchen:
There was some discussion around an example scenario (compiler with optimization flags and subcommands) so I implemented it: https://github.com/hwchen/opter-c3/blob/32ea95b00d78912119ffd0885287ba5c96c72d0c/examples/optimization.c3

- handles subcommands
- handles an -O flag which sets several optimizations
- more-specific overrides (over the -O level) for each optimization

I also have some notes on handling order of parsing flags, some of which I listed in the example.

- There's some work for the dev to apply specific overrides after the optimization level is set; this way, the config doesn't depend on the order of the `-O` and specific optimization flags.
- For specific optimizations, the `--no-x` overrides `--x` if both are present. This is just the dev's choice, could be the other way around. Or even just last-called (which would take more work if we need to apply only after `-O` optimizations are set)
- Also, for repeating flags (which we don't need repeated) and conflicting flags in general, the cli could just return an error instead of prioritizing. It's up to the dev.
- Could disallow flags before subcommand by calling `opter.next()` just for the subcommand, before entering the loop parsing the rest of the flags.
- There's nothing stopping implementation of global flags (in the worst case, it's a bunch of boilerplate). It would be more convenient if there were a way to merge enums, probably. I don't know that it makes sense to create any specific API around this, especially as subcommands themselves are pretty much hand-rolled.

So, the burden is definitely on the dev to handle flag interactions :)

---

Finally, note that the subcommand enums are not declared as part of one data structure (there's a separate enum for each subcommand), so it won't automatically generate a full help. Each subcommand can get its own help generated, except that its child subcommands won't be listed.

In fact, the fact that subcommands aren't declared in a data structure makes the enum API feel a bit inconsistent, since you have to look in the fn to check for subcommand names while flag names are in an enum.

I suppose that there's a way to declare subcommands in an enum, but then there's probably more API choices to be made (I haven't thought about this too much).

alexveden:
ok @hwchen challenge accepted! :) This is ArgParse version of the same problem:
https://github.com/alexveden/c3tools/blob/main/src/examples/argparse_subcommand_example.c3

It also:
1. Handles sub-commands
2. Handles custom logic of `-O` command
3. Handles flag business under the hood, including negation i.e. `--no-batch-opts` turn config flag to false
4. It doesn't require creating extra types for specifically parsing task (I simply don't get it why we need that)
5. As a side bonus, it can display help message separately for main program and subcommand.
6. No helper functions for parsing stuff, straight to business logic

Help:
```
Launching ./build/c3test --help

Usage:
c3test compiler [options] {compile,sub_cmd} arg1...argN

Compiler test
    -h, --help        show this help message and exit
    --batch-ops       batch_ops [default: true]
    --zero-loop       zero_loop [default: true]
    --reduce-loops    reduce_loops [default: false]
    --remove-dead     remove_dead [default: false]
    -O                optimization level [default: O0]
```

Subcommand help:
```
Launching ./build/c3test compile --help

Usage:
compile [options] [--] [arg1 argN]
Compile command flags
    -h, --help    show this help message and exit
    -o=<str>      compilation output [default: a.out]
```

Full job:
```
Launching ./build/c3test -O1 --no-batch-ops compile -ofoo.a foo.c3

optimization level arg callback: -O1
Compile command:
 - out: foo.a
Compile arguments:
 - [foo.c3]
optimization flags:
 - batch_ops: false
 - zero_loop: true
 - reduce_loops: false
 - remove_dead: false

```

hwchen:
This was not exactly a challenge :) . I do prefer the style of my library in the std lib, but not because I think it's better. For me, it fits how I think about the stdlib and c3, which is that they are oriented towards the most basic building blocks. Our visions may differ on what is the most useful building block, but it's not something we can compete on (at least not in the way we can judge if one of our implementations is better).

(I'm not exactly sure what @lerno 's vision is, although I would guess that with a switch style, using familiar language constructs and having a simple implementation/usage would be priorities)

So I don't think our implementations can really be directly compared, as they're trying to walk different paths (even if the end-product is the same). The choice is down to which one of those different paths we want in the stdlib.

---

Maybe I'm reading too much into it, but when you say that argparse handles things "under the hood", or that you can go "straight to business logic", I get the idea that you're comparing on convenience or overall brevity of code. For sure, argparse is more convenient and does more than my library.

I guess the question is whether c3 strives to put that level of convenience into the stdlib. Languages like Rust and Zig do not, although Odin does (and more, as you just declare the cli API on a struct). Python does (I know you patterned argparse on the python version), but I also consider it in a different class as a higher-level scripting language (opinions may differ on whether this matters).

@lerno mentioned that a switch + help generation (that's where the extra enum comes from) would be sufficient convenience, so it's not clear to me that features like flag negation count as a plus.

---

re: specifics of our implementations of this interface, I spend a decent amount of lines to ensure that e.g. `--no-batch-ops` could be place before or after `-O1`. From a quick test, it looks like your example would also need post-argparse logic for this, as currently the last flag will clobber. But sorry if I'm incorrect, I might have made a mistake trying it.

alexveden:
This will be distributed as library: https://github.com/alexveden/c3tools/tree/main/lib/c3lib/argparse/

Issue #1898, opened 2025-01-27
signature and body do not match for allocator::new_aligned


rhmoller:
The macro `allocator::new_aligned` does not work.

The body references an `allocator` which is not part of the arguments

```c3
macro new_aligned($Type, ...) @nodiscard
{
	$if $vacount == 0:
		return ($Type*)calloc_aligned(allocator, $Type.sizeof, $Type.alignof);
	$else
		$Type* val = malloc_aligned(allocator, $Type.sizeof, $Type.alignof);
		*val = $vaexpr[0];
		return val;
	$endif
}
```
Most other macros in the module take an `allocator` as first argument.

As a second note the `malloc_aligned` macro returns an optional so it cannot be assigned to `$Type* val`. I assume the optional should be rethrown.

PR #1899, opened 2025-01-27
fix allocator::new_aligned #1898

rhmoller:
Add `Allocator` as first argument for `allocator::new_aligned` and rethrow the optional returned from `malloc_aligned`.


lerno:
Thank you for finding and fixing the bug!

PR #1901, opened 2025-01-28
Fix missing r15 register in inline assembly

rexim:
Steps to reproduce:

```console
$ c3c --version
C3 Compiler Version:       0.6.7 (Pre-release, Jan 28 2025 20:50:05)
Installed directory:       /home/rexim/opt/c3/bin/
Git Hash:                  02c3d5419be88ad905445c33f8c9aa6d8d3236a5
$ cat ./main.c3
fn void test() @naked {
    asm {
        pushq $r15;
        ret;
    }
}

fn void main() {
    test();
}
$ c3c compile ./main.c3
1: fn void test() @naked {
2:     asm {
3:         pushq $r15;
                 ^^^^
(/home/rexim/Programming/probe/c3-inline-assembly-r15/main.c3:3:15) Error: Expected a valid register name.

$
```

### DIFF
--- a/src/compiler/asm_target.c
+++ b/src/compiler/asm_target.c
@@ -721,10 +721,10 @@ static void init_asm_x86(PlatformTarget* target)
 	}
 	else
 	{
-		reg_register_list(target, x64_quad_regs, 15, ASM_REG_INT, ARG_BITS_64, X86_RAX);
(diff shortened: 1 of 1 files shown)


lerno:
Ouch! Thank you!

PR #1914, opened 2025-01-31
Fixed nix build for macos.

diasbruno:
I'm not an expert on nix, so I'd appreciate some help with this.

On macOS, the nix was failing because it was finding the llvm 16 on the nix hash.

I updated the flake, and added the missing piece "compiler-rt" to find the rest of the runtime libraries.

This hash pushes llvm to version 19.


nix/default.nix
```diff
@@ -38,12 +38,13 @@ in llvmPackages.stdenv.mkDerivation (finalAttrs: {
...
     "-DC3_ENABLE_CLANGD_LSP=${if debug then "ON" else "OFF"}"
     "-DC3_LLD_DIR=${llvmPackages.lld.lib}/lib"
-  ];
+  ] ++ lib.optionals llvmPackages.stdenv.hostPlatform.isDarwin [ "-DC3_LLVM_CRT_LIBRARY_DIR=${llvmPackages.compiler-rt}/lib/darwin" ];
 
   nativeBuildInputs = [ 
     cmake 
     llvmPackages.llvm
     llvmPackages.lld 
+    llvmPackages.compiler-rt
```
diasbruno, line 47: Is compiler-rt now for every platform? This is the missing piece of the puzzle.

diasbruno:
I tested on a Ubuntu on WSL and it compiled and run ok.

nix/default.nix
diasbruno: It works just fine.

lerno:
This one looks wrong. You'd adding the configurable path yes, but you don't default to the LLVM one, so any build outside of nix breaks.

diasbruno:
You are right. I forgot to build with the local llvm. 🤦‍♂️

lerno:
Wouldn't it make more sense to make the override to change from the default if the override is set to a value, rather than have it depend on nix?

So "if LLVM_CRT_LIBRARY_DIR is empty string, use the default, otherwise use the code with LLVM_CRT_LIBRARY_DIR"

diasbruno:
This seems to be Apple only so I decided to only set on this if. But I'm ok setting as default.

lerno:
I meant to get rid of the C3_IS_NIX_BUILD entirely and only use LLVM_CRT_LIBRARY_DIR if it is set to a non-empty value.

CMakeLists.txt
```diff
@@ -261,10 +263,13 @@ if(C3_WITH_LLVM)
 
     if (APPLE)
         set(lld_libs ${lld_libs} xar)
-        find_file(RT_ASAN_DYNAMIC NAMES libclang_rt.asan_osx_dynamic.dylib PATHS "${LLVM_LIBRARY_DIR}/clang/${LLVM_MAJOR_VERSION}/lib/darwin")
-        find_file(RT_TSAN_DYNAMIC NAMES libclang_rt.tsan_osx_dynamic.dylib PATHS "${LLVM_LIBRARY_DIR}/clang/${LLVM_MAJOR_VERSION}/lib/darwin")
-        find_file(RT_UBSAN_DYNAMIC NAMES libclang_rt.ubsan_osx_dynamic.dylib PATHS "${LLVM_LIBRARY_DIR}/clang/${LLVM_MAJOR_VERSION}/lib/darwin")
-        find_file(RT_LSAN_DYNAMIC NAMES libclang_rt.lsan_osx_dynamic.dylib PATHS "${LLVM_LIBRARY_DIR}/clang/${LLVM_MAJOR_VERSION}/lib/darwin")
+        if (C3_IS_NIX_BUILD)
+                set(LLVM_CRT_LIBRARY_DIR "${LLVM_CRT_LIBRARY_DIR}/lib/darwin")
+        endif()
```
diasbruno, line 268:
```suggestion
```

CMakeLists.txt
```diff
@@ -62,9 +62,11 @@ option(C3_LINK_DYNAMIC "link dynamically with LLVM/LLD libs")
 set(C3_LLVM_VERSION "auto" CACHE STRING "Use LLVM version [default: auto]")
 option(C3_USE_MIMALLOC "Use built-in mimalloc" OFF)
 option(C3_USE_TB "Use TB" OFF)
+option(C3_IS_NIX_BUILD "Is building with nix" OFF)
```
diasbruno, line 65:
```suggestion
```

nix/default.nix
```diff
@@ -38,12 +38,15 @@ in llvmPackages.stdenv.mkDerivation (finalAttrs: {
   cmakeFlags = [
     "-DC3_ENABLE_CLANGD_LSP=${if debug then "ON" else "OFF"}"
     "-DC3_LLD_DIR=${llvmPackages.lld.lib}/lib"
+    "-DC3_IS_NIX_BUILD=ON"
+    "-DLLVM_CRT_LIBRARY_DIR=${llvmPackages.compiler-rt}"
```
diasbruno, line 42:
```suggestion
    "-DLLVM_CRT_LIBRARY_DIR=${llvmPackages.compiler-rt}/lib/darwin"
```

lerno:
Hmm.. that failed

lerno:
Maybe this is better:
```
option(LLVM_CRT_LIBRARY_DIR "Use custom llvm's compiler-rt directory" "")
...
find_file(RT_ASAN_DYNAMIC NAMES libclang_rt.asan_osx_dynamic.dylib PATHS ${LLVM_CRT_LIBRARY_DIR} "${LLVM_LIBRARY_DIR}/clang/${LLVM_MAJOR_VERSION}/lib/darwin")
find_file(RT_TSAN_DYNAMIC NAMES libclang_rt.tsan_osx_dynamic.dylib PATHS ${LLVM_CRT_LIBRARY_DIR} "${LLVM_LIBRARY_DIR}/clang/${LLVM_MAJOR_VERSION}/lib/darwin")
find_file(RT_UBSAN_DYNAMIC NAMES libclang_rt.ubsan_osx_dynamic.dylib PATHS ${LLVM_CRT_LIBRARY_DIR} "${LLVM_LIBRARY_DIR}/clang/${LLVM_MAJOR_VERSION}/lib/darwin")
find_file(RT_LSAN_DYNAMIC NAMES libclang_rt.lsan_osx_dynamic.dylib PATHS ${LLVM_CRT_LIBRARY_DIR} "${LLVM_LIBRARY_DIR}/clang/${LLVM_MAJOR_VERSION}/lib/darwin")
```

lerno:
Do you want me to look at that?

diasbruno:
Sure. I didn't had a chance to go back to this.

lerno:
Thank you!

diasbruno:
Good job! 👏

PR #1915, opened 2025-02-01
Removed the unused command `headers`from the `usage()` in `src/build/build_options.c`

movva-gpu:
There is also one internal command being undocumented, which is `utest`.

image

```c
if (arg_match("utest"))
{
	options->command = COMMAND_UNIT_TEST;
	return;
}
```

The name of the constant being used here implies that `utest` means "unit test", since it probably is a function that is used to debug and test some specific things, I think it's okay if it stays undocumented.


lerno:
Thank you!

movva-gpu:
You're welcome!

PR #1916, opened 2025-02-02
Clean up some linker/C compiler options

MaskRay:
-fno-pic/-fno-pie/-fpic/-fPIC/-fpie/-fPIE options belong to the same famility where the last option wins. These options have no effect in the link phase.

Clang and GCC usually pass `--eh-frame-hdr` to ld, with the exception that `gcc -static` does not pass `--eh-frame-hdr`. The difference is a historical choice related to `__register_frame_info`. We can behavle like Clang and always pass `--eh-frame-hdr`.

Remove a `-L` that does not specify a directory.

### DIFF
--- a/src/compiler/linker.c
+++ b/src/compiler/linker.c
@@ -388,7 +388,7 @@ static void linker_setup_linux(const char ***args_ref, Linker linker_type, bool 
 		if (is_pie(compiler.platform.reloc_model)) add_plain_arg("-pie");
 		if (is_no_pie(compiler.platform.reloc_model)) add_plain_arg("-no-pie");
 	}
-	if (compiler.platform.arch == ARCH_TYPE_X86_64) add_plain_arg("--eh-frame-hdr");
+	add_plain_arg("--eh-frame-hdr");
 	if (!link_libc()) return;
 	const char *crt_begin_dir = find_linux_crt_begin();
 	const char *crt_dir = find_linux_crt();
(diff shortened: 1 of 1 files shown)


MaskRay:
* https://maskray.me/blog/2020-11-08-stack-unwinding#eh_frame_hdr-and-pt_gnu_eh_frame

Side notes (not done in this PR as I am not familiar with the use cases)

* [`--export-dynamic`](https://maskray.me/blog/2020-11-15-explain-gnu-linker-options#export-dynamic), only relevant for executable, probably should not be coupled with `DEBUG_INFO_FULL`. It could make .dynsym very large
* `--gc-sections` could be enabled for -shared as well. Defined symbols that are exported to .dynsym will not be garbage collected. https://maskray.me/blog/2021-02-28-linker-garbage-collection#ld.lld

lerno:
For pie/pic/PIE/PIC here I've tried to follow Clang mostly. I see you removed a bunch of options, but pie and pic are interdependent in Clang what I could note.

MaskRay:
> For pie/pic/PIE/PIC here I've tried to follow Clang mostly. I see you removed a bunch of options, but pie and pic are interdependent in Clang what I could note.

For -fpic/-fpie/-fno-pic/-fno-pie options the last wins: https://github.com/llvm/llvm-project/blob/05a09e6e559e8253d49cc61052711f0c200129bf/clang/lib/Driver/ToolChains/CommonArgs.cpp#L1907

Regarding the module flags metadata "PIC Level" and "PIE Level". They are independent.

lerno:
All right, good!

lerno:
Thank you!

PR #1923, opened 2025-02-02
`clock_gettime` should return `CInt`, not `void`

radekm:
According to the documentation https://www.man7.org/linux/man-pages/man3/clock_gettime.3.html function
`clock_gettime` returns `CInt`. Here is the signature from the documentation:

```c
int clock_gettime(clockid_t clockid, struct timespec *tp);
```


lerno:
Thank you!

Issue #1918, opened 2025-02-02
New `test::` methods are unnecessarily verbose


alexveden:
Let me speak as a person who wrote the new `std::core::test` module and used 1571 of `test::` assertions in my current projects. Just in one month, I wrote almost the same amount of test cases as c3 std lib has. Currently, I made 99.9% of c3 `test::` use cases in the universe :)

So I think a decision of making `test::is_equal()` is unnecessarily verbose and doesn't add any value to the test readability, but significantly decreases test writability. Test cases are a special type of functions, which are written as "walls" of repeating text (e.g. `test::equals`).  

Let's put it into context, what we face when dealing with tests:
1. "test_" in file name and `@test` attribute in the function name, indicates a test case
2. Bunch of repeating `test::` "walls" of code, indicates a test case
3. There are many many functions with the same pattern of code organization in the test module.
4. We really read the test only when it fails, otherwise no one gives a sh*t about it.
5. Short names are faster to read, especially when we have a wall of them, the reader can quickly get used to it. It's like an Assembly language.

## Verbosity issue
Let's compare 3 naming conventions of the test cases.

### Step 1
Imagine this example below. And try to read `test::is_equal` out of loud 8 times. I bet you will end up with wanting to skip `test::is_` part, because `equal`` is the essence of the code line.
![Image](https://github.com/user-attachments/assets/9af87bbf-f1db-4372-832c-142506ad8e00)

### Step 2
What if we could remove redundant `is_` part. Do you feel any degradation of code readability?

![Image](https://github.com/user-attachments/assets/10764307-406f-4b13-8d6f-3d93ee279d3c)

### Step 3
What if we could remove redundant `qual` part. Do you feel any degradation of code readability?
![Image](https://github.com/user-attachments/assets/896dd07c-1a84-48d7-b536-a33e2e4b9063)

## Visual noise issue
When you have a wall of repeating test, the alignment of calls does matter to readability more than naming. Current, `test::` naming are bad from alignment standpoint.

### Example 1 (existing std::test)

![Image](https://github.com/user-attachments/assets/1dc2bcf8-6909-41b2-aba6-834c723a3829)

### Example 2 (no `is_`)

![Image](https://github.com/user-attachments/assets/29f0eff9-28f9-443c-8554-e4946e6b6557)

### Example 3 (`2 letter`)
![Image](https://github.com/user-attachments/assets/6c7900df-7d27-4371-8fa2-0e73c0a95f70)

## `test::equal` only matters
[...]

PR #1928, opened 2025-02-05
test::eq/ne/gt syntax

alexveden:
Terse tests syntax proposal #1918 

The test module is refactored in the following logic:
1.  2-letters `eq/ne/gt/le` are for regular use (90% of the test code), they are macros but without `@`.
2. Special cases have `@` prefix, and also reduced to 1 word (it's better to remember). Also, `@` will be a reminder that we might have special behavior (less mental overhead for reader).
- `test::setup` now `test::@setup`
- `test::is_almost_equal` now `test::@almost`
3. Let's keep `test::` module instead of introducing new `assert_eq/ne/gt` keyword, because in the future I think it's possible to add more testing features (mocks maybe?). And `test::` is visually more readable due to `::` separation and different highlighting.

### DIFF
--- a/lib/std/core/test.c3
+++ b/lib/std/core/test.c3
@@ -27,7 +27,7 @@ fn void! test_div() @test
 	test::gt(2, divide(3, 3)!);
 	test::lt(2, 3);
 	test::le(2, 3);
-	test::@almost(m::divide(1, 3)!, 0.333, places: 3);
+	test::eq_approx(m::divide(1, 3)!, 0.333, places: 3);
 	test::@check(2 == 2, "divide: %d", divide(6, 3)!);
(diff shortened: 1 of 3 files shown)


lib/std/core/test.c3
```diff
@@ -127,7 +127,7 @@ macro is_equal(left, right)
  @require delta >= 0, delta <= 1 "delta must be a small number"
  @require runtime::test_context != null "Only allowed in @test functions"
 *>
-fn void is_almost_equal(double left, double right, uint places = 7, double delta = 0, bool equal_nan = true)
+macro void @almost(double left, double right, uint places = 7, double delta = 0, bool equal_nan = true)
```
lerno, line 130: eq_delta

PR #1929, opened 2025-02-06
Introduce Socket.shutdown()

rexim:
One of the common operations used in TCP programming is shutting down parts of the full-duplex connection [[1]](https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-shutdown) [[2]](https://www.man7.org/linux/man-pages/man2/shutdown.2.html). In c3ws [[3]](https://github.com/tsoding/c3ws) we are employing the technique described in [[4]](https://blog.netherlabs.nl/articles/2009/01/18/the-ultimate-so_linger-page-or-why-is-my-tcp-not-reliable) on closing the connection to ensure more reliable delivery of data which relies on the shutdown operation. But since c3ws is supposed to work with different variety of sockets  (blocking, non-blocking, encrypted, etc, etc) via an abstract interface to separate concerns, it would be nice if the shutdown operation was part of the cross-platform Socket interface.

This PR tries to introduce such operation for POSIX and WIN32 environment. I haven't tested it on WIN32 though, but it should be almost the same as on POSIX. 

References:
[1] [shutdown function (winsock.h)](https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-shutdown)
[2] [shutdown(2) — Linux manual page](https://www.man7.org/linux/man-pages/man2/shutdown.2.html)
[3] [c3ws - WebSocket library in C3 ](https://github.com/tsoding/c3ws)
[4] [The ultimate SO_LINGER page, or: why is my tcp not reliable](https://blog.netherlabs.nl/articles/2009/01/18/the-ultimate-so_linger-page-or-why-is-my-tcp-not-reliable)

### DIFF
--- a/lib/std/libc/os/posix.c3
+++ b/lib/std/libc/os/posix.c3
@@ -1,5 +1,10 @@
 module libc @if(env::POSIX);
 
+const CInt SHUT_RD = 0;
+const CInt SHUT_WR = 1;
+const CInt SHUT_RDWR = 2;
+extern fn CInt shutdown(Fd sockfd, CInt how);
+
 extern fn isz recv(Fd socket, void *buffer, usz length, CInt flags);
 extern fn isz send(Fd socket, void *buffer, usz length, CInt flags);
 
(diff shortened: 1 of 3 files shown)


lerno:
Can I get edit rights to this branch so that I can add a small update I wanted to do to the pull request?

lerno:
Ok, I'll accept this and update in master.

Issue #1930, opened 2025-02-06
Compile tests without running


hwchen:
Currently, using `compile-test`, the testrun executable will be compiled and then automatically run. However, there may be some cases when it's not desirable to immediately run, such as compiling test executable for use with debugger.

This can be implemented with a flag like `--no-run`. `--build-only` has also been suggested. As a point of posible confusion, `compile-only` subcommand already exists, which compiles but does not link.

Not sure further discussion is needed, but as an initial thought, checking for this flag can just be placed after the check for `compile-test`, `compile-bench`, etc, and just switch off the run. Let me know if there's other considerations.

lerno:
Note that --testing and --benchmark already includes the test but without the actual test/bench runner created as the main.

PR #1931, opened 2025-02-06
add --no-run option

hwchen:
`--no-run` option.

For commands which may run executable after building (`compile-test`, `compile-bench`, `compile-run`, and `clean-run`, skip the "run" step.

See https://github.com/c3lang/c3c/issues/1930

### DIFF
--- a/releasenotes.md
+++ b/releasenotes.md
@@ -23,6 +23,7 @@
 - Test runner will also check for leaks.
 - Improve inference on `??` #1943.
 - Detect unaligned loads #1951.
+- Add flag `--no-run`. For commands which may run executable after building, skip the run step. #1931
 
 ### Fixes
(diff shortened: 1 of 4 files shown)


hwchen:
@lerno I know you'd also suggested `--build-only`. I saw that it's a bit similar to the command `compile-only` though.

Also, I think you may have mentioned some other concerns with this option in discord, but I'm not sure if they'll conflict with this implementation.

hwchen:
Looks like there's been a refactor around how build options are printed. But I don't know if that means the approach in this PR should change.

If we keep this approach, I'll update the PR

lerno:
I finally got around to this! I changed the name to --suppress-run and changed so that it's only on test/benchmarks. But I did it on a separate push. So it's in there but just not with this PR.

Thank you!

PR #1956, opened 2025-02-11
Add more Windows API types and structs

velikoss:
From [libuv.c3l/win.c3](https://github.com/velikoss/libuv.c3l/blob/0d3497e048de924d22af1de43f511552bfec3c9c/win.c3)


lerno:
Thank you!

PR #1957, opened 2025-02-11
Add UNIX boolean

velikoss:
Add UNIX boolean to check OS type more properly


lerno:
I don't think this is correct. POSIX doesn't imply unix, does it? Or if it does, then the rest of the checks are useless because BSD and DARWIN already implies POSIX.

velikoss:
Actually, even if Darwin and BSD are not 100% POSIX compatible, if you look that way it actually is just @if(env::POSIX)
Then this pull is unnecessary. Sorry for that
Source: https://www.reddit.com/r/linuxquestions/comments/143kyi2/whats_the_difference_between_unixlike_and/?rdt=56612

PR #1962, opened 2025-02-12
test runner arguments fix + fixed memcheck for teardown function

alexveden:
I fixed some bugs added refinements to test runner:
1. BUG c3c had `--test-disable-sort`, but test runner used `--test-nosort`
2. BUG teardown function was called after memory leak detection scope, so if memory was cleaned in teardown it didn't track by allocator and memory leak error was emitted
3. Added `--test-leak-report` to enable full report on leak, which was very verbose (default behavior now is `LEAKS DETECTED` print)
4. Refactored default arguments, added `--test-` prefix in order to maintain same argument syntax when running `./build/testrun` program directly. This is very important to me, when I launch tests in debugger.
5. Added short arguments and usage help in the case of direct `./build/testrun`.

### DIFF
--- a/lib/std/core/runtime_test.c3
+++ b/lib/std/core/runtime_test.c3
@@ -279,6 +279,7 @@ fn bool run_tests(String[] args, TestUnit[] tests) @private
 		name.appendf("Testing %s ", unit.name);
 		name.append_repeat('.', max_name - unit.name.len + 2);
 		io::printf("%s ", name.str_view());
+		(void)io::stdout().flush();
 		TrackingAllocator mem;
 		mem.init(allocator::heap());
 		if (libc::setjmp(&context.buf) == 0)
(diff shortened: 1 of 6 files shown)


alexveden:
I added `--test-quiet` flag which replaces test name by `.` which very helpful, when we have many tests, and some of them failing in random places. It drastically reduces amount of scrolling and unnecessary text.

image

lerno:
Please separate out these different things in different pull requests. Bundling fixes with features is problematic, because features may introduce new behaviour that needs to be handled in a different way, for example with deprecation and fallback.

alexveden:
I'm sorry about mixing everything together. But test runner changes from last weeks were frustrating for me and slowed my TDD workflow significantly. The only reason I coped with it because I was working hard on c3fmt tool. So this PR solves every problem I had with test runner, and it contains only backward compatible changes, and no API or deprecation side effects whatsoever. But I'm open to discussion.

Now I can see `Merging is blocked` red banner, so could you clarify guidelines for PRs I'm a little offtrack with this merging/rebasing things? Also, it would be useful to know red lines with PRs and contributing policy. I'd love to contribute more to c3c, but I need to have clear understanding of the process and reduce friction for PRs (on technical side).

PR #1974, opened 2025-02-15
[TEST] Add unit tests for libc

BWindey:
This PR aims to add unit tests for libc.

Along the way, I also discovered some issues with the bindings currently present. Some pointers being misaligned (`void *` instead of `void*`), but also a wrong return type for `libc::bsearch()`.

This is a draft PR, because I'm far from done, but would like some input on how thorough these tests should be. I feel like they shouldn't be that thorough, because they should be battle-tested C-code, we're just testing that we can call them from C3, and potentially providing an example for people.


lerno:
Is this still a draft?

BWindey:
Yes, I'm not ready yet with most functions.
I don't think I can write tests for every binding in libc, but I can at least check if they are right (correct types), and for the simple ones write a simple test to see if all works as expected.

I'm a bit busy at the moment, but would like to get this out of draft status within at most 2 weeks, preferably 1 week.

BWindey:
Update: I've been working on this when I have time, and the original timeline is a bit unrealistic. For  most functions I have to check if they are correct and if they are ISO C or just POSIX (and thus need `@if(!env::WIN32)`). The worst ones are where the specification does not the layout of a type, and need to be implemented differently on different platforms. 
It's an interesting learning experience for me, but it means that this will take a bit longer then expected.

lerno:
Let me know when it's ready. Alternatively we could cut the scope a little and merge what you have?

lerno:
And then you could add more later

BWindey:
Good idea, I'll get this ready for review as soon as possible then. Sad that I can't spend more time on this now.

lerno:
Thank you!

Issue #1977, opened 2025-02-16
bitstructs inside bitstructs


hypno2000:
Does it make sense allow `bitstruct` fields to be other `bitstruct`s, as they are basically integers as well?

I'm working on Vulkan bindings where this would come handy:

```c3
def Flags  = uint;

bitstruct GeometryInstanceFlagsKHR : Flags @overlap {
	bool triangle_facing_cull_disable       : 0;
	bool triangle_flip_facing               : 1;
	bool force_opaque                       : 2;
	bool force_no_opaque                    : 3;
	bool force_opacity_micromap_2_state_ext : 4;
	bool disable_opacity_micromaps_ext      : 5;
	bool triangle_front_counterclockwise    : 1;
	bool triangle_cull_disable_nv           : 0;
	bool triangle_front_counterclockwise_nv : 1;
	bool force_opaque_nv                    : 2;
	bool force_no_opaque_nv                 : 3;
}

bitstruct AccelerationStructureMatrixMotionInstanceNVBits : uint {
	uint                     instanceShaderBindingTableRecordOffset : 0..23;
	GeometryInstanceFlagsKHR flags                                  : 24..31;
}
```

Right now i get this error:
```
Error: 'GeometryInstanceFlagsKHR' is not supported in a bitstruct, only enums, integer and boolean values may be used.
```

This is the related C header:
```c
typedef enum VkGeometryInstanceFlagBitsKHR {
    VK_GEOMETRY_INSTANCE_TRIANGLE_FACING_CULL_DISABLE_BIT_KHR = 0x00000001,
    VK_GEOMETRY_INSTANCE_TRIANGLE_FLIP_FACING_BIT_KHR = 0x00000002,
    VK_GEOMETRY_INSTANCE_FORCE_OPAQUE_BIT_KHR = 0x00000004,
    VK_GEOMETRY_INSTANCE_FORCE_NO_OPAQUE_BIT_KHR = 0x00000008,
    VK_GEOMETRY_INSTANCE_FORCE_OPACITY_MICROMAP_2_STATE_EXT = 0x00000010,
    VK_GEOMETRY_INSTANCE_DISABLE_OPACITY_MICROMAPS_EXT = 0x00000020,
    VK_GEOMETRY_INSTANCE_TRIANGLE_FRONT_COUNTERCLOCKWISE_BIT_KHR = VK_GEOMETRY_INSTANCE_TRIANGLE_FLIP_FACING_BIT_KHR,
[...]

PR #1979, opened 2025-02-16
allow nested bitstructs #1977

hypno2000:
This is working for me for my (#1977) scenario.

I have no idea if it requires any other changes in compiler, or does it even make sense to support it, but opening this PR just in case.


lerno:
I'll close this since this would not be enough to implement it correctly.

PR #1991, opened 2025-02-21
std::math::bigint: Fixed init_with_u128 and init_with_array

JonasQuinten:
* `init_with_u128` creates negative number for values >= 2^127
 * `init_with_array` writes values to the input slice


lerno:
Thank you!

PR #2011, opened 2025-03-02
Fix Atomic.max

JonasQuinten:
Atomic.max used fetch_div instead of fetch_max.
I also corrected some copy-pasted comments and removed dead code.


lerno:
It's weird, shouldn't bools be allowed? Or bitstructs that lower to an int.

lerno:
I found some more bugs which I fixed. It's probably not quite right yet. Everything is severely lacking in tests.

PR #2015, opened 2025-03-03
Fix libc::stat on FreeBSD (x86_64)

erikbackman:
Fixes libc::stat segfault on FreeBSD for x86_64

I've tested on a 32bit system where it still segfaults so probably disable support until
fixed.

I ran in to this when trying to use file::exists which uses stat, even with stat fixed, perhaps file::exists should be using access instead (which works fine)?

Please let me know if anything should be changed.

### DIFF
--- a/lib/std/libc/os/freebsd.c3
+++ b/lib/std/libc/os/freebsd.c3
@@ -1,15 +1,14 @@
 module libc @if(env::FREEBSD);
 
-// Checked for x86_64, this is notoriously incorrect when comparing with Rust code etc
+// Checked for x86_64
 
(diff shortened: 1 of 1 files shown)


lerno:
So it works on 64 bit FreeBSD?

erikbackman:
> So it works on 64 bit FreeBSD?

Yes

lerno:
I think those were lifted from source files, but never tested, so it's great that you've checked them. I'd love to have a CI for *BSD variants but I also don't want to fiddle too much with GitHub CI...

lerno:
Thank you!

PR #2021, opened 2025-03-06
String.to_float("+") fails

stevemolloy:
Doing `String.to_float("+")` panics.

This is due to the string_to_real macro preemptively chopping a "+" or "-" from the start of the string.  Once this is done, the string is effectively empty, and so the `@require chars.len > 0` contract at the top of the `decfloat` macro is violated.

I wrapped the code that chops the +/- from the start of the string in a check for the length of the string, and this chop will only be done if the string is not 1 character long.

This fixes the bug.


stevemolloy:
I added a couple of tests for this.  Please note that I have been using C3 for only a few hours(!), and so am still learning.  I hope that I have respected coding styles and best-practices, but please forgive any mistakes.

lerno:
Excellent! Thank you.

PR #2027, opened 2025-03-08
Added `is_initalized` function to list, linkedlist, and anylist types

m0tholith:


lib/std/collections/anylist.c3
```diff
@@ -47,6 +47,8 @@ fn AnyList* AnyList.tinit(&self, usz initial_capacity = 16)
 	return self.init(tmem(), initial_capacity) @inline;
 }
 
+fn bool AnyList.is_initialized(&self) @inline => self.allocator && self.allocator.ptr != null;
```
lerno, line 50: It is sufficient doing `self.allocator != null`

lerno:
Thank you.

PR #2030, opened 2025-03-09
C3 Android logcat interface.

bokic:


lerno:
Thank you! I updated to tabs, since that's used elsewhere in the stdlib. One thing you might consider is using `enum LogPriority : (inline CInt val)` so that LogPriority implicitly converts to a CInt.

PR #2038, opened 2025-03-11
Generate compile_commands.json

bokic:
After build create compile_commands.json on root of the project so IDEs like vscode, zed, etc can properly parse c/h files.


lerno:
It looks like this is failing in CI, I think it needs to fix the permissions in some way.

bokic:
No, I need to detect ci and exclude this feature when biild on ci.

bokic:
There was already cmake C3_ENABLE_CLANGD_LSP option I missed.

PR #2040, opened 2025-03-11
Add cross products

jopadan:


lerno:
Can you please add some actual tests, look into test/unit how to construct them. You can then use `c3c compile-test <path to folder>` to test.

lib/std/math/math_vector.c3
```diff
@@ -51,6 +51,31 @@ fn double[<3>] double[<3>].unproject(self, Matrix4 projection, Matrix4 view) =>
...
+/* laplace */
+macro float[<*>].laplace2(self) => laplace2(self);
+macro float[<*>].laplace3(self, float[<*>] b) => laplace3(self, b);
+macro float[<*>].laplace4(self, float[<*>] b, float[<*>] c) => laplace4(self, b, c);
+/* cross[2,3,4] products */
+macro float[<*>].cross2(self, uint winding = 0x901) => cross2(self, winding);
+macro float[<*>].cross3(self, float[<*>] other) => cross3(self, other);
+macro float[<*>].cross4(self, float[<*>] b, float[<*>] c) => cross4(self, b, c);
+
+macro determinant2(a, b) { return ({ 1,-1} * a.xy * b.yx).sum(); }
```
lerno, line 67: Prefer:
```c
macro determinant2(a, b) => ({ 1,-1} * a.xy * b.yx).sum();
```
For these short functions.

jopadan: done

lerno:
Sorry, that part wasn't supposed to be there. Not that it harms anything but..

lerno:
One question I have is whether the vector operations are the most efficient this way (I'm thinking of the 4 cross product) or not, have you measured the resulting output compared to writing it out on -O1? Also, possibly the two different windings should be two different functions instead.

lerno:
I am confused. There is this: `macro float[<*>].determinant2(self, float[<*>] b) => determinant2(self,b);`

But this looks like it's assuming
```
| self.x self.y | 
| b.x       b.y  |
```
Which means it's not applicable to any float vectors that aren't len = 2. Also, it is only defined for floats, as opposed to both doubles and floats.

I am also a little concerned that only some of them have tests.

jopadan:
I don't see why the implementation should not work for vectors that aren't len = 2.
It supports len = n vectors with only taking the elements of determinan[m].

jopadan:
See updated implementation for double

jopadan:
Is your 2x2 post concern about column/row major format?

lerno:
I think it's fairly confusing to have this. If you really want the example taking a subset of the vector then that isn't hard. Like for the determinant case for a 4 vector and you want to do the 2x2 determinant, why not simply do it explicitly? `x.xy.det(y.xy)`

PR #2043, opened 2025-03-12
Added .hash() functions for vectors

sandervdbosch:
Currently only vectors up to 16 bytes in size can be used in a Hashmap due to the requirement of `$Type.alignof <= mem::DEFAULT_MEM_ALIGNMENT` on `allocator:new`
I recon allowing for larger types would require changes to the Hashmap module.

Suggestions on how to improve the functionalities are welcome.


lerno:
This looks wrong. You're hashing the vectors on their pointer, not on their content.

sandervdbosch:
It should be fixed now. I also added support for arrays of all the normal types.

lerno:
For the tests you want to create two vectors separately but with the same values and check that they give the same hash.

sandervdbosch:
I did do that in a test program I wrote, but since the normal hash functions did not do this I hadn't added it to the test yet.

lib/std/core/builtin.c3
```diff
@@ -423,16 +423,38 @@ macro @generic_hash(value)
...
+macro uint long[<*>].hash(long[<*>] i) => (uint)fnv32a::hash(((char*)&&i)[:$sizeof(i[0])*i.len]);
+macro uint long[*].hash(long[*] i) => (uint)fnv32a::hash(((char*)&&i)[:$sizeof(i)]);
 macro uint ulong.hash(ulong i) => @generic_hash(i);
+macro uint ulong[<*>].hash(ulong[<*>] i) => (uint)fnv32a::hash(((char*)&&i)[:$sizeof(i[0])*i.len]);
+macro uint ulong[*].hash(ulong[*] i) => (uint)fnv32a::hash(((char*)&&i)[:$sizeof(i)]);
 macro uint int128.hash(int128 i) => @generic_hash(i);
+macro uint int128[<*>].hash(int128[<*>] i) => (uint)fnv32a::hash(((char*)&&i)[:$sizeof(i[0])*i.len]);
+macro uint int128[*].hash(int128[*] i) => (uint)fnv32a::hash(((char*)&&i)[:$sizeof(i)]);
 macro uint uint128.hash(uint128 i) => @generic_hash(i);
+macro uint uint128[<*>].hash(uint128[<*>] i) => (uint)fnv32a::hash(((char*)&&i)[:$sizeof(i[0])*i.len]);
```
lerno, line 453: Please add space between operators to keep the code consistent, e.g. `$sizeof(i[0]) * i.len`

lerno:
If you take the address of char/ichar arrays / vectors they can be cast to a `char[]` and directly handed over to the hash function. No need to bitcast.

lerno:
Should I do it?

sandervdbosch:
I corrected the formatting but did not get the casts to `char[]` to work properly in some cases with alignment and `\0` at different points in the array or vector.

lerno:
Thank you

PR #2055, opened 2025-03-19
Added a progress bar to vendor-fetch in compiler section

tahadostifam:
Hi,

I wrote a progress bar for vendor-fetch. I hope you like it :)

Also I wanna mention that these two lines needs to be refactored, it does not seem pretty well :

```c
if (count == 0)	error_exit("Error: Failed to download any libraries.");
if (count < vec_size(options->libraries_to_fetch)) error_exit("Error: Only some libraries were downloaded.");
```

I'm okay to make the changes on these two lines but i rather to know your own idea about it.

Thanks,
Taha


lerno:
You should check that the output supports ansi first, use isatty() at least, but suppress it on windows. There is also a build setting for ansi you can use for overriding it.

tahadostifam:
Oh! You're right.

I forgot to check it.

I check that output is supporting ansi then we show the progress bar, if it's not supporting we just a show a plain text.

Is it okay now?

src/compiler/compiler.c
```diff
@@ -920,20 +920,20 @@ const char * vendor_fetch_single(const char* lib, const char* path)
 #endif
 
 void update_progress_bar(const char* lib, int current_step, int total_steps) {
-		int ansi_supported = isatty(fileno(stdout));
+	int ansi_supported = isatty(fileno(stdout));
```
lerno, line 923: On Win32 this is usually wrong, instead just assume it isn't available on windows.

src/compiler/compiler.c
```diff
@@ -920,20 +920,20 @@ const char * vendor_fetch_single(const char* lib, const char* path)
...
 		
     float progress = (float)(current_step + 1) / total_steps;
     int filled_length = (int)(progress * PROGRESS_BAR_LENGTH);
 
-		if (ansi_supported) {
+	if (ansi_supported) {
```
lerno, line 928: Can you also make sure that the style is consistent with the rest of the code, i.e. tabs, Allman brace style?

tahadostifam:
Yep. It's okay now.
Any other problem?

lerno:
Thank you!

PR #2061, opened 2025-03-23
String.c3 function parameters ambiguous

MaxBeaud:
Iv'e changed the parameters and vars throughout the `string.c3` file to make it more readable and more self evident from the method signature was is intended to be passed as parameters.
Eg: s => self, z => self, needle => delimiter ...


lib/std/core/string.c3
```diff
@@ -383,64 +383,64 @@ fn usz? String.index_of_chars(String s, char[] needle)
 <*
  Find the index of the first incidence of a character.
 
- @param [in] s
- @param needle : "The character to look for"
+ @param [in] self
```
lerno, line 386: In this case we prefer "needle", since that is the standard naming of what you search for in a string.

MaxBeaud: I think my original point still applies. To me, "character" is a lot more descriptive (even more for laymen) than needle. And to me at least, in the stdlib, being more verbose with the goal of making code more "readable", is a great trade-off.

PR #2066, opened 2025-03-27
Create project_schema.json

aleclarson:
By setting the `$schema` field of your `project.json` file to a URL leading to this JSON schema, your IDE may be able to provide auto-completion.

Alternatively, IDE-specific extensions may be able to "inject" this schema automatically, removing the need for an explicit `$schema` field, but that's for another PR, I suppose.

**Note:** The `c3c` command currently warns about the `$schema` field:
```
WARNING: Unknown parameter '$schema' in 'default target'
```

Example usage:
```json
{
  "$schema": "https://raw.githubusercontent.com/c3lang/c3c/refs/heads/master/resources/project_schema.json"
}
```

Ideally, it would be accessible from a nicer-looking URL, like this:
```json
{
  "$schema": "https://c3-lang.org/resources/v0.5.x/project_schema.json"
}
```

### DIFF
--- a/resources/project_schema.json
+++ b/resources/project_schema.json
@@ -1,0 +1,246 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "Project Configuration Schema for C3 project.json",
+  "description": "Schema for project.json configuration file based on C3 documentation",
+  "type": "object",
+  "properties": {
+    "langrev": {
+      "type": "string",
+      "description": "Language version of C3.",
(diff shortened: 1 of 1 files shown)


lerno:
Just a short note here. I'm preparing the 0.7.0 release and try to just do bug fixes. This will be approved for 0.7.1 after 0.7.0 is released.

lerno:
I see that some fields are missing, but let me merge this as is.

Issue #1566, opened 2024-10-19
Reconsider integer type names


nacaclanga:
The current names for integer types have the followring shortcomings:

1. The names of the suffices do not match up with the names of the types. Uses would have to remember both sets indipendently.
2. The `long` type in C does no necessarily have the same size as the `long` type in C3. E.g. on Windows C's long is 32 bit while C3's `long` is 64 bit. This could potentially create confusion when translating APIs. (Some other integer types have the same issue.). In contrast C has standartized type aliases in `<stdint.h>`, but these follow a systematic `[u]intXX_t` naming sheme.
3. If a programmer want's to define type aliases that match the C FFI names these names are allready taken.
4. The name `char` could also be confusing.

For this reason, one should think about whether it wouldn't be better to rename basic integer types to align more closely with their suffix names.

lerno:
Please have a look at #1178

And please note that the exact (OS and Arch dependent) C types are available as CInt, CShort, CChar and so on. 

In regards to the suffixes: it's possible that the bit sized suffixes disappear.

lerno:
Can I close this?

HMart81:
IMO this should stay open, because personally I think the only option we have, to create type alises, is not something I like, and all because of the hard requirement to start them with upper case, instead of me being able to use s32, I need to use S32 that IMO is visually ugly and requires more fingers and keys by being forced to press shift, because of this, I would love if this was something internal to the language, something like this: 

def byte = char;
def s8    = ichar;
def u8   = byte;
def s16  = ishort;
def u16  = short;
def s32  = int;
def u32  = uint;
def s64  = ilong;
def u64  = long;
def f32  = float;
def f64  = double;

But I'm forced to do this now... :(

def Byte = char;
def S8    = ichar;
def U8   = byte;
def S16  = ishort;
def U16  = short;
def S32  = int;
def U32  = uint;
def S64  = ilong;
def U64  = long;
def F32  = float;
def F64  = double;

HMart81:
IMO this should still be considered.  Being able to alias specially in this particular language rules... is not good imo...

What I wanted alias to be:

def byte = char;
def s8    = ichar;
def u8   = byte;
def s32  = int;
def u32  = uint;
def u64  = ulong;
def f32  = float;
def f64  = double;

What I needed to do for example for the alias to compile... :(

def Uint8   = char;
def Uint16  = ushort;
def Sint32  = int;
def Uint32  = uint;
def Sint64  = long;
[...]

Issue #1178, opened 2024-04-07
Discussion: Integer Names


wrenger:
Hi,

First of all, thanks for this cool language. 
I really do like many of its design decisions. Only one seems odd to me and that are C3's integer names. I think C's integer naming is already difficult (due to its history) and unfortunately C3 is not that much better. 

Fortunately you decided for fixed sizes independent from the target architecture but char/short/int/long are only familiar to C devs, which might become confused if these have different sizes than their C counterparts. Names like u8/u16/u32/u64/u128 (and i8/...) are much more meaningful and accurate, and give you the option to support C's char/short/int/long for FFI (although CInt is also fine). And int128 fits a lot better in this naming scheme.

Lastly the isz/usz and iptr/uptr: Again isize/usize is easier and can be said out loud, which is a plus.
Also they are a bit confusing. Are there architectures nowadays where isz and iptr differ? Maybe one architecture dependent integer type (or two with unsigned) is enough?

lerno:
Fixing the sizes while using C naming is not uncommon: see C#/Java for example. The downside of i8 etc is one of readability, and also would mean that code would look significantly different from the same C sample. 

```c
// Compare:
for (i8 i = 8; i >= 0; i--) { ... }
// and
for (int i = 8; i >= 0; i--) { ... }
```

Originally I used `i128` for the `int128` type, but I felt that I had to change it to `int128` to increase readability. And if we replace `int` with `int32`, `long` with `int64`, I don't feel much is won.

Regarding `isize` and `usize`:

Originally C3 has `isize`/`usize` and `iptrdiff`/`uptrdiff`. There were one type too many, as they for almost every use would be exactly the same size. However, dropping `iptrdiff` felt unfortunate. This: `&a - &b` seems hardly to be an "isize". The name "isize" is too explicitly a size, and felt incorrect to use. On the other hand, just having `usize`/`iptrdiff` seemed equally odd: not only would it violate the convention, but sometimes something like a signed `usize` is useful, and then `iptrdiff` seemed a bad thing to use.

For that reason the contraction `usz` and `isz` was chosen. They are shorter, but also more abstract, which means "isz" feels more natural to encompass both "isize" and "iptrdiff" usages, without nudging too much towards the "isize" meaning.

As for iptr sizes: Rust has been struggling with it's decision to say usize = uintptr (see https://internals.rust-lang.org/t/pre-rfc-usize-is-not-size-t/15369/28 etc) which seems to hint that conflating them is not a great idea.

[...]

PR #2072, opened 2025-04-03
Allow alternative more cohesive primitive type naming scheme

andmoos:
In my C code bases I heavily use a custom cohesive primitive type naming scheme that I find much more readable. In C3 this would end up like extra primitive types:

| Alias Type | C3 Type  |
|------------|----------|
| b8  	      | bool     |
| u8  	      | char     |
| u16 	      | ushort   |
| u32 	      | uint     |
| u64 	      | ulong    |
| u128       | uint128  |
| s8  	      | ichar    |
| s16 	      | short    |
| s32        | int      |
| s64        | long     |
| s128       | int128   |
| f16        | float16  |
| f16b       | bfloat   |
| f32 	      | float    |
| f64 	      | double   |
| f128       | float128 |

Unfortunately one of the very first things to discover in C3 was that this naming scheme cannot be implemented with `typedef` or `alias` due to the type naming rules. (Or did I miss a possible solution?)

As this is important to me enough to see what can be done about it, I altered the language parser to support these additional types directly there. Assuming that the type naming rules are "harder" to change due to their justification for use in parsing rules...

Don't get me wrong I do want to provide this (rather big) change to the language here mostly for discussion. Hopefully to open the possibility of having '**a**' solution for C3 allowing this one way or another in the end.

PS: A corresponding change for the c3intellij codebase can be found in this [fork](https://github.com/andmoos/c3intellij/tree/cohesiveprimitivetypes).

### DIFF
--- a/resources/grammar/c3.l
+++ b/resources/grammar/c3.l
@@ -177,6 +177,25 @@ typedef struct {
 "void"			{ return(VOID); }
 "while"			{ return(WHILE); }
 
+// BEGIN: alternative cohesive primitive type naming
+"b8"			{ return(BOOL); }
+"u8"			{ return(CHAR); }
+"u16"			{ return(USHORT); }
+"u32"			{ return(UINT); }
+"u64"			{ return(ULONG); }
+"u128"			{ return(UINT128); }
+"s8"			{ return(ICHAR); }
+"s16"			{ return(SHORT); }
+"s32"			{ return(INT); }
+"s64"			{ return(LONG); }
+"s128"			{ return(INT128); }
+"f16"			{ return(FLOAT16); }
+"f16b"			{ return(BFLOAT16); }
+"f32"			{ return(FLOAT); }
+"f64"			{ return(DOUBLE); }
+"f128"			{ return(FLOAT128); }
+// END: alternative cohesive primitive type naming
+
 @{CONST}        { return(AT_CONST_IDENT); }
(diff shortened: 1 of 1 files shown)


lerno:
As you see, the problem is not with implementing an alternative naming scheme. It is straightforward. The problem is what this means in practice: it will mean that every project will then have to decide what naming scheme to go with, and there will style guides either picking one or the other.

Just because the author of the language couldn't make up his mind.

The various schemes with bit size naming in C is mainly because this is the only possible option with `int` `long` etc already in use.

I appreciate seeing `s32` at least. Unlike the more popular `i32`, it doesn't mess up the readability of for loops. (Compare `for (i32 i = 32; i >= 0; i--)` vs `for (s32 i = 32; i >= 0; i--)` vs `for (int i = 32; i >= 0; i--)`, I find the i32 variant very hard to visually parse)

Previous issues relating to the same thing: #1566 #1178 #434

From what I can tell arguments boil down to the following so far:

1. "It's more modern"
2. Personal preference
3. "People won't understand int is 32 bits"
4. "I forget that int is 32 bits"
5. "I am used to another scheme from Rust"

All of them seem fairly subjective to me.

And while the sXX/uXX/fXX naming scheme is straightforward, we already know it breaks down for bfloats. It has its charms, but it would also make the code look less like C.

I weigh the latter as more important given that all arguments for either seem to be subjective.

PR #2081, opened 2025-04-12
Type mismatch fix

AlexCodesApps:
Fixed type mismatch in static function 'match_argopt' in file 'src/build/build_options.c',
where false was returned from the function which has a return type of 'const char *'.
I will note that clangd seems to be running c23 despite the compilation_commands.json I used,
allowing me to catch this (bug?) easily.


lerno:
Ah, super. It's too easy to get that broken result.

PR #2082, opened 2025-04-13
Fix build under Ubuntu with GCC 13.3.0

SRaimondi:
The build from source was failing in Ubuntu using GCC 13.3.0 because of the following error

Screenshot from 2025-04-13 19-16-20

Initializing  `bytes_type` to `NULL` as the other variables fixes the issue.


lerno:
Kind of incorrect analysis of it I think, but better add it.

lerno:
Thank you.

PR #2090, opened 2025-04-16
Add `default-visibility` compile option.

bokic:


lerno:
I don't understand. What is this for?

bokic:
for ex,
When building shared libraries you might only want explicitly marked functions to be exported, not all by default.
in C the solution is to add `visibilty=hidden`. Which will result in reduced number of exported functions.

And, when building executable functions should not to be exported by default.

And later at the linker phaze, linker will detect functions that are not exported and not used, so, will not put them into the binary, resulting smaller binary size.

lerno:
But that doesn't make sense for C3, since there is `@export` and `@public`. Only `@export` will be exported properly.

lerno:
So again I'm not sure here why this is needed.

lerno:
Ping

lerno:
I'm sorry but I need to close this due to inactivity

PR #2091, opened 2025-04-16
Adding array contains convenience macro

joshring:


lerno:
Can you add a test and fix the formatting to be consistent? `if(array.len` -> `if (array.len`

Simple guard statements can be placed on a single line:
```c
	if(array.len == 0)
	{
		return false;
	}
```
=>
```c
if (!array.len) return;
```
Another thing is that `foreach` will be really efficient, so if the array is empty you don't need to check the length. So the whole could be:
```c
foreach(&item : array)
{
    if (*item == element) return true;
}
return false;
```

data-man:
```rust
@require @typekind(array) == SLICE || @typekind(array) == ARRAY
```

What about `@require @typekind(array) == SLICE || @typekind(array) == ARRAY || @typekind(array) == VECTOR`?

joshring:
It's a great question, I was borrowing the general rules from the code in that module, will defer to @lerno on this one, there's some conversions which can happen with vectors and arrays also so I am not sure how that would work.

lerno:
I can add vector after merging this.

PR #2103, opened 2025-04-21
Improved CMake integration of LLVM

hyperpastel:
Added the paths for the CMake config files for LLVM 19 and higher to ``CMAKE_PREFIX_PATH``, so that they will also be searched automatically.


lerno:
I had one problem: if I passed in 
```
-DLLVM_DIR=/opt/homebrew/Cellar/lld@19/19.1.7/lib/cmake/llvm -DLLD_DIR=/opt/homebrew/Cellar/lld@19/19.1.7/lib/cmake/lld
```
It would ignore those settings and detect 20.0.0git instead

hyperpastel:
Yes. Find_package will, according to it's documentation, visit ``CMAKE_PREFIX_PATH`` before ``<Package>_DIR``. In this instance, ``20.0.0git`` is visited first, found, thereby effectively ignoring ``LLVM_DIR``. 

Perhaps instead of appending the glob to the ``CMAKE_PREFIX_PATH``, it should be appended to ``LLVM_DIR``, so a user provided location will be checked first and then followed by the ``llvm19``+ locations?

lerno:
Yes I think so. Otherwise it will be hard to test against multiple LLVM versions.

hyperpastel:
I've made some changes!

Since there can only be one ``LLVM_DIR``, I've now set it to check for that setting and respect that with highest priority, and only if this is not set will it consider the different locations. 

This has a side effect of failing to find LLVM if ``LLVM_DIR`` points to an invalid location and the build files will need to be regenerated, however I believe this is better than giving an invalid location and CMake then just picking something else without telling you.

lerno:
Thanks, now everything works! Great job

PR #2111, opened 2025-04-25
Rework Win32 mutex, condition variable and once flag

cbuttner:
Motivation for this PR:
- `ConditionVariable` was broken
- Usage of heavyweight `HANDLE`-based synchronization primitives

Changes
- Use more lightweight and faster inbuilt Win32 synchronization primitives, also used by the MSVC STL
- `NativeMutex` is now 16 instead of 48 bytes
- `TimedMutex` and `TimedRecursiveMutex` can no longer be used with a `ConditionVariable`, at least for now

This needs another pair of eyes to look for mistakes, but I've tested this in practice and couldn't find any bugs so far.


lerno:
Let's see if we can find someone on discord to test it out

test/unit/stdlib/threads/simple_thread.c3
```diff
@@ -69,7 +69,10 @@ fn void testrun_mutex_timeout() @test
 		unreachable("lock_timeout should fail");
 	}
 	m.unlock()!!;
-	m.lock_timeout(20)!!;
+	if (catch err = m.lock_timeout(20))
+	{
+		assert(err == thread::WAIT_TIMEOUT);
+	}
```
lerno, line 75: Is this correct? It looks wrong.

lib/std/threads/os/thread_win32.c3
```diff
@@ -5,239 +5,315 @@ typedef NativeThread = inline Win32_HANDLE;
...
+}
+
+<*
+ @require mtx.initialized : "Mutex was not initialized"
+ @ensure !mtx.initialized
+*>
+fn void? NativeMutex.destroy(&mtx)
+{
+	mtx.initialized = false;
+	$if env::COMPILER_SAFE_MODE:
```
lerno, line 61: This code could potentially be moved into the `@require` instead? So that it is a bug to not call it on its own thread.

lib/std/threads/os/thread_win32.c3
```diff
@@ -5,239 +5,315 @@ typedef NativeThread = inline Win32_HANDLE;
...
+	if (signal) win32::wakeConditionVariable(&mtx.cond_var);
+}
+
+fn void? NativeConditionVariable.init(&cond)
+{
+	cond.cond_var = {};
 }
 
 fn void? NativeConditionVariable.destroy(&cond) @maydiscard
 {
```
lerno, line 271: Could something be checked here?

lerno (changes requested):

lerno:
Could you also add something in the release notes, like:
### Features
- New Win32 mutex implementation

### Fixes
- ConditionVariable now properly works on Win32

lerno:
Other than the above it looks reasonable.

lerno:
Do you think you'll have time to fix this before 0.7.1 can be released?

cbuttner:
I've been wondering if the mutex functions should really return errors or just assert in safe mode for
- Locking a non-recursive mutex twice (deadlock)
- Unlock by thread not owning the lock
- Destroying a locked mutex

lerno:
I think the more we eventually move to contracts the better.

PR #2112, opened 2025-04-26
Added I18n library to std

tahadostifam:
Here is an uncompleted implementation of `i18n` that loads json files with language code and translates it. Format string works fine.

I know that it's not complete, But I opened PR to be sure my code is fine until this point.

I also consider to add following features:

- [x] Load from JSON
- [x] Formatting
- [ ] Load from `YAML`
- [ ] Pluralization Support
- [ ] Load `TranslationMap` from single file

Would you share more ideas with me if possible?

### DIFF
--- a/lib/std/text/i18n.c3
+++ b/lib/std/text/i18n.c3
@@ -5,10 +5,12 @@ import std::os;
 import std::core::mem::allocator;
 import std::collections::map;
 import std::collections::object;
+import std::collections::list;
 import std::encoding::json;
 import std::time;
 
 alias LanguageMap = HashMap{String, Object*};
+alias LoadedFilesList = List{Object*};
 alias TranslationMap = HashMap{String, LanguageMap};
 
 // This fault raised when translation key not found.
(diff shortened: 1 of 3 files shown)


lerno:
@tahadostifam I've adopted a new way to contribute to the stdlib: incubating the feature first.

https://github.com/c3lang/c3c/blob/master/CONTRIBUTING.md

So basically, create a repo for it, submit that finished repo for inclusion in the stdlib. And call for suggestion and improvements on that repo by posting in https://github.com/c3lang/c3c/discussions and on the Discord.

Ideally we use it a bit to iterate on it before it's included in the stdlib. This is the new way going forward and will apply to all additions (including my own!)

tahadostifam:
I've already created a repository for it :

https://github.com/tahadostifam/i18n-c3

I'll share it on Discord to have more discussions about it.

Thank you so much.

lerno:
Make a pull request on c3-showcase so it can be added there as well. That would make it easy to find.

PR #2119, opened 2025-05-03
Support for Homebrew LLD

lucianferoiu:
If people install LLVM et.al. using Homebrew, chances are that the libs are in the /opt/homebrew/lib (by default).
Let's include them if they exist.


lerno:
Thank you

PR #2137, opened 2025-05-16
Add `Maybe.equals` when inner type is equatable.

gcoakes:


lib/std/collections/maybe.c3
```diff
@@ -1,5 +1,6 @@
 module std::collections::maybe{Type};
 import std::io;
+import std::core::types;
```
lerno, line 3: Core is implemented implicitly, so it doesn't need to be included

lib/std/collections/maybe.c3
```diff
@@ -34,3 +35,11 @@ macro Type? Maybe.get(self)
 {
 	return self.has_value ? self.value : NOT_FOUND?;
 }
+
+fn bool Maybe.equals(self, Maybe other) @operator(==) @if(types::is_equatable_type(Type))
+{
+	if (self.has_value) {
```
lerno, line 41: Please follow the brace code standard here 😅

lerno (changes requested):
Could you also add a note to the changelist for this?

lib/std/collections/maybe.c3
gcoakes: Done.

Would it make sense to make that a compile warning in the future?

lib/std/collections/maybe.c3
gcoakes: Whoops. Done.

gcoakes:
> Could you also add a note to the changelist for this?

Done. Please let me know if there is anything else.

lerno:
Thank you!

Issue #2012, opened 2025-03-02
Feature request: `inherit_stdio` option inside `std::os::process::SubProcessOptions`


angelcaru:
Always creating a set of pipes doesn't seem like the best idea

lerno:
If it's easily achievable? Feel free to make a PR with the enhanced behaviour (cross platform needed)

chrischtel:
Hello @angelcaru @lerno 
I've came up with a solution. You can check it out on PR #2138

lerno:
This should be implemented with #2138 , can you check @angelcaru

PR #2138, opened 2025-05-16
Feature: Add `inherit_stdio` Option for SubProcess

chrischtel:
Related issue #2012 

This PR adds enhanced functionality SubProcesses. It adds an option `inherit_stdio` to `SubProcessOptions` that allows child processes to inherit the parent's stdin, stdout, and stderr handles instead of creating new pipes.

**Details**
- Added `inherit_stdio` boolean flag to `SubProcessOptions` bitstruct
- Modified `create()` function for both Windows and POSIX platforms to handle the flag

> - For **Windows**: Skip setting `STARTF_USESTDHANDLES` flag when inheriting stdio

>  - For **POSIX**: Skip pipe creation and file actions when inheriting stdio
> Updated `read_stdout()`, `read_stderr()`, `stdout()` and `stderr()` methods to handle null file handles

**Added tests that verify:**

1. When using `inherit_stdio`, stdout/stderr file handles are null
2. Reading from stdout/stderr with `inherit_stdio` returns 0 bytes
3. Processes still complete correctly with `inherit_stdio`
4. Direct comparison between normal operation and `inherit_stdio`

### DIFF
--- a/lib/std/os/subprocess.c3
+++ b/lib/std/os/subprocess.c3
@@ -45,6 +45,8 @@ bitstruct SubProcessOptions : int
 	// Note: this will **not** search for paths in any provided custom environment
 	// and instead uses the PATH of the spawning process.
 	bool search_user_path;
+    // Inherit the parent's stdin, stdout, and stderr handles instead of creating pipes
+    bool inherit_stdio;
 }
 
 fn void? create_named_pipe_helper(void** rd, void **wr) @local @if(env::WIN32)
--- a/lib/std/os/subprocess.c3
+++ b/lib/std/os/subprocess.c3
@@ -117,126 +117,150 @@ fn WString convert_command_line_win32(String[] command_line) @inline @if(env::WI
 *>
 fn SubProcess? create(String[] command_line, SubProcessOptions options = {}, String[] environment = {}) @if(env::WIN32)
 {
-	void* rd, wr;
-	Win32_DWORD flags = win32::CREATE_UNICODE_ENVIRONMENT;
-	Win32_PROCESS_INFORMATION process_info;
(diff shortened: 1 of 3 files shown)


lib/std/os/subprocess.c3
```diff
@@ -115,126 +117,150 @@ fn WString convert_command_line_win32(String[] command_line) @inline @if(env::WI
...
+    CFile stdout;
+    CFile stderr;
+    void* rd = null;
+    void* wr = null;
+    
+    // Only create pipes if not inheriting stdio
+    if (!options.inherit_stdio)
+    {
+        if (!win32::createPipe(&rd, &wr, &sa_attr, 0)) return FAILED_TO_CREATE_PIPE?;
+        // TODO defer catch
```
lerno, line 142: What is this TODO?

lerno (changes requested):
I can't test this on windows I'm afraid, so I'll have to take your word for it! 😄 Can you add a note in the releasenotes as well?

chrischtel:
Hello, thanks for reaching out to me. I will do some more in depth testing and if everything still works fine I'll add a word to the release notes. 
Thanks for your trust.

Christian

lib/std/os/subprocess.c3
chrischtel: This already existed so it comes not from me (you can check on the file diff and on the master branch where it also exist.

chrischtel: I guess "it" wants us to defer catch the possible failure on the `create_pipe`.

chrischtel:
Thank you! I am looking forward to contribute more to C3

cbuttner:
Guess I'm too late, but this would've been improved by indenting with tabs, the diff is not very readable and doesn't follow code style for the stdlib.

PR #2161, opened 2025-05-27
Fix #1718: Add --header-output option to specify header file directory

### LINKED ISSUE #1718: Can't set the path of generated C header (while compiling static or dynamic lib)
Steps to reproduce:
1. Create `project.json` with the following content:
```json
{
  "output" : "build",

  "targets" : {
    "mylib" : {
      "type" : "static-lib",
      "sources" : [ "main.c3" ],
    },
  },
}
```
2. Create `main.c3` with arbitrary content.
3. Run command `c3c build mylib`.

This command will produce a header `mylib.h` in root dir. Of course, it would be better to generate everything into `build`, but also it would be essential to have option like `--header-output`, which represents the emitted header path.

chrischtel:
## Summary
Fixes #1718 by adding a `--header-output <dir>` command line option that allows users to specify where generated C header files should be placed, instead of always outputting them to the project root directory.

## Changes
- `--header-output <dir>` to specify custom header directory
- Headers now go to output directory by default

- [ ] Add releasenotes entry once PR is reviewed


lerno:
Thank you!

PR #2166, opened 2025-05-29
[Feat] add `quiet` to project.json settings

BWindey:
This pull requests adds the `quiet` setting to projects, both at top-level and target-level.
It behaves the same as adding the `--quiet`/`-q` flag on the commandline.

Let me know if this needs any tests, and where about those should be written.


lerno:
This should work the other way around. The target options form the basis of the settings and the command line options override them. This means that if you have "quiet" set to true, but on the command line you specify -v then the -v has precedence. The target option only happens if there is no setting involving it on the command line.

BWindey:
Oh, yes. 
That should be an easy fix.

lerno:
Thank you!

PR #2171, opened 2025-05-30
Fix missing __errno_location() to __errno() on android

Yuuki1578:
Hello

I know this fix would break for env::LINUX, but is there any macro-ish for android? Like env::ANDROID or how do i add those macro?

### DIFF
--- a/lib/std/libc/os/errno.c3
+++ b/lib/std/libc/os/errno.c3
@@ -1,9 +1,10 @@
 module libc::os @if(env::LIBC);
 
-// Linux
-extern fn int* __errno_location() @if(env::LINUX);
-macro int errno() @if(env::LINUX) => *__errno_location();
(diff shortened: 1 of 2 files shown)


lib/std/libc/os/errno.c3
```diff
@@ -1,9 +1,10 @@
 module libc::os @if(env::LIBC);
 
-// Linux
-extern fn int* __errno_location() @if(env::LINUX);
-macro int errno() @if(env::LINUX) => *__errno_location();
-macro void errno_set(int err) @if(env::LINUX) => *(__errno_location()) = err;
+// Is there any env::ANDROID?
```
lerno, line 3: Yes, use `env::ANDROID`

lerno:
Should I close this one?

lerno:
Superseded by #2182

PR #2179, opened 2025-06-02
Fix #2121

### LINKED ISSUE #2121: [Feature Request] Specify run directory
I needed an executable that would load some files to accompany it. The problem is that the c3c run command runs the executable from your the project directory and not the build directory. Therefore the paths were different for the executable.

build/
├ executable
├ required_file.txt

Could we add an extra field to project.json where you could either specify where to run at or if you should run from build directory or project directory?

Currently i have to do
c3c compile
cd build
./run
cd ..

BWindey:
Fixes #2121 where specifying `"run-dir"` in project.json was not properly updated inside the compiler.

### DIFF
--- a/src/build/builder.c
+++ b/src/build/builder.c
@@ -492,6 +492,8 @@ static void update_build_target_from_options(BuildTarget *target, BuildOptions *
 		}
 		if (!target->script_dir) target->script_dir = "scripts";
 	}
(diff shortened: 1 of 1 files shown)


BWindey:
Apparently "fixes #..." also sets up github to auto-close the linked issue, but I don't think it should auto-close that issue as it is not an issue I created.

PR #2182, opened 2025-06-03
Add external __errno() function for env::ANDROID

Yuuki1578:
Using __errno() for Android 9 rather than __errno_location(), as defined in errno.h

```c
int* _Nonnull __errno(void) __attribute_const__;

#define errno (*__errno())
```


lerno:
Does this close the previous pull request?

lerno:
Ping

PR #2193, opened 2025-06-07
fix `types::may_load_atomic` with enums

Book-reader:
it could probably just return true directly, but this works as well


lerno:
This is correct, because an enum may use int128 backing which would not be atomic.

Issue #1873, opened 2025-01-21
Better prefix for $echo


tomaskallup:
Currently using `$echo` adds `] ` as a prefix, which is a bit confusing.

It might be better to use something like `$ ` or `c3c > `.

alexveden:
File name and line also could be useful, maybe.

BWindey:
File-name can already be done with `$$FILE` (https://c3-lang.org/misc-advanced/builtins/#compiler-builtin-functions).
Same for `$$LINE` and `$$LINE_RAW` (same url).

An example:
```c
module main;
import std::io;

fn int main() {
	$echo "Compiler says hello in file " +++ $$FILE +++ " on line ";
	$echo $$LINE;
	io::printn("Hello, World!");
	return 0;
}
```
Produces:
```
$ c3c compile test.c3
] Compiler says hello in file test.c3 on line
] 8
Program linked to executable 'main'.
```

So I would not include filename and line into `$echo`, as it is already possible to do so now. The user could very well write their own macro to include that info without having to explicitly type it:
```c
module main;

import std::io;

macro to_string($num) {
	char[] $res;
	int $i = 0;
	$for (;$num != 0; )
		$res = { (char) ('0' + $num % 10) } +++ $res;
		$num = $num / 10;
		$i++;
	$endfor
	return (String) $res;
}

macro compiler_message($msg) {
	$echo $msg +++ " in file " +++ $$FILE +++ " on line " +++ to_string($$LINE);
}

fn int main() {
	$echo "Compiler says hello in file " +++ $$FILE +++ " on line ";
	$echo $$LINE;
	io::printn("Hello, World!");
	compiler_message("Hello from compiler");
	return 0;
}
```
Correctly prints:
```
c3c compile test.c3
] Compiler says hello in file test.c3 on line
] 25
] Hello from compiler in file test.c3 on line 27
Program linked to executable 'main'.
```

Though I must admit that this `to_string` macro is a bit excessive, but I haven't found a shorter yet.

alexveden:
ok, makes sense. I like macros in C3, one can solve any problem with them :)

re `to_string($$LINE)`, does `$stringify($$LINE)` work?

tomaskallup:
> re `to_string($$LINE)`, does `$stringify($$LINE)` work?

Nope, that just prints `] $$LINE`

BWindey:
No, that does not work. That's why I opened #1874, Christoffer told me in Discord that it would be a good addition. 
[...]

PR #2206, opened 2025-06-13
Add an optional 'Prefix' parameter to $echo

NotsoanoNimus:
Addresses Issue #1873 and attempts to resolve the `$echo` "better prefix" discussion by letting developers choose their own.

Code written against older versions of C3 which do not support prefixes will not be affected/broken because the prefix is the second (optional) param, i.e. $echo becomes `$echo MESSAGE[, PREFIX] ;`. Thus, messages without a `PREFIX` will still output the typical `] ` prefix.

The only notable constraint at the time of submission is that surrounding the two constant expressions with parentheses (_see below_) will not compile properly.

```c
module main;

import std::io;


macro beacon($msg)
{
    $echo $msg, @sprintf("[%s:%s]>  ", $$FILE, $$LINE);
}


fn void main()
{
    $echo "Old-school, backwards-compatible";
    $echo("simple, still works");

    $echo "example,test", "PFX>>> ";
    // DOES NOT WORK:  $echo("parens", "<<< WITH >>>  ");

    beacon("It's this location!");

    io::printn("Hello, World!");
}
```

```
$ ./build/c3c compile-run /tmp/tmp.YHrcPSCpdl/hw.c3
] Old-school, backwards-compatible
] simple, still works
PFX>>> example,test
[hw.c3:18]>  It's this location!
Program linked to executable 'hw'.
Launching ./hw
Hello, World!
```

### DIFF
--- a/src/compiler/compiler_internal.h
+++ b/src/compiler/compiler_internal.h
@@ -1417,11 +1417,18 @@ typedef struct
 
 typedef struct
 {
-	bool is_ensure;
-	ExprId message;
-	ExprId expr;
-	Expr **args;
+    bool is_ensure;
+    ExprId message;
+    ExprId expr;
+    Expr **args;
 } AstAssertStmt;
+
+typedef struct
+{
+    bool has_prefix;
+    Expr *prefix;
+    Expr *message;
+} AstEchoStmt;
 
 
 typedef struct
@@ -1479,6 +1486,7 @@ typedef struct Ast_
 		AstAsmStmt asm_stmt;
 		const char *asm_label;
 		AstAssertStmt assert_stmt;          // 16
+        AstEchoStmt echo_stmt;
 		AstCaseStmt case_stmt;              // 32
 		AstCompoundStmt compound_stmt;      // 12
 		AstId ct_compound_stmt;
--- a/src/compiler/copying.c
+++ b/src/compiler/copying.c
@@ -730,6 +730,15 @@ RETRY:
 			fixup_astid(c, &ast->defer_stmt.prev_defer);
 			break;
 		case AST_CT_ECHO_STMT:
+            copy_reg_ref(c, source, ast);
+            SCOPE_FIXUP_START
+                MACRO_COPY_EXPR(ast->echo_stmt.message);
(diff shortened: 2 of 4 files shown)


BWindey:
Not sure about what I think of this solution, but the parens not working is because the `$echo` never needs parens anyway, just like the `$for` and `$if`, the parens are only used for operator precedence.

So something like `$echo ("Hello"), ("prefix> ");` works, I think.

NotsoanoNimus:
@BWindey Thanks for the feedback, and makes sense about the parentheses.

```c
macro @echo($message)
{
    $echo $message, "$echo: ";
}

// ...
    @echo(@sprintf("Test message on line %s", $$LINE));
```
```
$echo: Test message on line 12
```

Something like this for those who want to customize a prefix is an easy solution, without settling on one option. It only adds a slight overhead to the builtin but resolves the "what should it be" problem for everyone who cares. Those who don't care can just keep `$echo`ing as they do now. :smiley: 

Anything else is too complex IMHO.

lerno:
An alternative is to make it configurable as a build parameter: `--echo-prefix myecho`

NotsoanoNimus:
> An alternative is to make it configurable as a build parameter: `--echo-prefix myecho`

I think that's a great idea. Would this still work if someone wanted to use a prefix that includes tokens like `$$FILE`?

lerno:
It could I guess, although that is more work.

NotsoanoNimus:
> It could I guess, although that is more work.

I suppose the original proposal is perhaps a more flexible option (whatever that gains C3).

But I'm willing to help with an implementation for either option/both; as no matter what is chosen, _something_ extra on top of `$echo` would need to be configured. Whether that's a cmdline param or an extra macro.

lerno:
I added a simple --echo-prefix build option now. It can definitely be improved on. Currently it supports {FILE} and {LINE}

NotsoanoNimus:
> I added a simple --echo-prefix build option now. It can definitely be improved on. Currently it supports {FILE} and {LINE}

Nice! We can close this in favor of your changes if the branch should be axed. Thank you.

lerno:
Thank you!

PR #2225, opened 2025-06-18
Add build instructions for Arch Linux

tanisperez:
Building `c3c` on Arch Linux can be tricky due to its rolling release model and frequent updates to system packages. I ran into several issues while trying to compile with the latest versions of `clang` and `llvm`, so I’m adding a small section to the README with step-by-step instructions tailored for Arch users.

**Note**: I had to add the flag `CMAKE_C_FLAGS_RELEASE="-Wno-error=maybe-uninitialized"` to avoid a compilation error caused by Clang 20.1.6 treating certain warnings as errors.

Hopefully, these instructions will save others some time and frustration in the future.


lerno:
Thank you!

PR #2227, opened 2025-06-19
Add SHA512 Module to stdlib

NotsoanoNimus:
See the (passing) :white_check_mark: unit tests file for this new module.

To use it just `import std::hash::sha512` and cook.

I've added some base information for a near-future expansion of the module to include the truncated hash versions of SHA512 as well (i.e., SHA384, SHA512/224, & SHA512/256).

I also plan to later adapt the SHA256 module to include SHA224, that way C3 will have the full SHA hashing suite available.

=====
I have not benchmarked this, but tried to keep it clean and optimized where I could.


lerno:
Thank you!

PR #2230, opened 2025-06-19
Fix error maybe-uninitialized

tanisperez:
On Arch Linux, the current Clang version is 20.1.6, which triggers a warning treated as an error in `src/compiler/llvm_codegen_expr.c`. Specifically, the variable `is_ptr` may be read before it is initialized when the `is_integer` branch is taken.

To address this, the initialization logic for `is_ptr` has been adjusted to ensure it's always set before use, even in branches where it's not explicitly assigned beforehand.

This change makes the code compatible with Clang 20.1.6, avoiding future compilation failures on other systems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -379,8 +379,7 @@ The c3c binary should be created in the build directory. You can try it out by r
 ```bash
 cmake -B build \
     -D C3_LINK_DYNAMIC=ON \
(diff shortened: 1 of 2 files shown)


lerno:
Thank you!

PR #2231, opened 2025-06-20
Implement import aliasing

bagoette:
## Motivation

When working with `C3` modules, I found myself wanting to use shorter or more descriptive names for imported modules.
This feature request adds Python/TypeScript-style import aliasing to `C3`, allowing developers to import modules with custom aliases.

## Changes

This PR implements `import as` syntax that allows aliasing imported modules:

```cpp
import std::core:dstring as sb;
import std::io as c;

fn void main() 
{
    DString builder = sb::temp_with_capacity(32);
    defer builder.free();
    builder.append("Hello, C3!")
    
    c::printfn(builder.str_view());
}
```

### Key Implementation Details

- **New `TOKEN_AS` keyword**: Added to lexer and parser infrastructure
- **Extended `ImportDecl` & `Module_` structure**: Added `alias` field to store the alias name
- **Enhanced import parsing**: Modified `parse_import()` to handle optional `as alias_name` syntax
- **Updated name resolution**: Added `sema_find_decl_by_alias()` to resolve symbols through aliases
- **Error handling**: Validates alias names and prevents reserved keywords

### Features Supported

- ✅ Basic aliasing: `import std::io as console;`
- ✅ List imports with aliases: `import std::ascii as a, std::io as foo;`
- ✅ Mixed aliased and non-aliased imports `import std::ascii, std::io as foo;`
- ✅ Error messages for invalid alias names
- ✅ Prevention of reserved keyword usage as aliases

## Testing

Added tests covering:
- ✅ Basic functionality (`import_as_works.c3t`)
- ✅ List imports (`import_as_list_works.c3t`) 
- ✅ Error cases for invalid identifiers (`import_as_error_ident.c3t`)
- ✅ Error cases for reserved keywords (`import_as_error_keyword.c3t`)
- ✅ Edge cases and boundary conditions

## Implementation Notes

- **Minimal invasive changes**: Only touched necessary files for clean integration
- **Backwards compatible**: Existing import syntax continues to work unchanged

## Concerns

- Seems like `C3` is allowing redundant module names and is handling ambiguous declarations so I didn't add a check for redundant import aliases


lerno:
So, as you've already heard, this is not a thing to be included as it is.

Generally the path shortening in C3 is intended to automatically create aliases for you, rather than having to come up with (inconsistent) aliases (naming is the hardest thing in programming and all of those things). This is why you consistently see ` io::printn("Hello, world")` rather than `std::io::printn("Hello, world")`

By having this scheme, I also try to encourage library writers to think about their prefixes, to already keep them short and sweet to begin with. Aliasing allows people to get away with much worse module names.

*That said* the lack of aliasing for modules is kind of a natural thing to add and a missing feature in the cases the library writer was an idiot when doing naming.

So I was thinking of adding something like `alias foo = module bar::baz;`, but there was no real consensus on the syntax.

The discussion was here: https://discord.com/channels/650345951868747808/820761627967488040/1374730243092381769

lerno:
So I'm going to close this, but is positive to adding an `alias some_path = module foo::bar` or something similar. The exact syntax should be worked out, so if you open an issue for it, we'll discuss this there. Keep this code around though, you can probably reuse part of it.

PR #2235, opened 2025-06-23
Tuple update

EroMrinin134:
Renamed to Pair & Triplet + added a Quartet.


EroMrinin134:
Oh, there is a couple of tests based of tuple...

lerno:
This is not allowed. You cannot rename types unless it's in a 0.x.0 release. Instead you need to deprecate the old type and add the new. As you see, there are also previous uses of it elsewhere, they need to be updated as well.

lerno:
Also, it's not been accepted yet. So hold on

PR #2238, opened 2025-06-23
Add 'strdup' reference to libc

NotsoanoNimus:
Noticed that the quick-and-easy `strdup` function was missing from `libc::` imports while working on a library. Adding it.


lerno:
Thank you

Issue #2195, opened 2025-06-08
Add functions to create escaped and unescaped strings


lerno:
There is functionality for urlencoding, but a common thing to need is to create a standard string within `""` and conversely, parse such a string. Some of this functionality is already in the JSON parser, so that could possibly be extracted.

lerno:
Implemented

PR #2243, opened 2025-06-27
Add string escaping and unescaping functionality

sudison:
Refer to issue: https://github.com/c3lang/c3c/issues/2195
Most of the code are generated by AI

- Add std::core::string_escape module with comprehensive string escape/unescape functions
- Support standard C-style escape sequences: \n, \t, \r, ", \, \b, \f, \v, \0
- Support hex escapes (\xHH) and Unicode escapes (\uHHHH, \UHHHHHHHH)
- Provide both allocating and temp allocator versions of functions
- Include robust error handling with specific fault types
- Add comprehensive unit tests with 6 test cases covering all functionality
- Extracted and generalized from existing JSON parser string handling
- All tests passing with proper C3 slice syntax (inclusive end index)


lib/std/core/string_escape.c3
```diff
@@ -0,0 +1,274 @@
...
+                if (c >= 32 && c <= 126)
+                {
+                    // Printable ASCII
+                    result.append_char(c);
+                }
+                else
+                {
+                    // Non-printable, use hex escape
+                    result.appendf("\\x%02x", (uint)c);
+                }
```
BWindey: You can write "\\\"" as '\`"`' (so use the backtick instead of quote).
See the docs: https://c3-lang.org/language-overview/types/#string-literals-and-raw-strings

lib/std/core/string_escape.c3
```diff
@@ -0,0 +1,274 @@
...
+
+ @param allocator : "The allocator to use for the result"
+ @param s : "The string to escape"
+ @return "The escaped string with surrounding quotes"
+*>
+fn String escape(Allocator allocator, String s)
+{
+    // Calculate the maximum possible length (worst case: every char becomes \xHH)
+    usz max_len = s.len * 4 + 2; // +2 for quotes
+    DString result = dstring::new_with_capacity(allocator, max_len);
```
BWindey: How much do other implementations (for other languages) allocate? 
It feels like assuming the string will quadruple in size is a bit much, and `DString` can dynamically scale if needed. I think that a typical usecase will only need 1x to 1.2x the amount of bytes?

lib/std/core/string_escape.c3
BWindey:
```c3
case '"' : result.append(`\"`);
case '\\': result.append(`\`);
case '\b': result.append(`\b`);
case '\f': result.append(`\f`);
case '\n': result.append(`\n`);
case '\r': result.append(`\r`);
case '\t': result.append(`\t`);
case '\v': result.append(`\v`);
case '\0': result.append(`\0`);
```

lib/std/core/string_escape.c3
```diff
@@ -0,0 +1,274 @@
...
+        }
+    }
+    return len;
+}
+
+<*
+ Unescape a quoted string by parsing escape sequences.
+
+ @param allocator : "The allocator to use for the result"
+ @param s : "The quoted string to unescape (must start and end with quotes)"
```
BWindey: The "must start and end with quotes" can be enforced using contracts:
```cpp
<*
	@require s[0] == '"' && s[^1] == '"' : "String must start and end with quotes"
*>
```
While this approach does not give that nice fault, it can already be checked at compile-time if the string passed to `unescape` is known at compile-time.

Maybe @lerno has a different opinion though.

lib/std/core/string_escape.c3
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    // Handle empty string case
+    if (s.len == 2)
+    {
+        return "";
+    }
+
+    // Remove surrounding quotes (end index is inclusive)
+    String content = s[1..s.len-2];
```
BWindey: `^2` is a shorthand for `s.len-2`, it reads better:
```cpp
String content = s[1..^2];
```

lib/std/core/string_escape.c3
```diff
@@ -0,0 +1,274 @@
...
+            case 'v':  result.append_char('\v');
+            case '0':  result.append_char('\0');
+            case 'x':
+                // Hex escape \xHH
+                if (i + 2 >= content.len) return INVALID_HEX_ESCAPE?;
+                char h1 = content[++i];
+                char h2 = content[++i];
+                if (!h1.is_xdigit() || !h2.is_xdigit()) return INVALID_HEX_ESCAPE?;
+                uint val = (h1 > '9' ? (h1 | 32) - 'a' + 10 : h1 - '0') << 4;
+                val += (h2 > '9' ? (h2 | 32) - 'a' + 10 : h2 - '0');
```
BWindey: Here the brackets are not needed as you're not shifting it.
If I were to write this, I'd do
```cpp
uint val = ...;
val = val << 4;
val += ...;
```
But that's my opinion.

BWindey (commented):
I don't have any power over contributions, but have a few remarks about the code. Nothing major though.

lib/std/core/string_escape.c3
BWindey: Can be found here at the end of that paragraph: https://c3-lang.org/language-common/arrays/#fixed-size-1d-arrays

lib/std/core/string_escape.c3
sudison: Fixed, use 1.2x space allocation.

lib/std/core/string_escape.c3
sudison: fixed

lib/std/core/string_escape.c3
sudison: fixed

lib/std/core/string_escape.c3
sudison: fixed

lib/std/core/string_escape.c3
sudison: I'd prefer runtime check, as the content of string may not be available during the compile time.

lerno:
Thank you. I changed it a little, so that the quoted / non quoted used the same function. I also made it a method on String, which will probably be a little more convenient.

sudison:
Woo, mission completed, my first commit to C3.

PR #2252, opened 2025-06-28
Fix typo in libc_allocator.c3

Book-reader:
Can you please fix this in the `cenum_branch` branch as well, since that's also affected (unless that branch is going to be merged into master soon)


lerno:
Thank you

PR #2253, opened 2025-06-28
Added colored error and warning compiler messages.

cubedium:
Makes the error and warning messages colored


Book-reader (commented):
It's probably a good idea to tie this in with the `--ansii` flag to allow it to be disabled

cubedium:
ok

cubedium:
done

lerno:
Does this work both with dark and light background?

cubedium:
i believe so

cubedium:
it does infact work with light backgrounds

lerno:
Thank you!

PR #2259, opened 2025-07-01
Add read_until_delimiter macro

sudison:
This macro reads from an input stream byte by byte until a specified delimiter is found, writing each byte to an output stream. It returns the number of bytes read (excluding the delimiter) or an error.

Key features:
- Reads byte-by-byte until delimiter is encountered
- Writes data to output stream as it reads
- Proper error handling for EOF and write failures
- Compile-time type checking with @require constraints

Potential use cases:
- Parse HTTP headers (read until '\r\n')
- Extract fields from CSV files (read until ',')
- Parse configuration files (read until '=')
- Process line-based protocols
- Extract tokens from streams

Added comprehensive tests covering:
- Basic functionality with delimiter found
- EOF handling when delimiter not found
- Empty input streams
- Immediate delimiter at start
- Output stream write failures

### DIFF
--- a/lib/std/io/stream.c3
+++ b/lib/std/io/stream.c3
@@ -165,6 +165,73 @@ fn usz? copy_to(InStream in, OutStream dst, char[] buffer = {})
 	$endswitch
 }
 
+
+<*
+ Read from input stream byte by byte until delimiter is found, writing each byte to output stream.
+
+ @param [&inout] in : "The input stream to read from"
+ @param [&inout] out : "The output stream to write to"
+ @param delimiter : "The delimiter (single character or string) to stop reading at"
+ @return "Number of bytes read from input stream (excluding delimiter), or error"
+ @require @is_instream(in)
+ @require @is_outstream(out)
+*>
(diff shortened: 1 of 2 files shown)


BWindey:
It might be worth adding `delimeter = '\n'` to `readline` in `io/io.c3` instead of a whole new function.

lerno:
Delimiter should be char[]. Temp allocation here is a bit unfortunate, and wrapping it in a `@pool` is not ideal. Consider `@stack_mem` for it.

lerno:
This will not work for the following example:

Data:
```
apapapbb
```

Delimiter:
```
apapbb
```
The code will not find the delimiter.

lerno:
Regarding the above, I need the implementation fixed if it is to be included.

lerno:
Just pinging you on this @sudison so you have a look at this.

lerno:
If there is no further work on this I'll close it without merging.

lerno:
Closing as per comment above

PR #2261, opened 2025-07-01
Add ansi.c3 for AnsiColor support

FoxKiana:


lib/std/core/ansi.c3
```diff
@@ -0,0 +1,95 @@
...
+
+<*
+ 8-bit color code
+
+ @require index >= 0 && index <= 255 : `value must be within range 0-255`
+ @return `the formatting char for the given background color`
+*>
+fn String color_8bit_fg(int index)
+{
+    return string::format(mem, "\u001B[38;5;%dm", index);
```
BWindey: There is no option provided to choose a different allocator, and the name does not imply that any heap allocations are happening. 
I'd maybe rename to `new_color_8bit_fg` and use `Allocator allocator = mem` as second argument.

lib/std/core/ansi.c3
```diff
@@ -0,0 +1,95 @@
+module std::core::ansi;
+
+import std::core::string;
+
+enum AnsiColor : const inline String
+{
+    RESET             = "\u001B[0m",
```
BWindey: Any reason to choose `\u001B` over `\e`?

lib/std/core/ansi.c3
```diff
@@ -0,0 +1,95 @@
...
+    BG_BRIGHT_BLUE    = "\u001B[104m",
+    BG_BRIGHT_MAGENTA = "\u001B[105m",
+    BG_BRIGHT_CYAN    = "\u001B[106m",
+    BG_BRIGHT_WHITE   = "\u001B[107m",
+}
+
+<*
+ 8-bit color code
+
+ @require index >= 0 && index <= 255 : `value must be within range 0-255`
```
BWindey: Could also make `index` a `char` instead.

BWindey (commented):
Aside from some practical code-nitpicks, I'd also consider merging the 8bit functions together like `color_8_bit(char index, bool fg = true)` (personal preference).

Also neat would be a comptime implementation using `+++` and `$stringify` to not need allocations. I think that'll be quite valuable as these are things that quite often happen at compile-time.
An example implementation:
```cpp
<*
	Returns a String that will turn the following text into the specified
	RGB colour if the terminal supports it.
*>
macro String @rgb(char #r, char #g, char #b, bool $bg = false) @const {
	var $s;
	$if $bg :
		$s = "48";
	$else
		$s = "38";
	$endif

	return "\e[" +++ $s +++ ";2;"
		+++ $stringify(#r) +++ ";"
		+++ $stringify(#g) +++ ";"
		+++ $stringify(#b) +++ "m";
}
```

Lastly, the enum-list is missing some entries that could be valuable as well, like:
```cpp
	DOUBLE_UNDER = "\e[21m",
	DEFAULT_NO_DIM = "\e[22m",
	DEFAULT_NO_ITALIC = "\e[23m",
	DEFAULT_NO_UNDER = "\e[24m",
	DEFAULT_NO_INVERT = "\e[27m",
	DEFAULT_NO_HIDDEN = "\e[28m",
	DEFAULT_NO_CROSSED = "\e[29m",
```
(naming could be improved)

lib/std/core/ansi.c3
lyranie: IIRC some shells don't support `\e` but i'm unsure

lib/std/core/ansi.c3
lyranie: will do that

lib/std/core/ansi.c3
FoxKiana: i could do `\u033`

BWindey: `\u033` is a fancy way of writing the byte `27` in decimal or `1B` in hex. `\e` is the escape-literal that C3 provides, mapping to byte `27`. Same as `\n` maps to byte `10`, or `b` to byte `8`. 

So basically, `\e == \u033 == 27 == 033`

lyranie: i see but to be 100% sure I'd like to keep it as is. primarily because I'm lazy

lerno:
Can you add some unit tests for sanity testing?

lerno:
Make the module std::core::string::ansi perhaps? Or maybe move it out of core entirely 🤔 I am not sure.

lyranie:
> Make the module std::core::string::ansi perhaps? Or maybe move it out of core entirely 🤔 I am not sure.

i was gonna make it like ascii is currently `std::core::ascii`

lyranie:
> Can you add some unit tests for sanity testing?

add tests for the function returns right? can't really test the color output

lyranie:
> ```c++
> <*
> 	Returns a String that will turn the following text into the specified
> 	RGB colour if the terminal supports it.
> *>
> macro String @rgb(char #r, char #g, char #b, bool $bg = false) @const {
> 	var $s;
> 	$if $bg :
> 		$s = "48";
> 	$else
> 		$s = "38";
> 	$endif
> 
> 	return "\e[" +++ $s +++ ";2;"
> 		+++ $stringify(#r) +++ ";"
> 		+++ $stringify(#g) +++ ";"
> 		+++ $stringify(#b) +++ "m";
> }
> ```

@lerno thoughts on making it compile time?

lerno:
> add tests for the function returns right? can't really test the color output

For the functions doing 8 bit color and such, it's in case someone later would modify it.

lerno:
> @lerno thoughts on making it compile time?

CT variants are possible, but then you should be using `$foo` arguments, not `#foo`. Use `@sprintf` to concatenate strings, not `+++`. This also helps avoiding `$stringify`.

```c
return "\e[" +++ $s +++ ";2;"
             +++ $stringify(#r) +++ ";"
             +++ $stringify(#g) +++ ";"
             +++ $stringify(#b) +++ "m";
// =>
return @sprintf("\e[%s;2;%s;%s;%sm", $s, $r, $g, $b);
```

lyranie:
> CT variants are possible, but then you should be using `$foo` arguments, not `#foo`. Use `@sprintf` to concatenate strings, not `+++`. This also helps avoiding `$stringify`.

can do. will update this later

lyranie:
also what about the module name? i can change it to `std::core::string::ansi` and would then write the tests in `test/test_suite/strings`

BWindey:
> > @lerno thoughts on making it compile time?
> 
> CT variants are possible, but then you should be using `$foo` arguments, not `#foo`. Use `@sprintf` to concatenate strings, not `+++`. This also helps avoiding `$stringify`.
> 
> ```c
> return "\e[" +++ $s +++ ";2;"
>              +++ $stringify(#r) +++ ";"
>              +++ $stringify(#g) +++ ";"
>              +++ $stringify(#b) +++ "m";
> // =>
> return @sprintf("\e[%s;2;%s;%s;%sm", $s, $r, $g, $b);
> ```

That's way smarter! Of course, at the time of me writing that macro, `@sprintf` wasn't added to the language yet.

lyranie:
@BWindey i added the other formatting codes you suggested. just waiting on a reply from lerno regarding the module

lib/std/core/ansi.c3
```diff
@@ -0,0 +1,71 @@
...
+
+<*
+ 8-bit color code
+
+ @return `the formatting char for the given background color`
+*>
+macro String color_8bit(char $index, bool $bg = false)
+{
+	int $mode = $bg ? 4 : 3;
+    return @sprintf("\u001B[%s8;5;%sm", $mode, $index);
```
BWindey: Different indentations. Same for the other function.

lyranie: oop! looked the same to me. i probably had a space and then tabbed so it indented improperly. will fix

lerno:
Thank you!

PR #2268, opened 2025-07-03
Disable libxml2 dependency on linux

ryuukk:
I get:

```
$ c3c compile-run ../resources/testfragments/helloworld.c3 
c3c: /usr/lib/libxml2.so.2: no version information available (required by c3c)
c3c: /usr/lib/libxml2.so.2: no version information available (required by c3c)
Program linked to executable 'helloworld'.
Launching ./helloworld
Hello, World!
```

Due to: https://gitlab.archlinux.org/archlinux/packaging/packages/libxml2/-/issues/4

LDC had same issue: https://github.com/ldc-developers/ldc/issues/4921


lerno:
Thanks, looks like it works and it's always great to get rid of dependencies.

PR #2271, opened 2025-07-04
Add null check for thread function pointer

velikoss:
Check thread.thread_fn before calling to prevent null pointer panic.


velikoss:
Closing because the issue wasn't here

PR #2273, opened 2025-07-04
Add WHIRLPOOL hashing to stdlib

NotsoanoNimus:
Dedicating my recent work on a [WHIRLPOOL hashing library](https://github.com/NotsoanoNimus/whirlpool.c3l) to the C3 stdlib.

This is an alternative way to generate a 512-bit hash that has a strong focus on __security__ over speed. Therefore, it should be used in applications like digital signatures rather than latency-sensitive areas like real-time communication.

:exclamation: :exclamation: :exclamation:
A really neat thing about the C3 implementation is that it can successfully hash up to the original WHIRLPOOL's proclaimed 2^256 bits of data, unlike popular implementations such as RHash which can only do 2^64 bits for [“optimization purposes”](https://github.com/rhash/RHash/blob/master/librhash/whirlpool.h#L18).

Additionally, C3's version has HMAC and PBKDF2: very rare to see these so accessible for WHIRLPOOL. :)
:exclamation: :exclamation: :exclamation:

-----
Related unit tests have been run before committing and are passing:
```
--------------------------------- TESTS ----------------------------------
Testing std::hash::whirlpool_test::test_whirlpool_empty ............. [PASS]
Testing std::hash::whirlpool_test::test_whirlpool_expected_outputs .. [PASS]
Testing std::hash::whirlpool_test::test_whirlpool_gigahash_zeroes ... [PASS]
Testing std::hash::whirlpool_test::test_whirlpool_hmac .............. [PASS]
Testing std::hash::whirlpool_test::test_whirlpool_pbkdf2 ............ [PASS]
Testing std::hash::whirlpool_test::test_whirlpool_pbkdf2_2 .......... [PASS]
Testing std::hash::whirlpool_test::test_whirlpool_pbkdf2_3 .......... [PASS]
Testing std::hash::whirlpool_test::test_whirlpool_pbkdf2_4 .......... [PASS]
Testing std::hash::whirlpool_test::test_whirlpool_sweep_block_size .. [PASS]

9 tests run.

Test Result: PASSED: 9 passed, 0 failed, 0 skipped.
```


lib/std/hash/whirlpool/whirlpool.c3
```diff
@@ -0,0 +1,188 @@
...
+fn void Whirlpool.init(&self) @inline => *self = { {}, {}, 0, 0 };
+
+
+fn void Whirlpool.update(&self, char[] data)
+{
+    char remainder = (char)(self.counter_low & 0x3F);   // if not at an even BLOCK_SIZE, how many bytes are over it
+    uint to_pad = BLOCK_SIZE - remainder;   // how many bytes must be filled to get the BLOCK_SIZE
+
+    // Trim the input down to INT64_MAX in length. Anything above that is unrealistic anyways.
+    if (@unlikely(data.len > long.max)) data = data[:long.max];
```
lerno: I don't think the data is over 8 extabyte of data either. But more importantly, you should restrict the yourself to an isz.max I think.

lerno: Since the compiler will assume that isz.max is the largest difference between two pointers.

lerno: But on the positive side you're showing a bug in the compiler 😅

NotsoanoNimus: I could change that but it’s possible that `update` might receive a lot more than 2 or 4 GiB at a time to consume. Is there other way to trim or window the data properly with this limitation? I just want it to be future-proof without needing to window the algorithm any further.

lerno: If you leave it at a long, then we run this on a 32 bit machine that has 32 bit pointers. What will a slice starting from 0 to +2^63 be? It just doesn't work. In practice anything past isz.max is broken. And on 32 bit machines that's 2^31

That's why you shouldn't use `long.max` but `isz.max`

NotsoanoNimus: Ahh I see, makes total sense! I hadn’t considered a 32-bit architecture.

Unfortunately there isn’t a good way to cause a fault here because I still want the HMAC stuff to work without making any extraneous stdlib modifications, and the HMAC module doesn’t do any unwrapping. 

I’ll have to chew on this a bit more perhaps.

NotsoanoNimus: Okay, so I deleted my previous comment because I literally wasn't paying attention thanks in part to holiday festivities and whatnot. I've updated my code to the proper architecture-based reasonable limitation. Thanks again.

lerno:
Do my changes look good?

NotsoanoNimus:
> Do my changes look good?

LGTM @lerno, thanks for the closer look at this submission. It looks a lot more _C3_ than _C_ now, and I learned a couple things!

lerno:
I added it to test that update works when calling it twice. Since all of the other tests were running update only once.

PR #2287, opened 2025-07-10
Add SipHash Family of Keyed PRFs

NotsoanoNimus:
Implementation of the [SipHash family](https://en.wikipedia.org/wiki/SipHash) of pseudorandom functions, complete with accompanying tests.

Predefined modules include:
- `siphash24` - _Performant_ - 64-bit hash from 2 rounds-per-block and 4 finalization rounds.
- `siphash24_128` - Same as above but for 128-bit outputs.
- `siphash48` - _Hardened_ - 64-bit hash from 4 rounds-per-block and 8 finalization rounds.
- `siphash48_128` - Same as above but for 128-bit outputs.

Out-of-the-box usage:
```c
uint128 hash = siphash24_128::hash(input, runtime_key);

for (usz i = 0; i < 4; ++i)
{
    // overloaded '[]' operator sets Bloom filter bits accordingly
    bloom_filter[(hash >> (32 * i)) & 0xFFFF_FFFF];
}
```

Other non-standard variations of SipHash can be used with an `alias` declaration like so:
```
import std::hash::siphash;

alias MySipHashFlavor = SipHash { uint128, 8, 16 };
alias hash = siphash::hash { uint128, 8, 16 };
```

### DIFF
--- a/lib/std/hash/siphash.c3
+++ b/lib/std/hash/siphash.c3
@@ -1,0 +1,176 @@
+// Copyright (c) 2025 Zack Puhl <github@xmit.xyz>. All rights reserved.
+// Use of this source code is governed by the MIT license
+// a copy of which can be found in the LICENSE_STDLIB file.
+//
+// SipHash is a secure pseudorandom function (PRF) which digests a 128-bit key
+//   and a variable-length message to produce a 64- or 128-bit hash value. 
+//
+// SipHash can be employed in numerous useful ways and structures, e.g.:
+//   - Hash Tables
+//   - Message Authentication Codes
+//   - Denial of Service (hash flooding) resistance
+//   - Bloom filters
+//   - Keyed runtime identifier derivation
+//
+// Read more: https://en.wikipedia.org/wiki/SipHash
+//
+module std::hash::siphash { OutType, BLOCK_ROUNDS, FINALIZE_ROUNDS };
+
+
+struct SipHash
+{
+	ulong[4] v;
+	long m;
+	int m_idx;
+	usz len;
+	union
+	{
+		uint128 key;
+		ulong[2] key_64;
+	}
+}
+
+
+<*
+@require OutType.typeid == uint128.typeid || OutType.typeid == ulong.typeid : "Module OutType must be either uint128 or ulong."
+*>
+fn OutType hash(char[] data, uint128 key)
+{
+	SipHash s;
+	s.init(key);
+	s.update(data);
+	return s.final();
+}
+
+<*
+@require OutType.typeid == uint128.typeid || OutType.typeid == ulong.typeid : "Module OutType must be either uint128 or ulong."
+*>
+fn void SipHash.init(&self, uint128 key)
+{
+	self.key = key;
+	self.v = {
+		0x736f6d6570736575 ^ self.key_64[0],
+		0x646f72616e646f6d ^ self.key_64[1],
+		0x6c7967656e657261 ^ self.key_64[0],
+		0x7465646279746573 ^ self.key_64[1],
+	};
+
+	$if OutType.typeid == uint128.typeid :
+		self.v[1] ^= 0xEE;
+	$endif
+}
+
+
+fn void SipHash.update(&self, char[] data)
+{
+	self.len += data.len;
+
+	foreach (byte : data)
+	{
+		self.m |= ((long)byte << (self.m_idx++ << 3));
+
+		if (self.m_idx < 8) continue;
+
+		// This runs every time the m_idx is 8, then it resets to 0.
+		self.v[3] ^= self.m;
+
+		$for var $i = 0; $i < BLOCK_ROUNDS; ++$i :	// unrolled loop
+			self.round();
+		$endfor
+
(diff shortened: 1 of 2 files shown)


lib/std/hash/siphash.c3
```diff
@@ -0,0 +1,176 @@
...
+fn OutType hash(char[] data, uint128 key)
+{
+	SipHash s;
+	s.init(key);
+	s.update(data);
+	return s.final();
+}
+
+<*
+@require OutType.typeid == uint128.typeid || OutType.typeid == ulong.typeid : "Module OutType must be either uint128 or ulong."
```
lerno: Place this above the module declaration itself, it should correctly enforce the contract.

NotsoanoNimus: Wait _what_, generic constraints for modules?! So sick :fire:

Fix pushed.

lib/std/hash/siphash.c3
```diff
@@ -0,0 +1,174 @@
...
+
+		self.m_idx = 0;
+		self.m = 0;
+	}
+}
+
+
+fn OutType SipHash.final(&self)
+{
+	char message_length_mod_256 = (char)(self.len % 0xFF);
```
lerno: This looks like a bug. You probably intended `(self.len & 0xFF)`. But casting to `char` will just give the last bits anyway, so `(char)self.len` is fine. This is implementation defined in C but well defined in C3. It will do a truncation.

lib/std/hash/siphash.c3
```diff
@@ -0,0 +1,174 @@
...
+
+	while (self.m_idx < 7) self.update({0});
```
lerno: A bit obscure perhaps, but this can be written:
```c
if (self.m_idx < 7) self.update((char[7]){}[:7 - self.m_idx]);
```
For a single call to update.

lerno: In fact, we can replace that and the following row with this:
```c
char[8] last = { [7] = message_length_mod_256 };
self.update(last[(self.m_idx < 7 ? self.m_idx : 7)..]);
```

lib/std/hash/siphash.c3
```diff
@@ -0,0 +1,174 @@
...
+	$else
+		self.v[2] ^= 0xFF;
+	$endif
+
+	$for var $i = 0; $i < FINALIZE_ROUNDS; ++$i :	// unrolled loop
+		self.round();
+	$endfor
+
+	$if OutType.typeid == ulong.typeid :
+		return (ulong)(self.v[0] ^ self.v[1] ^ self.v[2] ^ self.v[3]);
```
lerno: Redundant cast, it's already ulong

lib/std/hash/siphash.c3
```diff
@@ -0,0 +1,174 @@
...
+	$else
+		ulong lo = (self.v[0] ^ self.v[1] ^ self.v[2] ^ self.v[3]);
```
lerno: Unnecessary parenthesis

lib/std/hash/siphash.c3
```diff
@@ -0,0 +1,174 @@
...
+
+		self.v[1] ^= 0xDD;
+
+		$for var $i = 0; $i < FINALIZE_ROUNDS; ++$i :	// unrolled loop
+			self.round();
+		$endfor
+
+		ulong hi = (ulong)(self.v[0] ^ self.v[1] ^ self.v[2] ^ self.v[3]);
```
lerno: Unnecessary cast

lib/std/hash/siphash.c3
```diff
@@ -0,0 +1,174 @@
...
+
+		return ((uint128)lo | ((uint128)hi << (uint128)64));
```
lerno: Reduce the need for casts to:
```c
return lo | ((uint128)hi << 64);
```

lib/std/hash/siphash.c3
```diff
@@ -0,0 +1,174 @@
...
+{
+	SipHash s;
+	s.init(key);
+	s.update(data);
+	return s.final();
+}
+
+fn void SipHash.init(&self, uint128 key)
+{
+	self.key = key;
```
lerno: The key seems like it's not actually used after init? If so, then the key init code can be replaced by:
```c
	ulong[2] key_64 = bitcast(key, ulong[2]);
	self.v = {
		0x736f6d6570736575 ^ key_64[0],
		0x646f72616e646f6d ^ key_64[1],
		0x6c7967656e657261 ^ key_64[0],
		0x7465646279746573 ^ key_64[1],
	};
```

lerno:
I posted some suggested improvements above.

NotsoanoNimus:
I incorporated your feedback, and then some. I tend to put a lot of idiosyncrasies around explicit casts and parentheses in my code because I'm paranoid of C's UB and other spooky behaviors, so it'll take a bit for me to be a little more lax with the help of C3.

Thank you for your time reviewing this.

lerno:
Updated the test to use this:
```c
char[64] plaintext = math::iota(char[64]);
```

PR #2290, opened 2025-07-11
Fix json parser

sudison:
Added a few more fixes to pass all the tests in https://github.com/nst/JSONTestSuite/tree/master/test_parsing

### DIFF
--- a/lib/std/encoding/json.c3
+++ b/lib/std/encoding/json.c3
@@ -132,9 +132,12 @@ fn JsonTokenType? lex_number(JsonContext *context, char c) @local
 		if (c == '.')
 		{
 			t.append(c);
-			while (c = read_next(context)!, c.is_digit())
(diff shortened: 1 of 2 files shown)


lerno:
Does this strictness have any practical use?

lerno:
I am tempted to accept this, but only if it's possible to opt out. The reason is that in practice you run into json code generators that are BAD, and you unfortunately need to work around those or write your own JSON parser. So add a "lenient mode" for this, that accepts these (and perhaps even more!) of the *strictly incorrect* json.

Also, I've very sorry being very later to pick this up.

lerno:
Alternatively I could use your code as a basis and create this "lenient mode" myself on top of this.

lerno:
I need some response to this or I will need to close it.

lerno:
Closing it then.

PR #2297, opened 2025-07-12
Implement Ed25519

Airtz:
Implementation of the [Ed25519 Digital Signature Algorithm](https://en.wikipedia.org/wiki/EdDSA#Ed25519).

1 main module : `std::crypto::ed25519`
3 internal submodules : `std::crypto::ed25519::c25519`, `std::crypto::ed25519::f25519` and `std::crypto::ed25519::fbase`

API:
* `alias PrivateKey = char[32]`
* `alias PublicKey = char[PrivateKey.len]`
* `alias Signature = char[2 * PublicKey.len]`
* `fn void public_keygen(char[] private_key)`
* `fn Signature sign(char[] message, char[] private_key, char[] public_key)`
* `fn bool verify(char[] message, char[] signature, char[] public_key)`

Exemple usage :
```c
PrivateKey private_key = x"97b0d0435be3a583df0b77aff277e5fc8f857cae1a827723241bb9bbf7e96018";

PublicKey public_key = ed25519::public_keygen(&private_key);
assert(public_key == x"fd462c5c7044b224f4a6cc0f0e5e16511be881be639b3fbc600f7b5d6da54e2f");

String message = "abcdefghijklmnopqrstuvwxyz0123456789";

Signature signature = ed25519::sign(message, &private_key, &public_key);
assert(signature == x"c03cf7b181583f6353a9434be910a484e1720247a0dc384d601e495294d0202728dda5f31bc62b00158b1588bc2a65f48ccfaa1b1431aff3d8da9271b7de9403");
assert(ed25519::verify(message, &signature, &public_key));

```

### DIFF
--- a/lib/std/crypto/ed25519.c3
+++ b/lib/std/crypto/ed25519.c3
@@ -1,0 +1,135 @@
+/*
+ Ed25519 Digital Signature Algorithm
+*/
+
+module std::crypto::ed25519;
+import std::hash::sha512;
+
+alias PrivateKey = char[32];
+alias PublicKey = char[PrivateKey.len];
+alias Signature = char[2 * PublicKey.len];
+
+<*
+ Generate a public key from a private key.
+
+ @param [in] private_key : "32 bytes of cryptographically secure random data"
+ @require private_key.len == PrivateKey.len
+*>
+fn PublicKey public_keygen(char[] private_key) => c25519::pack(&&c25519::unproject(&&BASE.mul(expand_private_key(private_key)[:fbase::Int.len])));
+
+<*
+ Sign a message.
+
+ @param [in] message
+ @param [in] private_key
+ @param [in] public_key
+ @require private_key.len == PrivateKey.len
+ @require public_key.len == PublicKey.len
+*>
+fn Signature sign(char[] message, char[] private_key, char[] public_key) {
+    Signature r @noinit;
+
+    char[*] exp = expand_private_key(private_key);
+
+    Sha512 sha @noinit;
+    sha.init();
+
+    sha.update(exp[fbase::Int.len..]);
+    sha.update(message);
+
+    fbase::Int k = fbase::from_bytes(&&sha.final());
+
+    r[:f25519::Int.len] = c25519::pack(&&c25519::unproject(&&BASE.mul(k[..])))[..];
+
+    sha.init();
+
+    sha.update(r[:f25519::Int.len]);
+    sha.update(public_key);
+    sha.update(message);
+
+    fbase::Int z = fbase::from_bytes(&&sha.final());
+    fbase::Int e = fbase::from_bytes(exp[:fbase::Int.len]);
+
+    r[f25519::Int.len..] = z.mul(&e).add(&k)[..];
+
+    return r;
+}
+
+<*
+ Verify the signature of a message.
+
+ @param [in] message
+ @param [in] signature
+ @param [in] public_key
+ @require signature.len == Signature.len
+ @require public_key.len == PublicKey.len
+*>
+fn bool verify(char[] message, char[] signature, char[] public_key) {
+    char ok = 1;
+
+    f25519::Int lhs = c25519::pack(&&c25519::unproject(&&BASE.mul(signature[f25519::Int.len..])));
+
+    Unpacking unp_p = c25519::unpack_on_curve((f25519::Int*)public_key);
+    Projection p = c25519::project(&unp_p.point);
+    ok &= unp_p.on_curve;
+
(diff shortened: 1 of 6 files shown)


lerno:
I note that this includes a 256 bit int implementation. Is it possible to perhaps use the BigInt implementation already in std::math?

lib/std/crypto/ed25519/c25519.c3
```diff
@@ -0,0 +1,193 @@
...
+    return {p, f25519::eq(&x2, &_x2)};
+}
+
+// 2*d
+const Int DD = {
+    0x59, 0xf1, 0xb2, 0x26, 0x94, 0x9b, 0xd6, 0xeb,
+    0x56, 0xb1, 0x83, 0x82, 0x9a, 0x14, 0xe0, 0x00,
+    0x30, 0xd1, 0xf3, 0xee, 0xf2, 0x80, 0x8e, 0x19,
+    0xe7, 0xfc, 0xdf, 0x56, 0xdc, 0xd9, 0x06, 0x24
+};
```
lerno: Something like this can use the hex literals:
```c
const Int DD = x"59f1b226949bd6eb"
                "56b183829a14e000"
                "30d1f3eef2808e19"
                "e7fcdf56dcd90624";
```

lib/std/crypto/ed25519.c3
```diff
@@ -0,0 +1,136 @@
...
+ Verify the signature of a message.
+
+ @param [in] message
+ @param [in] signature
+ @param [in] public_key
+ @require signature.len == Signature.len
+ @require public_key.len == PublicKey.len
+*>
+fn bool verify(char[] message, char[] signature, char[] public_key) {
+    char ok = 1;
```
lerno: As a general guide, stick to the coding style in the rest of the standard library (Allman braces, tabs for indentation)

lib/std/crypto/ed25519.c3
```diff
@@ -0,0 +1,136 @@
...
+    sha.init();
+
+    sha.update(r[:f25519::Int.len]);
+    sha.update(public_key);
+    sha.update(message);
+
+    fbase::Int z = fbase::from_bytes(&&sha.final());
+    fbase::Int e = fbase::from_bytes(exp[:fbase::Int.len]);
+
+    r[f25519::Int.len..] = z.mul(&e).add(&k)[..];
```
lerno: Following the general code style, type names are made unique on their own. They are not made unique through module namespacing, so `F25519Int` would be fine, `f25519::Int` would not be. In this case, rather than making it long, make c25519 module part of the std::crypto::ed25519, just with private visibility.

as2te:
> I note that this includes a 256 bit int implementation. Is it possible to perhaps use the BigInt implementation already in std::math?

No, since :
* Not leveraging the fact that you're operating on a prime field (that is, mod 2^255-19) would be much slower.
* BigInt is not constant-time.
* BigInt is limited to 32 bytes and this needs to handle up to 64.

Thank you for the feedback. I'll see to address the style issues.

lerno:
See if you can try to make it self contained in a single file, unless there are more that will share the same implementation.

lib/std/crypto/ed25519.c3
```diff
@@ -125,12 +121,625 @@ const Projection BASE @private = {
...
+	*s = f25519_select(&sub, s, (char)(c >> 15));
+}
+
+<*
+ Constant-time equality comparison. Return is non-zero if true.
+
+ @param [&in] a
+ @param [&in] b
+*>
+fn char eq(F25519Int* a, F25519Int* b)
```
as2te, line 389: This is `std::crypto::safe_compare` but constant-time. Unsure why `safe_compare` use `==` to return a boolean since this operator is most likely not constant-time.

lib/std/crypto/ed25519.c3
```diff
@@ -125,12 +121,625 @@ const Projection BASE @private = {
...
+	return e ^ 1;
+}
+
+<*
+ Constant-time conditonal selection. Result is undefined if condition is neither 0 nor 1.
+
+ @param [&in] zero : "selected if condition is 0"
+ @param [&in] one : "selected if condition is 1"
+*>
+fn F25519Int f25519_select(F25519Int* zero, F25519Int* one, char condition)
```
as2te, line 407: Maybe this should be abstracted in `std::crypto` alongside `safe_compare`.

as2te:
So, operator overloading not working on pointers was indeed a bug 😬 
I assumed it was on purpose to enforce purity and defaulted to method chains. Looks much better now !

lerno:
Thank you!

PR #2302, opened 2025-07-15
Add `ConditionVariable.wait_until` and `ConditionVariable.wait_for`

cbuttner:


cbuttner:
I should mention `wait_for` overlaps with `wait_timeout`, so maybe `wait_timeout` could be deprecated.

lerno:
What about `wait_duration` rather than `wait_for`

cbuttner:
Actually I think `wait_timeout` is a good name, because it doesn't imply that it guarantees to block for the full duration (it might not due to spurious wakeups). 
But how should it take a Duration instead of ms without it being a breaking change?

lerno:
Well, you can be tricky and use macros for it. But there's a problem with Duration / NanoDuration and just ms because of how they're converted from literals. So for now you need to assume passing "1" means ms and not Duration. But this is a bigger problem, where I wonder if typedefs should be able to opt out of implicit literal conversion.

lerno:
I modified so that timeout worked for either, allowing us to deprecate ms later.

lerno:
Thank you!

PR #2303, opened 2025-07-15
Improve atomic.c3

cbuttner:
- Change `Atomic.xor` from function to macro
- Allow function pointers as native atomic type
- Use enum inference


lerno:
Thank you, looks good!

PR #2306, opened 2025-07-16
Fix `Formatter.print` returning incorrect size

cbuttner:


cbuttner:
How do I fix this test case?

lerno:
It's breaking because the size is changing of Formatter, so find the following line in the test file:
```
@llvm.memset.p0.i64(ptr align 8 %formatter, i8 0, i64 48, i1 false)
```
And change it to:
```
@llvm.memset.p0.i64(ptr align 8 %formatter, i8 0, i64 40, i1 false)
```

That should make it pass.

lerno:
Can you also add a test that shows that print now returns the correct length?

lerno:
Thank you!

PR #2309, opened 2025-07-17
Tidy up path.c3

lyranie:
- Added `@operator(==)` to `equals` function
- Inlined non-pool functions
- made `path` untyped in `mkdir`

### DIFF
--- a/lib/std/io/path.c3
+++ b/lib/std/io/path.c3
@@ -155,10 +155,7 @@ fn Path? from_wstring(Allocator allocator, WString path) => @pool()
 	return path::new(allocator, string::tfrom_wstring(path)!);
 }
 
-fn Path? from_win32_wstring(Allocator allocator, WString path) @deprecated => @pool()
-{
-	return path::new(allocator, string::tfrom_wstring(path)!);
-}
+fn Path? from_win32_wstring(Allocator allocator, WString path) @deprecated("Use 'from_wstring' instead") => from_wstring(allocator, path);
 
 fn Path? for_windows(Allocator allocator, String path) => new(allocator, path, WIN32);
 


lib/std/io/path.c3
```diff
@@ -148,27 +148,18 @@ fn Path? new(Allocator allocator, String path, PathEnv path_env = DEFAULT_ENV)
 
  @return? INVALID_PATH : `if the path was invalid`
 *>
-fn Path? temp(String path, PathEnv path_env = DEFAULT_ENV)
-{
-	return new(tmem, path, path_env);
-}
+fn Path? temp(String path, PathEnv path_env = DEFAULT_ENV) => new(tmem, path, path_env);
 
-fn Path? from_win32_wstring(Allocator allocator, WString path) => @pool()
```
lerno, line 156: Changing the API isn't allowed, adding a new and deprecating the old is the way to go.

lyranie: i'll fix it

lib/std/io/path.c3
```diff
@@ -148,27 +148,23 @@ fn Path? new(Allocator allocator, String path, PathEnv path_env = DEFAULT_ENV)
...
+fn Path? from_wstring(Allocator allocator, WString path) => @pool()
 {
 	return path::new(allocator, string::tfrom_wstring(path)!);
 }
 
-fn Path? for_windows(Allocator allocator, String path)
+fn Path? from_win32_wstring(Allocator allocator, WString path) @deprecated => @pool()
```
lerno: You can just have this call `from_wstring` instead. And in `@deprecated` write `"Use 'from_wstring' instead"` to help migration.

lib/std/io/path.c3
```diff
@@ -148,27 +148,20 @@ fn Path? new(Allocator allocator, String path, PathEnv path_env = DEFAULT_ENV)
...
-{
-	return new(allocator, path, WIN32);
-}
+fn Path? from_win32_wstring(Allocator allocator, WString path) @deprecated("Use 'from_wstring' instead") => from_wstring(allocator, path);
 
-fn Path? for_posix(Allocator allocator, String path)
-{
-	return new(allocator, path, POSIX);
-}
+fn Path? for_windows(Allocator allocator, String path) => new(allocator, path, WIN32);
```
lerno, line 160: I think shortening is not really necessary here. In general I feel the short syntax is for prototyping and for when you have a lot of similar forwarding. Here it's already typed out, so let's keep it.

lerno:
Let me accept these changes as is and I'll look at Path a bit myself today. It's long overdue.

PR #2310, opened 2025-07-18
Add ability to compile on OpenBSD

LowByteFox:
I have added the ability to configure and compile c3c on OpenBSD, some features like cross-compiling to Darwin is not possible due to OpenBSD LLVM port maintainers

I have also fixed CURL include dir issue and fixed typo in OS_TYPE_FREEBSD enum

I hope everything looks okay and can fix if needed

I also increased/decreased(?) the min_size for arena because of OpenBSD memory limits


LowByteFox:
Just noticed I have commented out LLVM_MACHO and forgot about it

will fix tomorrow

lerno:
Doesn't this print the warning like 3 times? Maybe 1 time is enough 😅

lerno:
Seems to compile, so let's add it!

PR #2314, opened 2025-07-20
Support asm x86 popcnt

hamkoroke:
Hi there,

I'm still pretty new to the language and the repository but I wanted to start contributing a bit to this project.
In my humble opinion this language has a huge potential. I'm especially excited about the Allocators, it is a good paradigm that needs to be brought back instead of the more common RAII mindset.

Left this PR small on purpose because I was unsure if the instruction is properly added as well as the style of testing which maybe is not acceptable (I saw some tests related to ASM which involved some embedded output of LLVM but honestly did not know how to create those tests).

Thank you.

### DIFF
--- a/src/compiler/asm_target.c
+++ b/src/compiler/asm_target.c
@@ -709,6 +709,10 @@ static void init_asm_x86(PlatformTarget* target)
 	reg_instr(target, "pushw", "r16/mem/imm16");
 	reg_instr(target, "popw", "w:r16/mem");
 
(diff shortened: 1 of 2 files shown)


lerno:
Thank you for wanting to contribute to the compiler!

There's just one thing you need to do - since those tests can only work on x86 anyway, is to add the following to the test module. Change:

```c
module test::inline_asm::popcnt;
```
To
```c
module test::inline_asm::popcnt @if (env::X86_64);
```
This conditionally includes this test, so that it's only compiled on x86_64 archs.

lerno:
Awesome! Thank you

PR #2315, opened 2025-07-20
More Ed25519 cleanup

Airtz:
- Removed a few unecessary copies
- `foreach` instead of `for` when suitable
- Fixed a typo


lerno:
Nice cleanup! Thank you

PR #2316, opened 2025-07-21
Add libc::readlink()

FoxKiana:
Added

```cpp
isz readlink(ZString pathname, char* buf, int bufsize)
```

to `libc.c3`


lerno:
I think this one needs to be excluded for env::WIN32

lerno:
I'll fix

PR #2317, opened 2025-07-22
Improved OpenBSD support

LowByteFox:
This PR improves support for C3 under OpenBSD:
- `c3c` is capable of building programs
- `errno` is handled properly
- `c3c` should handle higher memory usage better, still advised to sometimes limit memory using `--max-mem`
- added backtrace support, still needs finishing touches
- `build-openbsd` has been added to CI allowing to build `c3c` under OpenBSD 7.7, usage on other version is discouraged
- running compiler tests is now limiting memory usage to not exhaust it


src/compiler/linker.c
```diff
@@ -532,6 +532,11 @@ static void linker_setup_freebsd(const char ***args_ref, Linker linker_type, boo
 {
 	if (linker_type == LINKER_CC) {
 		linking_add_link(&compiler.linking, "pthread");
+		linking_add_link(&compiler.linking, "execinfo"); // for backtrace
```
lerno, line 535: Is this valid for all bsds?

LowByteFox: According their man pages yes
FreeBSD: https://man.freebsd.org/cgi/man.cgi?query=backtrace&sektion=3
OpenBSD: https://man.openbsd.org/backtrace
NetBSD: https://man.netbsd.org/backtrace.3
DragonflyBSD: https://man.dragonflybsd.org/?command=backtrace&section=ANY

a lot of them borrowed it from NetBSD

lerno:
Thank you!

PR #2321, opened 2025-07-23
Support memory mapped files and add File.map

hamkoroke:
Hi again,

I made an attempt to extend the use of the new virtual memory module to memory mapped files.

* Added a map_file function to mem_vm (almost same than existing alloc function)
* Then extended the File interface to support memory mapping
* Added some tests

There are some points open to discussion:

* It allows both the offset and the size to be not page aligned (it will round down and up internally)

* The actual bytes that the user is interested in are provided by FileMap.bytes(), it discounts the extra bytes due to misalignment on offset and/or size.

* File.map() does not return a VirtualMemory struct, it returns a struct that contains a VirtualMemory object and the actual range expected by the user. So there is the range ptr+len that munmap needs and the range that the user expects.

Please help me with some feedback, happy to adapt or reduce functionality if allowing misalignment feels convoluted.


hamkoroke:
@lerno Sorry for the noise, I was trying to pass all the tests and was getting confused with the different environments.

My interpretation of the directives `@if(env::xxx)` was that if you use them in some module, for example:

```
fn Fd File.fd(self) @if(env::LIBC) {...}
```

then any further use of `File.fd()` on other module or function would automatically inherit the conditional environment.
But then I realized that I should set the attribute on the consumers of the module/function as well.

lib/std/core/os/mem_vm.c3
```diff
@@ -219,6 +219,39 @@ fn void? decommit(void* ptr, usz len, bool block = true)
...
+fn void*? map_file(Fd fd, usz size, usz offset = 0, VirtualMemoryAccess access = READ, bool shared = false)
+{
+	$switch:
+		$case env::POSIX:
+			CInt flags = shared ? posix::MAP_SHARED : posix::MAP_PRIVATE;
+			void* ptr = posix::mmap(null, aligned_alloc_size(size), access.to_posix(), flags, fd, offset);
+			if (ptr != posix::MAP_FAILED) return ptr;
+			switch (libc::errno())
+			{
+				case errno::ENOMEM:    return mem::OUT_OF_MEMORY?;
```
lerno: I think several errno should be handled here as well: EBADF, EACCESS. There might be others.

lib/std/core/os/mem_vm.c3
```diff
@@ -219,6 +219,39 @@ fn void? decommit(void* ptr, usz len, bool block = true)
...
+				case errno::EOVERFLOW: return RANGE_OVERFLOW?;
+				case errno::EPERM:     return ACCESS_DENIED?;
+				case errno::EINVAL:    return INVALID_ARGS?;
+				default:               return UNKNOWN_ERROR?;
+			}
+		$case env::WIN32:
+			unsupported("'vm.map_file' not implemented on Win32");
```
lerno: Rather than doing this it might be better to exclude it entirely. So set `@if (env::POSIX)` as an attribute on the function. This makes the function only appear in posix environments.

lerno:
Thank you for this contribution. Here are some thoughts around further improvements

Should perhaps the FileMap contain the file? I'm thinking in terms of lifetimes how they should relate to each other.

For example, will the usecase be this: `FileMap m = file::open("Foo.txt", "r").map()`. In this case we need the FileMap to contain the file, so that you can close both.

There's also the question whether it should be a method. I'm thinking it should perhaps not be, so rather than `file.map()` we should perhaps do `file::mmap_file(f)` and then also add one that does everything in a single sweep `FileMap map = file::mmap_open("Foo.txt", "r")`

If the file is contained in the struct, then that works fine. And perhaps the name should change to `FileMmap` or `MmapFile` or `MappedFile` or `MmappedFile`.

As for the rounding, I haven't given it a lot of thought. Can you run me through the pros and cons?

And overall I like it, it seems like a good addition. I'm just making sure we're covering all of the cases.

lib/std/core/os/mem_vm.c3
hamkoroke: Thank you, updated.

lib/std/core/os/mem_vm.c3
hamkoroke: Thank you, updated.

hamkoroke:
Good point about the r-value use case, addressed the issue with your proposed solution. Used functions instead of methods.
There is now a function to mmap an already opened file (which will not be closed, that is up to the user who opened it).
And another function to directly mmap the file from a filename which, in this case, because we own the file, it will be closed when the mapping is destroyed.

lerno:
Thanks for the update, I'll try to have time tomorrow, ping me if I forget

lerno:
Good work on this and solid tests. Thank you very much for the contribution.

hamkoroke:
@lerno Sorry, I have been quite busy at work these days and forgot the reminder. Hope I could get more involved on the project.

PR #2322, opened 2025-07-23
Add HashSet implementation

velikoss:
Add a generic HashSet with allocator support and standard set operations.

- Basic operations: add/remove/contains/clear
- Set operations: merge/intersection/difference/is_subset
- Iteration support

Example Usage:
```c3
import std::collections::set;

fn int main() {
    HashSet{int} set;
    set.tinit();
    
    // Add values
    set.add(1);
    set.add(2);
    set.add(3);
    
    // Check contains
    if (set.contains(2)) {
        io::printn("Set contains 2!");
    }
    
    // Iterate
    set.@each(; int value) {
        io::printfn("%d", value);
    };
    
    // Set operations
    HashSet{int} other;
    other.tinit();

    other.add(2);
    other.add(3);
    other.add(4);
    
    io::printn(set);
    io::printn(other);

    HashSet{int}* merged = set.tmerge(&other);
    HashSet{int}* common = set.tintersection(&other);

    io::printn(merged);
    io::printn(common);
    
    return 0;
}
```

Output:
```
Set contains 2!
2
3
1
{ 2, 3, 1 }
{ 4, 2, 3 }
{ 3, 4, 1, 2 }
{ 2, 3 }
```

### DIFF
--- a/lib/std/collections/hashset.c3
+++ b/lib/std/collections/hashset.c3
@@ -340,7 +340,7 @@ fn void HashSet.reserve(&set, usz capacity)
  @param [&inout] allocator : "Allocator for the new set"
  @return "A new set containing the union of both sets"
 *>
-fn HashSet HashSet.union_with(&self, Allocator allocator, HashSet* other)
+fn HashSet HashSet.set_union(&self, Allocator allocator, HashSet* other)
 {
 	usz new_capacity = math::next_power_of_2(self.count + other.count);
 	HashSet result;
@@ -350,7 +350,7 @@ fn HashSet HashSet.union_with(&self, Allocator allocator, HashSet* other)
 	return result;
 }
 
-fn HashSet HashSet.tunion_with(&self, HashSet* other) => self.union_with(tmem, other);
+fn HashSet HashSet.tset_union(&self, HashSet* other) => self.set_union(tmem, other);
 
 <*
  Returns the intersection of the two sets (A & B)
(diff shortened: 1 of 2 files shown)


lerno:
Could you add unit tests as well please?  Ideally testing every function.

velikoss:
I've added unit testing for every function. Also fixed some functions naming

lib/std/collections/hashset.c3
```diff
@@ -0,0 +1,561 @@
...
+{
+	return !set.count;
+}
+
+<*
+ Get the value or add and
+ @require @assignable_to(#expr, Value)
+*>
+macro Value HashSet.@add_or_get(&set, Value #expr)
+{
```
lerno, line 168: Isn't this kind of a weird macro?

lib/std/collections/hashset.c3
```diff
@@ -0,0 +1,561 @@
...
+
+macro HashSet.@each(set; @body(value))
+{
+    set.@each_entry(; Entry* entry)
+	{
+		@body(entry.value);
+	};
+}
+
+macro HashSet.@each_entry(set; @body(entry))
```
lerno: This also seems to be something that only makes sense for hash maps?

lib/std/collections/hashset.c3
```diff
@@ -0,0 +1,561 @@
...
+
+// --- Set Operations ---
+
+<*
+ @param [&in] other : "The other set to union with"
+ @param [&inout] allocator : "Allocator for the new set"
+ @return "A new set containing the union of both sets"
+*>
+fn HashSet* HashSet.merge(&self, Allocator allocator, HashSet* other)
+{
```
lerno: This should return a HashSet with the allocator inside of it set to the allocator, but it shouldn't necessarily be a pointer itself or?

lib/std/collections/hashset.c3
velikoss: I used HashMap as a reference for what functions should container have, so some functions were "inherited"

lib/std/collections/hashset.c3
```diff
@@ -0,0 +1,561 @@
...
+}
+
+fn HashSet* HashSet.tmerge(&self, HashSet* other) => self.merge(tmem, other);
+
+<*
+ @param [&in] other : "The other set to intersect with"
+ @param [&inout] allocator : "Allocator for the new set"
+ @return "A new set containing the intersection of both sets"
+*>
+fn HashSet* HashSet.intersection(&self, Allocator allocator, HashSet* other)
```
lerno: Same regarding being a pointer as the other there.

lib/std/collections/hashset.c3
velikoss: I actually don't know for sure how to do this kind of thing. I also thought about that but landed on this solution

Issue #2325, opened 2025-07-23
Unwrap Variables from Caught Assignment Expressions


NotsoanoNimus:
Title might be a bit misleading. I'm hoping to be able to use optional variables as their non-optional types after evaluating an assignment expression within `if catch` as follows:

```c
File? f;

if (catch err = (f = file::open(options.output_path, "w"))) @die("blah blah");

File some_other_nonoptional_file = f;
```

Results in this error for the final line:
```
Error: It is not possible to cast from 'File?' to 'File'.
```

... even though at this point in the code, `f` has been evaluated to have not been a fault. This same code of course compiles just fine if the assignment operation `f = file::open...` occurs before/outside the `if catch`. 

Additionally, I should note that catching faults is not the problem here; that works just fine.

-----
For the record, I am aware of why the below would likely ***not work***, and is not what I'm asking for:
```c
File some_other_nonoptional_file;

if (catch err = (some_other_nonoptional_file = file::open(options.output_path, "w"))) @die("blah blah");
```

Lastly, I'm aware syntax is ugly here, so if this behavior doesn't fit the C3 vision I am fine with that too.

lerno:
Is `@die` here marked as `@noreturn`?

NotsoanoNimus:
It wasn't, the `main` method in my program is wrapping an optional-style `main` that's easy to fault out of and still run my `defer` statements.

But with the below, it still doesn't work as expected:

```c++
module aux_test;

import std::io;
import std::os;

File static_file_handle;

macro @die(msg, args...) @noreturn
{
    io::eprintf("AAAAAAHHHH:  ");
    io::eprintfn(msg, ...args);

    os::exit(1);
}

fn int main() => main_inner() ?? 1;

fn int? main_inner()
{
    // Bunch of `defer` statements I MUST execute... NO ABORT ALLOWED.

    File? f;
    // DOES NOT WORK
    // if (catch err = (f = file::open("/tmp/ooo", "w"))) @die("oh no");

    // WORKS
    f = file::open("/tmp/ooo", "w");
    if (catch err = f) @die("oh no");

    static_file_handle = f;
    defer (void)f.close();

    // do stuff

    return 0;
}
```

```text
27:     // f = file::open("/tmp/ooo", "w");
28:     // if (catch err = f) @die("oh no");
29: 
30:     static_file_handle = f;
                             ^
(aux_test.c3:30:26) Error: It is not possible to cast from 'File?' to 'File'.
```

[...]

PR #2333, opened 2025-07-26
Add @try_set Builtin

NotsoanoNimus:
See Issue #2325 for more details on this.

Instead of needing to do this:
```c++
    File? f;
    // DOES NOT WORK
    // if (catch err = (f = file::open("/tmp/ooo", "w"))) @die("oh no %s", err);

    // WORKS
    f = file::open("/tmp/ooo", "w");
    if (catch err = f) @die("oh no");
```

Enable the "DOES NOT WORK" section in a slightly different way with this macro:
```c++
    File f;
    if (catch err = @assign_or(f, file::open("/tmp/ooo", "w"))) @die("oh no %s", err);
```

In this case, `f` is now of type `File` instead of `File?`, and we can use the assigned `f` just fine after the `if catch` if no fault was captured.

In the macro `#expr` is assigned to `res` in order to ensure the expression is only evaluated  once when `@assign_or` is called.

### DIFF
--- a/lib/std/core/builtin.c3
+++ b/lib/std/core/builtin.c3
@@ -410,6 +410,7 @@ macro swizzle2(v, v2, ...) @builtin
 {
 	return $$swizzle2(v, v2, $vasplat);
 }
+
 <*
  Return the excuse in the Optional if it is Empty, otherwise
  return a null fault.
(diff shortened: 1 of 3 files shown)


vssukharev:
I would rather rename it to `@setor`

lerno:
The name perhaps needs some thinking on. Maybe `@try_assign` or `@try_set` or something? I am not sure.

NotsoanoNimus:
> The name perhaps needs some thinking on. Maybe `@try_assign` or `@try_set` or something? I am not sure.

`@try_set` sounds like a nice compromise, and fits intuitively with the `if catch` style statement. What do you think, @vssukharev?

vssukharev:
Agree, `@try_set` is better

Issue #2332, opened 2025-07-26
LinkedHashMap tests: hash_map_copy.tkeys()==[,,];hash_map_copy.tvalues()=[0,0,0]


LukyGuyLucky:
Testing on [https://github.com/c3lang/c3c/tree/master/test/unit/stdlib/collections](url)/linked_map.c3 all pass,but,
when I tried to fetch 
```
hash_map_copy.tkeys() ;
```
and 
```
hash_map_copy.tvalues();
```
I got empty result set.Was I missing something?

which means below codes prints nothing:
```
    usz index=0;
    hash_map_copy.@each(;String key,usz value)
    {
    
        io::printfn("key=%s\tvalue=%s",key,value);
        index++;
    };
```

velikoss:
Could you send whole code for reproduction? Thanks!

LukyGuyLucky:
```
import std::collections::list;
import std::collections::map;
import std::sort;
import std::io;

alias printf=io::printf;
alias printfn=io::printfn;
alias print=io::printn;

alias TestLinkedHashMap=LinkedHashMap{String,usz};

fn void main()
{
   
    linked_map_copy();
  
}

fn void linked_map_copy()
{
    TestLinkedHashMap hash_map;
    hash_map.tinit();

    hash_map.set("aa",1);
    hash_map.set("b",2);
    hash_map.set("bb",1);
    printfn("hash_map.tkeys()=%s",hash_map.tkeys());
    printfn("hash_map.tvalues()=%s",hash_map.tvalues());
    TestLinkedHashMap hash_map_copy;
    hash_map_copy.tinit_from_map(&hash_map);

    printfn("(hash_map_copy.len()==hash_map.len())=%s",hash_map_copy.len()==hash_map.len());
    
    /* No!!
    foreach(i,ele:hash_map_copy.tkeys())
    {
        printfn("#%s\t%s",i,hash_map_copy.get(ele))!!;
    }
    */
    
    print("\n");
    
    //all three below ok
    printfn("hash_map_copy.get(\"aa\")=%s",hash_map_copy.get("aa"))!!;
    printfn("hash_map_copy.get(\"b\")=%s",hash_map_copy.get("b"))!!;
    printfn("hash_map_copy.get(\"bb\")=%s",hash_map_copy.get("bb"))!!;
    
    print("\n");
    
    //but these two are not right
    printfn("has_map_copy.tkeys()=%s",hash_map_copy.tkeys());
[...]

PR #2334, opened 2025-07-26
LinkedHashMap: Fix head being null when initializing from map

velikoss:
Fix for the #2332 
The issue was that `put_all_for_create` (which `init_from_map` uses) just inserted into the table without any linking logic
Therefore although `.get()` worked, because its traversing in the table, `.@each()` and `.values()` did not.

So my solution is to change `put_all_for_create`. Because we are receiving LinkedHashMap we just could iterate using `@each` while saving the correct order and call `map.set(ket, value)` to store values properly with linking logic


lerno:
Add a test too, proving it is fixed, that's always good practice.

velikoss:
i accidentally pinged person with nickname `@each` when commiting. oops

lerno:
Thank you!

PR #2335, opened 2025-07-26
Support full paths with `$embed`

NotsoanoNimus:
_Call for c3c Windows testers_: Please try out both `\path\to\file` and `C:\path\to\file` (and perhaps another drive letter from a different mount), and let me know if it works properly!

-----

Adding support for full-path file inclusion with the `$embed` CT statement. This came up because I had a few paths to well-known test files under my `/tmp/` folder and I got sick of using symlinks and/or `$embed("../../../../../../../../../tmp/file")` as a cheap hack. :)

Test:
```c++
fn void main()
{
    char[] u = $embed("/path/to/test/int/misc/test_input.data");
    char[] v = $embed("lib/std/atomic.c3");

    char[] w = $embed("/path/to/test/int/misc/test_input.data", 456);
    char[] x = $embed("lib/std/atomic.c3", 200);

    io::printfn("Bytes: %d and %d // %d and %d", u.len, v.len, w.len, x.len);
}
```

Result:
```
> c3c compile-run main.c3
Bytes: 270008320 and 19964 // 456 and 200

> ls -al /path/to/test/int/misc/test_input.data lib/std/atomic.c3
-rw-r--r--. 1 zpuhl zpuhl 270008320 Jun 26 22:49 /path/to/test/int/misc/test_input.data
-rw-r--r--. 1 zpuhl zpuhl     19964 Jul 24 17:48 lib/std/atomic.c3
```

### DIFF
--- a/releasenotes.md
+++ b/releasenotes.md
@@ -30,6 +30,7 @@
 - '$assignable' is deprecated.
 - Deprecate allocator::heap() and allocator::temp()
 - Add `thread::fence` providing a thread fence.
+- Allow absolute paths for `$embed`.
 
 ### Fixes
 - mkdir/rmdir would not work properly with substring paths on non-windows platforms.
--- a/src/compiler/sema_expr.c
+++ b/src/compiler/sema_expr.c
@@ -9984,7 +9984,7 @@ static inline bool sema_expr_analyse_embed(SemaContext *context, Expr *expr, boo
 	const char *string = filename->const_expr.bytes.ptr;
 	char *path;
 	char *name;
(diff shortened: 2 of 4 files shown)


lerno:
Cool, we just need the Win32 test then.

NotsoanoNimus:
Added an update for Win32 from testing/suggestions by @TechnicalFowl. Everything here should be g2g now @lerno

lerno:
As a possible extension: `$PROJECT/resources/foo` could also be nice.

PR #2348, opened 2025-07-29
Fix get_config_dir() implementation

BWindey:
There is an issue with the current `get_config_home()` implementation, as it puts `".config"` behind `$XDG_CONFIG_HOME`, but that is not needed (see the spec: https://specifications.freedesktop.org/basedir-spec/latest/).

I also noticed that only config-home is implemented, but the spec also defines data-home, state-home, cache-home and runtime-dir (less interesting than the previous). Do I implement those too?

### DIFF
--- a/lib/std/os/env.c3
+++ b/lib/std/os/env.c3
@@ -79,7 +79,7 @@ fn String? get_home_dir(Allocator allocator)
 
 
 <*
-Returns the current user's config directory.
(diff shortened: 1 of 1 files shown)


lerno:
We should probably do a full overhaul of this so for now a fix is fine

lerno:
Thank you!

Issue #2354, opened 2025-07-31
Results of `math` Functions Should be Assignable as Constants


NotsoanoNimus:
This may be a difficult issue due to the way the `@LinkMath` attribute works (I literally have no idea), but I'm trying to do the following:

```
module some_module { F };

const A_CONST = 12 + 2 * math::log2(F / 25);
```

which results in:
```
16: const MAX_ROUNDS = 12 + 2 * math::log2(F / 25);
                                ^^^^^^^^^^^^^^^^^^
(localtesting.c3:16:29) Error: Cannot call macro with '@links' outside of a function.
```

It seems this should be easily assignable into a constant value given the context.

lerno:
This should be fixed now, please try it.

NotsoanoNimus:
Fixed, thank you

PR #2355, opened 2025-07-31
Add `@intlog2` for Floored CT log-base2

NotsoanoNimus:
Need a quick way to get the `lg2` of a value at compile-time. This will return a value of `floor(log2(x))`, when `x` is evaluated as an _integer_ (not a floating point value).

Use-case:
```
module some_module { F };

const SOME_CONST = 12 + 2 * math::@intlog2(F / 25);
```

Some values:
```
LOG2[0]: 0
LOG2[1]: 0
LOG2[2]: 1
LOG2[3]: 1
LOG2[4]: 2
LOG2[5]: 2
LOG2[6]: 2
LOG2[7]: 2
LOG2[8]: 3
LOG2[9]: 3
LOG2[10]: 3
LOG2[11]: 3
LOG2[12]: 3
LOG2[13]: 3
LOG2[14]: 3
LOG2[15]: 3
LOG2[16]: 4
LOG2[17]: 4
```

-----

See #2354 for the background on why this came up.


TechnicalFowl:
This will loop infinitely for any integer above `0x80000001`

NotsoanoNimus:
> This will loop infinitely for any integer above `0x80000001`

Thank you! Most recent push should fix this as I've changed it to use a bit shift instead, and I added a `SLOW_TESTS` compile-time unit test as well.

PR #2359, opened 2025-08-01
Pair and Triple Compare w/ Unit Tests

NotsoanoNimus:
Quickly compare the contents of a Pair or Triple type. This delegates comparison work to each contained type's own `==` method.

Also wrote unit tests for these collection types.

-----

_NOTE: Needed to add the `==` operator for some time-related stdlib constructs for this as well, because `==` gets feisty about member types having the overload, even when you never use `==`._


NotsoanoNimus:
@lerno While I'm here, should I add the `[]` operators to get Pair/Triple items by an index? Or do you think that might get a little counterintuitive?

lerno:
I removed `==` from Time, because that should work out of the box. It's a bug if it doesn't. [] would only work if all are the same type, so that doesn't seem reasonable?

PR #2360, opened 2025-08-01
Stdlib: implement SingleSizeObjectPool

LowByteFox:
Adds Object Pool `SingleSizeObjectPool` to the stdlib, allows users to quickly allocate/reserve objects from the pool and to quickly deallocate/return objects back to the pool, adds module `std::core::mem::objectpool` and tests.

Example usage:
```c
fn int main(String[] args)
{
	SingleSizeObjectPool pool;
	pool.init(mem, int);
	int *obj = pool.alloc();
	// do something with obj
	pool.dealloc(obj);
	return 0;
}
```

### DIFF
--- a/lib/std/core/mem_objectpool.c3
+++ b/lib/std/core/mem_objectpool.c3
@@ -13,8 +13,7 @@ struct SingleSizeObjectPoolNode {
 }
 
 union SingleSizeObjectPoolEntry {
-	SingleSizeObjectPoolEntry *previous;
-	void *data;
+	void *previous;
 }
 
 struct SingleSizeObjectPool {
@@ -23,7 +22,7 @@ struct SingleSizeObjectPool {
 	SingleSizeObjectPoolNode *tail;
 	void *next_free;
 	void *freelist;
-	usz object_size, grow_size;
+	usz object_size, grow_capacity;
 	bool initialized;
 }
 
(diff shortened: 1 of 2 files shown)


lerno:
It could be interesting to track:
1. The total memory allocated by the pool
2. The total memory in use (or free either works)

That would also help testing it for correctness.

Secondly, it would be interesting to have a *concurrent* version of the same. That might be something to consider in the future.

LowByteFox:
> 1. The total memory allocated by the pool

I'm guessing to keep track of the buffer size and extra nodes (in bytes), right? Also could in theory try to allocate memory once and have it store both the node and the buffer, not sure if I can do something like flexible array member like in C, but then I suppose there will be alignment issues, unless I would align it to `usz`?

> 2. The total memory in use (or free either works)

I'm assuming this should also be in bytes?

> Secondly, it would be interesting to have a _concurrent_ version of the same. That might be something to consider in the future.

I'm wondering how this could be done than maybe just protecting it with a mutex? I don't know, maybe some discussion could be had and implemented in the future.

lerno:
Well, given that we allocate them a slab at a time, we only need to see how many slabs we have created. Then the amount free is the number of elements in the pool.

lerno:
The simplest way is just protecting it with a mutex. That could be a different implementation, so no need to reuse exactly this one.

lerno:
Thank you for this.

PR #2364, opened 2025-08-03
Added `dl_iterate_phdr`

sudokit:
Added `dl_iterate_phdr` and `dl_phdr_info` from signal.h. Tried following the naming convention from stdlib as well as possible, but let me know if it needs fixing. Also wondering if it would be better to handle 32 and 64 with `@if`, like the signal.h does with `ElfW`. For now did it like `Elf*_(P/E)hdr` was done.


lerno:
I think this looks ok. It's not breaking anything, so I'll just accept it as is. If there is anything wrong with it we can update it later.

PR #2367, opened 2025-08-04
Add `@clz` CT macro

NotsoanoNimus:
Quickly added a compile-time equivalent to `$$clz` or `[numeric_type].clz()` with a handful of tests. Feel free to add on as needed.


lerno:
Thank you for the addition!

PR #2372, opened 2025-08-05
improve & fix libc termios bindings

Book-reader:
* fixed some constants using C octal literals that had incorrect values in C3 (eg: all the `Speed` constants and many of the `Tcflags`)
* Moved all `Tcflags` to bitstructs
* moved some other constants to const enums
* changed the naming of termios methods to be snake_case

after doing some testing with the termios flags things seem to work fine.

### DIFF
--- a/lib/std/core/builtin.c3
+++ b/lib/std/core/builtin.c3
@@ -18,10 +18,10 @@ import libc, std::hash, std::io, std::os::backtrace;
 macro foo(a, #b = EMPTY_MACRO_SLOT)
 {
 	$if @is_valid_macro_slot(#b):
-        return invoke_foo2(a, #b);
-    $else
-        return invoke_foo1(a);
-    $endif
+		return invoke_foo2(a, #b);
+	$else
+		return invoke_foo1(a);
+	$endif
 }
 *>
 const EmptySlot EMPTY_MACRO_SLOT @builtin = null;
@@ -57,6 +57,10 @@ faultdef TYPE_MISMATCH @builtin;
  Use `CAPACITY_EXCEEDED` when trying to add to a bounded list or similar.
 */
 faultdef CAPACITY_EXCEEDED @builtin;
+/*
+ Use `NOT_IMPLEMENTED` when something is conditionally available.
+*/
+faultdef NOT_IMPLEMENTED @builtin;
 
 alias VoidFn = fn void();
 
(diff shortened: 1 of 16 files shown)


lerno:
We need to retain the old names for backwards compatibility, but please annotate them with `@deprecated` and move them to the end of the file.

Book-reader:
I've added operator overloads that depend on #2374 with the old constants to have almost full backwards compatibility (it won't work if someone stored the flags in a `Tcflags` variable and assigned it directly, but it should work everywhere else. let me know if I should change it to use bitstruct constants instead), and I have changed the constants to be octal literals so they have the correct value.


something like this (example from my old project):
```zig
term.c_lflag &= ~(termios::ECHO | termios::ICANON | termios::ISIG | termios::IEXTEN);
term.c_iflag &= ~(termios::IXON | termios::ICRNL | termios::BRKINT | termios::INPCK | termios::ISTRIP);
term.c_oflag &= ~(termios::OPOST);
term.c_cflag |= (termios::CS8);
```
will work properly (after the operator overload bug is fixed), and behave exactly the same as this:
```zig
term.c_lflag &= ~(Tc_lflags){.echo, .icanon, .isig, .iexten};
term.c_iflag &= ~(Tc_iflags){.ixon, .icrnl, .brkint, .inpck, .istrip};
term.c_oflag.opost = false;
term.c_cflag.csize = CS8;
```

Book-reader:
I've changed it so the deprecated constants use bitstructs instead of relying on operator overloading that shouldn't work.
the example I gave in the above comment still works

lib/std/libc/os/posix.c3
```diff
@@ -105,7 +105,7 @@ bitstruct Tc_cflags : CUInt
 {
 	T_csize csize : 4..5;
 	bool cstopb : 6;
-	bool creadb : 7;
+	bool cread : 7;
```
lerno, line 108: Is this change correct?

ManuLinares: This seems to be correct

https://www.man7.org/linux/man-pages/man3/termios.3.html#:~:text=cread
https://sourceware.org/glibc/manual/latest/html_mono/libc.html#:~:text=tcflag_t%20CREAD

Book-reader: it is correct, when I was copying the flags from the termios headers on my system I must have started copying `cread` then read the last letter of `parenb` on the line below as a continuation of it by accident.
I have checked through the other ones and I haven't seen any more typos.

lerno:
Thank you!

PR #2376, opened 2025-08-06
Add `bitsizeof` to Builtins

NotsoanoNimus:
There are countless places where the pattern `$sizeof(something) * 8` is used to express the conversion of a type's byte-width to bits.

Most languages usually have a flavor of this macro in order to make conversions readily apparent and readable by maintainers.

More unit tests are welcome for corner cases, but it's rather straightforward.

### DIFF
--- a/lib/std/core/builtin.c3
+++ b/lib/std/core/builtin.c3
@@ -85,6 +85,10 @@ macro void @swap(#a, #b) @builtin
 	#b = temp;
 }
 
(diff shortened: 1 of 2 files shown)


lerno:
Looks good. Add something in the releasenotes for it.

lerno:
Multiplying by "8" will make it signed, so I switched to 8u and also set the type of the macro, so that it's clear what the return type is.

PR #2381, opened 2025-08-07
Fix typo in bf16 suffix parsing logic

sipekdan:


lerno:
Looks correct, thank you!

Issue #2128, opened 2025-05-11
`attrdef` with any invalid name causes compiler assert


Book-reader:
```zig
attrdef @FOO = @test;
attrdef 1 = @test;
attrdef @foo = @test;
attrdef foo = @test;
attrdef $foo = @test;
attrdef #foo = @test;
attrdef "foo" = @test;
attrdef 'f' = @test;
```
any of these lines will cause the compiler to assert with the message:
```
⚠️ The compiler encountered an unexpected error: "Violated assert: context->tok == token_type".

- Function: advance_and_verify(...)
- Source file: /build/3k5w91avcbags2mxvfwlv5gmw2z39nl9-source/src/compiler/compiler_internal.h:2914

🙏 Please consider taking the time to file an issue on GitHub, so that we can get it fixed:

https://github.com/c3lang/c3c/issues/new so that we can get it fixed.
```

```
C3 Compiler Version:       0.7.2 (Pre-release, Jan  1 1980 00:00:00)
Installed directory:       /nix/store/2nvbz1b1nba9jl7cxkk8wf11avzl6x7p-c3c-0.7.2/bin/
Git Hash:                  unknown
Backends:                  LLVM
LLVM version:              19.1.7
LLVM default target:       x86_64-unknown-linux-gnu
```
c3c commit hash is `fc5c70a6283dae1e659ba2fe87dca69c00160510`

lerno:
This should work properly now.

Book-reader:
Thank you!

PR #2385, opened 2025-08-09
Create an issue template for the bug issue

ArtemkaKun:
Hi,

I added an issue template for the bug issue so that a person who files a bug has a better UX. This will also help C3 developers to have better and more structured bug reports that should speed up the fix process and increase C3 developers' work comfort.

The new bug issue filing form will look like the following after merging this change:
2025-08-09 12 46 27 github com 5ceafa4e404b

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,100 @@
+name: "🪲 Bug Report"
+description: "Use this template to report a bug."
+labels: ["Bug"]
+body:
+    - type: "markdown"
+      attributes:
+          value: |
+              > [!IMPORTANT]
+              > To save time for both you and our development team, we highly encourage discussing your bug on our Discord
+              server in the **#bugs** channel before submitting a new issue.
+              > This step helps in understanding the issue and determining if it's a bug at all.
+              Only proceed with this report if your issue remains unresolved after the Discord discussion.
+              >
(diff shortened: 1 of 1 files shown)


BWindey:
For me that's way way too complicated.
It's nice, but kinda assumes that the reporter has like no experience with submitting issues. Most issues submitted are from frequent submitters.

Maybe this could be something that people can choose to use, and not forced?

.github/ISSUE_TEMPLATE/bug_report.yml
Book-reader, line 1: Seems ok but is way too long. The mentions of discord should probably be removed since not every bug should be discussed there. Possible workarounds could also be removed with a comment on mentioning possible workarounds in the extra information section.
Version & environment details could be merged into asking for just the entire output of `c3c --version` as that has most relevant information (you can tell os from the installed path starting with `C:` vs `/bin` vs `/Library`)

Many first time bug reports are from a compiler crash/assert which could have a seperate issue template that asks for a minimal repro, the error message, and c3c version (eg https://github.com/c3lang/c3c/issues/2128) but that might not be in the scope of this pr

Book-reader: mentions of the discord would probably be better for a feature request issue template instead since that's what a lot of the discussion on the discord is about

ArtemkaKun:
While all your points are valid and I would try to discuss those, unfortunately, I found that I have no time for a new project right now.

PR #2386, opened 2025-08-09
Add HashSet and String methods

sandervdbosch:
Added the `HashSet.values` and HashSet.tvalues` method to export the HashSet to an array.
Added the `String.contains_char` method.


lib/std/collections/hashset.c3
```diff
@@ -329,7 +329,26 @@ fn void HashSet.reserve(&set, usz capacity)
 	}
 }
 
+fn Value[] HashSet.tvalues(&set)
+{
+	return set.values(tmem) @inline;
+}
```
Book-reader:
```suggestion
fn Value[] HashSet.tvalues(&self) => self.values(tmem) @inline;
```
just a small style nitpick, otherwise lgtm.

Book-reader (changes requested):

lerno:
Looks good, but can you add a note in releasenotes.md and add unit tests for the methods too?

sandervdbosch:
Unit tests have been added, I expanded the existing test a bit while I was working on the HashSet tests.

lerno:
Thank you!

PR #2387, opened 2025-08-09
Add/Fix OpenBSD `native_cpus`

NotsoanoNimus:
Added `sysctl` and related identifiers. Fix `native_cpus` to actually return the right information.

Tested on OpenBSD 7.7.


LowByteFox:
Good catch! I didn't even notice it slipped through, that being said I mostly just wanted C3C to compile

lerno:
Thank you

PR #2393, opened 2025-08-11
Install scripts

EchoPouet:
Hello, 

I am proposing two scripts to easily install C3 lang on Windows and Linux/MacOS. The scripts are based on the [Pixi](https://github.com/prefix-dev/pixi/tree/main/install) installation scripts.

At this point, I assume that the scripts are located in the “install” folder of the main branch after the pull request has been merged.

I cannot test the scripts on MacOS and OpenBSD, but they will be tested before merging.

I also modify the Readme file to integrate these scripts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,7 +209,7 @@ This installs the latest prerelease build, as opposed to the latest released ver
 3. If you don't have Visual Studio 17 installed you can either do so, or run the `msvc_build_libraries.py` Python script which will download the necessary files to compile on Windows. 
 4. Run `c3c.exe`.
 
-#### Installing on Windows with an install script
+#### Installing on Windows with the install script
 
 Open a PowerShell terminal (you may need to run it as an administrator) and run the following command:
 ```bash
@@ -232,7 +232,7 @@ If you don't have Visual Studio 17 installed you can either do so, or run the `m
 2. Unpack executable and standard lib.
 3. Run `./c3c`.
 
-#### Installing on Debian using the install script
+#### Installing on Debian with the install script
 
 Open a terminal and run the following command:
 ```bash


ManuLinares:
this is bad, hard to maintain, assumes a binary in the releases with the name `c3-debian.tar.gz` when the hard coded checks detect Linux but unknown `/etc/issue`. This binary doesn't exist.

EchoPouet:
> this is bad, hard to maintain, assumes a binary in the releases with the name `c3-debian.tar.gz` when the hard coded checks detect Linux but unknown `/etc/issue`. This binary doesn't exist.

Thanks for your constructive feedback.
Don't worry about maintenance. The only thing to keep me up to date will be adding new versions for different OS. The rest shouldn't have changed.

As for the release names “debian” or “ubuntu-22”, I didn't choose them.

I take note that “/etc/issue” is not necessarily already installed.

ManuLinares:
> Thanks for your constructive feedback. Don't worry about maintenance. 

The problem isn't just /etc/issue or the release naming. The whole approach assumes things about OS detection, archive naming, and installation layout that simply aren't true for most environments. This isn't a "just keep adding OS cases" situation; it's a maintenance nightmare because this isn’t just "a script that works with minor fixes", it's a fragile OS guessing game tied to assumptions about release packaging that will break often and be a pain to maintain.

I don't agree with this approach, but if it is prefered, I could write a simpler bash script that could cover _most_ cases.

waveproc:
[https://github.com/BWindey/c3vm](https://github.com/BWindey/c3vm) covers linux and macos. Then you need a script for Windows and other platforms. 

I think using a real language like python or rust is better than bash; it would be easier to maintain, and you could support all the platforms in one script. Adding the version manager script to your PATH automatically would be easy with Rust's package manager cargo. Just look at all the platforms Rust supports: [https://doc.rust-lang.org/nightly/rustc/platform-support.html](https://doc.rust-lang.org/nightly/rustc/platform-support.html) (note: cargo isn't necessarily supported on all of those platforms, but Rust is). Python platform support: [https://peps.python.org/pep-0011/](https://peps.python.org/pep-0011/). You don't even need to use cargo; you could distribute a binary installer, an apt package, etc, but cargo would handle that complexity for you.

> Don't worry about maintenance.

Developers should care about maintainability in my opinion.

EchoPouet:
I removed the distribution check and now use the “linux” version for all distributions.

@waveproc Thank you for your link. c3cm does more than my scripts. I only offer a single-use installation program, not a C3 version manager.
I agree with you, bash is difficult to maintain, but it is installed in 90% of Linux distributions and has been stable for some time.

Python can be a good alternative, but you need to have the right version, perhaps by installing an external package in a virtual environment for the latest version of Python. That's already too restrictive.

You mention the Rust language, and to install it on Linux, it offers “rustup.” Guess how you install it? You have to use this shell script “https://sh.rustup.rs” with 877 lines of code 😄

You can also create a C3 program based on c3cm or rustup, but you will need a bash or shell script to install the correct binary on your platform. All this to avoid bash. It's a miss 🫤 

I'm not promoting the bash language, I'm just being pragmatic, and my script only has 144 lines of code and is fairly simple. The powershell script is more complicated, I admit.

> In my opinion, developers should be concerned about maintainability.

I agree, and I suppose I misused it at that point.

I think the code shouldn't change much unless architectures are added in the releases or bugs of course.

waveproc:
@EchoPouet rust-all is available as an apt package which includes cargo, if you wish to avoid the shell script:
```
$ apt info rust-all
Package: rust-all
Version: 1.84.0ubuntu1
...
```
There are many high quality Rust programs available, so it wouldn't hurt to have someone install cargo on their system IMO. For example, I have [buck2](https://github.com/facebook/buck2), [jj VCS](https://github.com/jj-vcs/jj), [hyperfine](https://github.com/sharkdp/hyperfine), [ripgrep](https://github.com/BurntSushi/ripgrep). And we can support Windows, all of the Linux distros, freebsd, openbsd, redox, haikuOS, etc. from one script that is written in one language. When the script is tested and complete, we can put it in the C3 installation directions as an optional version manager for convenience. That's peak maintainability.

If you wanted to use Python, there's the [uv](https://docs.astral.sh/uv/) package manager. The default shell (bash) and package managers for Linux distros aren't necessarily set up in a way that's easy to use, or easy to maintain. OpenBSD uses ksh; I haven't heard of it until now.

Currently, all you have to do to install c3c on Linux is download the binary release, extract, and symlink it to a directory that's in your PATH. Pretty simple.

> You have to use this shell script “[https://sh.rustup.rs]"(https://sh.rustup.rs) with 877 lines of code

[left-pad](https://github.com/left-pad/left-pad/blob/master/index.js) has 47 lines of code.
Screenshot_20250813_214526

> The powershell script is more complicated, I admit.

Powershell supports Windows XP and Vista, just have to install it first. However, Rust can be made to work on older Windows versions, too: [https://github.com/felixmaker/thunk](https://github.com/felixmaker/thunk)

EchoPouet:
I never install Rust or Cargo except one day to test the language. I use the native package manager to install hyperfine.
I understand that Rust is THE solution. So if I code a program in Rust to install C3, I have to compile it for each platform and architecture.
Or make a new bash script that download, build and run this program on the host PC, but first I have to install Rust with Cargo. I also have to be careful to manage all the different package managers and names and find another way to do it on Windows. I don't agree with forcing the installation of a large number of packages to solve a small problem. Why make things complicated when you can make them simple?

As for Python, it doesn't install natively on Windows.

The only solution to meet your criteria, which are :
- No bash
- No powershell 
- A single file for maintainability
- Works everywhere
Is to write a program in C/C++ and use [cosmopolitan](https://github.com/jart/cosmopolitan) to compile it with its dependencies. But it’s not easy.

In addition, I offer 2 small scripts to install an archive from a url. Don't talk about a version manager, because that's not what I'm proposing and you don't have to use it.

In the end, it's @lerno who decides, and it's okay if he doesn't want it.

@waveproc Do you compare the number of lines in one code file with the number of lines in the entire bash interpreter?

waveproc:
> Why make things complicated when you can make them simple?

Agree with this. Look at just how easy it is to install Rust and its tooling, and start pulling in third party packages with Cargo:
- Go to https://www.rust-lang.org/
- Click "Get started"
- Run a bash script (the site automatically detects I'm using Linux) which installs everything and lets you update the Rust version, and lets you go back to older versions. There's also a link to other installation methods if you don't trust a bash script.

Why Rust and not another language like C3? Because C3's ecosystem is incomplete at the moment.

I like the opinionated nature of C3. There's an emphasis on readability so you can easily create a mental model of what the code is doing underneath the abstraction layers. So, if we wanted the installation instructions to be more readable, IMO we would leave the manual instructions as is and offer convenience scripts that are written in a (good) systems programming language, because bash/powershell is hard to read. Let's minimize the amount of [context switching](https://en.wikipedia.org/wiki/Human_multitasking) developers auditing or contributing to the code have to do. Let's deprecate legacy code.

> As for Python, it doesn't install natively on Windows.

I agree, Python kinda sucks. But, installing python is pretty easy on windows with [uv](https://docs.astral.sh/uv/getting-started/installation/).

> Don't talk about a version manager

The version manager pattern is not just a convenience - it's essential.

EchoPouet:
Could you evaluate this PR, @lerno?

lerno:
If I understand correctly the benefits of the scripts is setting up paths? Because right now it's just taking the zip/tar and unpacking it.

Currently, ready made versions of `c3c` are available through package managers.

There is also an in-development version of a "c3up" to grab the latest and install in the current install.

It's also interesting I think to have scripts that make things even easier once it's downloaded.

Can you explain where these scripts fit into this?

EchoPouet:
You're right, scripts allow you to decompress an archive in the user's HOME directory and update the shell without doing it manually. It can do the same for c3up.

But that's not the most important part of the script. What's handy is that, depending on the PC you run it on, it will choose the right binary for you. For now, it only checks the OS, but when you will support multiple architectures (AMD64, AARCH64, RISC-V, etc.), it will be more complicated. Example with [Zig](https://ziglang.org/download/).

So yes, those of us who chat on programming language forums know how to get the right archive for the right PC. But that's not the case for everyone, even developers. I see this regularly.

It's not a crazy idea. To install “rustup” on Linux, you use the same kind of script because it requires programs that are already present or very common. It's actually quite common and avoids the nightmare of having to manage dozens of package managers on countless Linux distributions. If I have to choose between a future C3 Ubuntu package or c3up, I'll install c3up because I can get the latest updated version.

lerno:
I appreciate the effort you put into this @EchoPouet. I'd like to keep the old instructions how to download and just add the using the .sh script as a an alternative, rather than removing the instructions in the readme. I can rewrite this in the way I want it if that's ok?

EchoPouet:
You can change whatever you want. It's your project ;)

lerno:
Thank you!

EchoPouet:
Thank you too.

PR #2397, opened 2025-08-13
Fix an issue with c3c debug symbols on Nix

vssukharev:
Now we can properly use c3c debug build from nix!

### DIFF
--- a/lib/std/net/socket.c3
+++ b/lib/std/net/socket.c3
@@ -25,31 +25,54 @@ macro void @loop_over_ai(AddrInfo* ai; @body(NativeSocket fd, AddrInfo* ai))
 
 const Duration POLL_FOREVER = (Duration)-1;
 
-typedef PollSubscribes = ushort;
-typedef PollEvents = ushort;
+enum PollSubscribe : const ushort
+{
+	ANY_READ     = os::POLLIN,
+	PRIO_READ    = os::POLLPRI,
+	OOB_READ     = os::POLLRDBAND,
+	READ         = os::POLLRDNORM,
+	ANY_WRITE    = os::POLLOUT,
+	OOB_WRITE    = os::POLLWRBAND,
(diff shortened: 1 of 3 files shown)


Book-reader:
this doesn't seem to have worked for me, lldb only shows assembly when stepping through c3c and gdb shows this warning:
image
with the TUI saying "No source available". 
this is better than before when gdb would say "No debugging symbols found in c3c", but it looks like the build directory the debug symbols point to has been deleted

vssukharev:
So we need to somehow retain the debug source tree

vssukharev:
Ahhh, it's more complicated than I expected. We can build with `separateDebugInfo = true;` but it requires additional configuration on the system specifically.

Book-reader:
Oh wow that's bad, the only solutions the nixos wiki have other than installing a third-party program are to either enable debug info for all packages systemwide or to manually run a command every time you run GDB to set the search path.

We could always use a hack in the dev shell like `alias gdb = gdb --eval-command="set debug-file-directory ${c3c.debug}"` but I'm not sure what else can be done

lerno:
So this should not be merged or?

vssukharev:
For now it strives for a bit different approach, so no

vssukharev:
Got it working. Just set `separateDebugInfo` to `true` in derivation build. This is on user's responsibility of how to use this option, so I didn't fall into details in README

Book-reader:
still doesn't work for me, when running gdb with `NIX_DEBUG_INFO_DIRS='<c3c-debug.debug>/lib/debug'`  as the nixos wiki instructs, it can't find the source:
image

vssukharev:
I use `nixdebuginfod`, which aitomatically manages everything, so maybe try to use it. I've just enabled the service and added `set debuginfod on` in gdbinit.

vssukharev:
Finding debug symbols is weird but pretty powerful.

PR #2399, opened 2025-08-13
Add Freestanding OS Types

NotsoanoNimus:
Add types to runtime `env::` module for FREESTANDING builds.

Also fixed `WASM_NOLIBC` references. It was mixed between `env::WASM_NOLIBC` and without the module prefix, because it was labeled a builtin. (and its definition was missing parentheses)


lerno:
You can't remove `@builtin` because that would be a breaking change. You can create a FREESTANDING_WASM and change to use env:: with that one and then put a deprecation notice to use FREESTANDING_WASM on WASM_NOLIBC

NotsoanoNimus:
> You can't remove `@builtin` because that would be a breaking change. You can create a FREESTANDING_WASM and change to use env:: with that one and then put a deprecation notice to use FREESTANDING_WASM on WASM_NOLIBC

Makes sense, I completely missed this consideration. I will also restore the original logic of `WASM_NOLIBC` which was not written correctly (at least it seems so), because I also don't want to affect current software. Thanks for the feedback!

lerno:
Thank you.

PR #2400, opened 2025-08-13
Add `benchmarkrun` and `*.wasm` to .gitignore

NotsoanoNimus:
Noticed these files filling always waiting to be staged in my git view.


lerno:
Good idea, thank you!

PR #2403, opened 2025-08-14
Fix `$$str_hash` to use `a5hash` like `String.hash()`

NotsoanoNimus:
Per Discord-submitted bug by @Book-reader, `"some string".hash()` was not correctly matching with the compile-time builtin `$$str_hash("some string")`, because the builtin was still using the FNV hashing algorithm.

Implemented `a5hash` in the compiler, swapped out the builtin's hash call, and added/updated tests as appropriate.

Note that testing variations of the `seed` value for `a5hash` doesn't really matter in this case, since the compiler will only ever use a default seed value of `0` with each `$$str_hash` call. Theoretically, this optional seed value could be added to the builtin later as an optional parameter.


lerno:
Thank you!

PR #2405, opened 2025-08-15
string: use correct allocator in `replace`

konimarti:
`replace` accepts an Allocator but uses `mem` instead.


lerno:
Oh, good catch!

PR #2410, opened 2025-08-17
Fixes slicing with negative value error message - llvm_codegen_expr.c

ManuLinares:
Fixes printing negative value

Example:
```
fn void main()
{
    int[100] arr;
    usz a = 35;
    usz b = 36;
    arr[b:a - b];
}
```

output should be:
ERROR: 'Negative value (-1) given for slice length.'


lerno:
Thank you. I added a description to the release notes as well.

Issue #2391, opened 2025-08-10
Better compiler error message for missing $endif


waveproc:
Error:
`Error: Mismatched '}' found.`

repro, in main fn:
```
    $if $feature(PLATFORMWEB):
        raylibJsSetEntry(&gameFrame);
    $else
    while (!raylib::windowShouldClose()) {
        gameFrame();
    }
   // $endif should be here
}
```

chrischtel:
Would something like this be good?

```bash
 $ ./build/c3c compile -C main.c3 
 7:         gameFrame();
 8:     }
 9:    // $endif should be here
10: }
    ^
    Error: Expected '$endif' for '$if' started at line 3.

**Comment by lerno** [comment]
Can you verify this looks fine now?

**Comment by waveproc** [comment]
Yes, it looks fine. Here's the new error message:
```
 1: fn void main() {
 2:     $if $feature(PLATFORMWEB):
 3:     $else
 4: }
    ^
(/home/chad/programming/c3c/test_error_msg.c3:4:1) Error: Expected '$endif' for '$if'

1: fn void main() {
2:     $if $feature(PLATFORMWEB):
       ^^^
(/home/chad/programming/c3c/test_error_msg.c3:2:5) Note: The unterminated '$if' was here.
```

PR #2413, opened 2025-08-17
#2391: Improve missing $endif error message

chrischtel:
Improves the error message when encountering a missing `$endif`.


lerno:
Thank you.

chrischtel:
Thanks for the sema_note addition

Issue #2384, opened 2025-08-08
Compiler TODO hit when attempting to build a dynamic lib with an exported function referencing a const enum


TechnicalFowl:
Attempting to build a dynamic-lib or static-lib with an exported function that references a const enum will hit a compiler TODO while generating the headers.

```
The compiler encountered an unexpected error: "TODO reached".

- Function: header_gen_maybe_generate_type(...)
- Source file: C:\Data\c3c\src\compiler\headers.c:556
```

Minimal reproducer:

```c++

enum Foo : const int
{
    A,
}

fn void test(Foo f) @export
{

}
```

Version Info:
```
c3c.exe --version
C3 Compiler Version:       0.7.5 (Pre-release, Aug  8 2025 16:10:46)
Installed directory:       C:/Data/c3c/build/RelWithDebInfo
Git Hash:                  1d25197bfda25b75ce88727cea0ad7cf4ace8157
Backends:                  LLVM
LLVM version:              19.1.5
LLVM default target:       x86_64-pc-windows-msvc
```

lerno:
This should be fixed then?

lerno:
Looks like it works.

PR #2414, opened 2025-08-17
Fix crash when exporting functions with const enums #2384 

chrischtel:
Right now if you export a function that takes a const enum,
the compiler crashes with "TODO reached" in header generation.

see #2384 
minimal repro.:
```
enum Foo : const int
{
    A,
}

fn void test(Foo f) @export
{

}
```

This patch makes header_gen_maybe_generate_type fall back to the enum’s
underlying integer type, so header generation works instead of failing.

It’s ABI-correct since const enums lower to their base type anyway.  
This just fixes the crash for now; nicer typedef/constant output could
be added later if wanted.


lerno:
Yeah, better have it working. Thank you.

Issue #2398, opened 2025-08-13
Compiler unreachable reached when declaring a macro method with `@body` on builtin or slice/array type


vssukharev:
Unfortunately, I couldn't reproduce the problem in a sandbox manner but you can check it out on [bindgen.c3l](https://github.com/vssukharev/bindgen.c3l/tree/dc57da7e49b6e9a449d124a1fdcf918501923852) repository. The emitted message is:
```console
$ c3c run
⚠️ The compiler encountered an unexpected error: "Should be unreachable".

- Function: sema_trace_decl_liveness(...)
- Source file: /build/nwamb53gw34gaqm5qxrldy9i7amc507g-source/src/compiler/sema_liveness.c:637

🙏 Please consider taking the time to file an issue on GitHub, so that we can get it fixed:

https://github.com/c3lang/c3c/issues/new so that we can get it fixed.
```
I've written a couple of test cases, based on my discovery:
1. The bug itself is under [misc.c3:355](https://github.com/vssukharev/bindgen.c3l/blob/dc57da7e49b6e9a449d124a1fdcf918501923852/bindgen.c3l/src/misc.c3#L355) - you should uncomment this line: `macro CXToken[].@foreach(self; @body(/*...*/)) {}`
2. The rest are working test cases:
- [misc.c3:359](https://github.com/vssukharev/bindgen.c3l/blob/dc57da7e49b6e9a449d124a1fdcf918501923852/bindgen.c3l/src/misc.c3#L359)
- [translators.c3:6](https://github.com/vssukharev/bindgen.c3l/blob/dc57da7e49b6e9a449d124a1fdcf918501923852/bindgen.c3l/src/translators.c3#L6)

The most obfuscate thing here is that bug doesn't occur in all files, if you move the declaration to them individually, like on the translators.c3 example above!
List of *working* files: bg.c3, translators.c3, user.c3
List of *broken* files: bgstr.c3, check.c3, data.c3, err.c3, misc.c3, valgen.c3, visitors.c3, writers.c3

lerno:
What's the easiest way to trigger this? If I just enter the project and hit "build" I am not triggering it.

vssukharev:
You should uncomment the line in misc.c3 like I said above.

lerno:
I did so.

vssukharev:
Sorry for late response - was a bit busy. I will try to debug it then, probably, the issue may occur on NixOS specifically:
```
[...]

PR #2415, opened 2025-08-17
Temporary fix for #2398

vssukharev:
A possible solution for issue above. The reason why I propose it, is that it's pretty similar to the line on `sema_liveness.c:577`. Look at issue comments for details.


lerno:
This is actually the completely correct fix. Both function and macro methods will appear in that code, but macros should be skipped, and that's what's missing.

lerno:
Thank you!

PR #2417, opened 2025-08-18
Fix out of bounds access when stdout is empty

ma-ale:
In execute_stdout_to_buffer() when the process exits without printing to stdout the function would access the buffer to index -1


lerno:
Looks good, thank you!

PR #2419, opened 2025-08-18
Add array `@reduce`, `@filter`, `@any`, `@all`, & `@indices_of`

NotsoanoNimus:
Add another wave of new core `array` operations to the standard library.

In brief:
1. `@reduce` - apply a function iteratively to an array and store the result into an accumulator.
  a. `@sum` - terse way to sum up all array elements with `@reduce`.
  b. `@product` - terse way to multiply all array elements with `@reduce`.
3. `@filter` - creates a shallow copy of an array, containing only copies of elements which match a given predicate function.
4. `@any` - return `true` when _any_ element in the array matches a given predicate function.
5. `@all` - return `true` when _all_ elements in the array match a given predicate function.
6. `@indices_of` - underlies the `@filter` macro; returns the indexes of the array which match the given predicate function.

Please see the `test/unit/stdlib/core/array.c3` file for a detailed look at each function's unit testing. Also, these functions (like `@zip` ones) are compatible with `List` types and other iterables.


lerno:
Thank you!

PR #2421, opened 2025-08-19
fix Socket.get_option calling `setsockopt` instead of `getsockopt`

Book-reader:


lerno:
Turns out there was another error there. Because getsockopt should take an in-parameter for size.

PR #2433, opened 2025-08-24
Add generic `InterfaceList` type for storing values that implement a specific interface

Book-reader:
* adds the interfacelist, and makes the anylist an extension of it.
 * fixes `remove_if` and `retain_using_test` always passing an `any` pointing to an `any`  pointing to the allocated list element instead of just an `any` pointing to the allocated list element.


lerno:
Can you add some tests as well please?

lerno:
I'm pushing this to 0.7.6, to merge after the 0.7.5 release.

lerno:
Thank you!

PR #2436, opened 2025-08-26
Fix issues with missing LLD directory

Ashvith10:
This should resolve the issue with missing LLD directory in certain platforms like OSX, Guix and Nix.


lerno:
I think this is wrong? You're asking to merge it into 0.5.x?

PR #2438, opened 2025-08-29
Add LinkedList Operators and Update Tests

NotsoanoNimus:
Noticed recently that LinkedList is a bit crusty. Updated it with some operator overloads and some other quality-of-life methods.

It's probably not well-advised to use `foreach_r` on this structure as it would be quite expensive. I don't know of a way to explicitly override that behavior to start from the tail of the list.

Also updated the corresponding unit tests to use the more "standard" `test::` functions, as well as added `LinkedList` capability tests to some core `array::` functions.


lerno:
I changed it a bit to drop the random access, because that's hinting at a contract that's simply bad. So instead it's just returning an array_view which is that iterator, since it then can be used for general access. That makes it more obvious what's happening. The only thing I'm not happy about is that `array_view` is a bit long as a name

PR #2440, opened 2025-08-30
FixedBlockPool: fix buffer overflow and implmeneted asan poisoning

LowByteFox:
This PR fixes a buffer write overflow when de-allocating an object that went unnoticed here:
```c++
mem::set(ptr + FixedBlockPoolEntry.sizeof, 0xAA, self.block_size);
```
Adding `FixedBlockPoolEntry.sizeof` would overflow, `self.block_size` covers either the bigger type or the entry size.

Now when compiling with `--sanitize=address`, `ASan` poisons blocks that we don't want the user working with.


lerno:
Thank you!

PR #2447, opened 2025-08-31
Add `@iter` to provide `.iter()` if it exists

NotsoanoNimus:
This change depends on #2438 and will stay in Draft until that request has been reviewed.

Adds an `@iter` macro to array to provide the proper iterable type at compile-time based on whether the passed list has an `.iter()` method attached to it.

### DIFF
--- a/benchmarks/stdlib/collections/linkedlist.c3
+++ b/benchmarks/stdlib/collections/linkedlist.c3
@@ -1,0 +1,50 @@
+module linkedlist_benchmarks;
+
+import std::collections::linkedlist;
+
+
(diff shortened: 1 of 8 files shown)


lerno:
As you saw I removed `iter` from LinkedList. I'm thinking about possibly making it into an overload, which would make this moot.

So something like:
```c
fn Foo.iter() @operator(iterator)
{
  ...
}
```
And now the folding is automatic.

lerno:
With `lengthof` you can check for foreach using `$defined(f[0], lengthof(f))`

lerno:
I will close this one then.

PR #2448, opened 2025-09-01
Make `enum Cc` `inline` so it can index `Termios.c_cc`

Book-reader:
I forgot to make it inline, so it couldn't be used to index the termux `c_cc`array it was meant to without a cast. This fixes it.

### DIFF
--- a/lib/std/libc/os/posix.c3
+++ b/lib/std/libc/os/posix.c3
@@ -218,7 +218,7 @@ enum Speed : const CUInt
 	MAX_BAUD = B4000000,
 }
 
-enum Cc : const char
+enum Cc : const inline char
 {
 	VINTR = 0,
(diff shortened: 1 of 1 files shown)


Book-reader:
wait, I've just realised another mistake that I've made. I'll have to fix it as well

Book-reader:
Originally `typedef Cc = char` was used for the values of the `c_cc` array, and the constants used to index it were of type `Tcflags`. My changes made both the values of the `c_cc` array and the enum used to index it into the same type, which could be a bit confusing. I'm not sure if I should change this though.

lerno:
So... should it be in or not?

Book-reader:
yes, better to have it be able to index the array without a cast.

lerno:
Let's add it then!

PR #2457, opened 2025-09-03
Typo fixes

hyperpastel:
Several typos fixed across both the compiler and the standard library.

Typos fixed in:
- Comments
- Strings
- Symbols

Compilation with checks is successful on my ``x86_64-linux`` machine running NixOS.


lerno:
Thank you!

PR #2464, opened 2025-09-06
Build linux binaries for releases with proper curl support

### LINKED ISSUE #2460: vendor-fetch does not work for Ubuntu build
I downloaded the compiler binary for Ubuntu from the getting started doc: https://github.com/c3lang/c3c/releases/latest/download/c3-ubuntu-22.tar.gz

The vendor-fetch command fails with the following message:

```
c3c vendor-fetch raylib55                                                                                                                   
Error: vendor-fetch only available when compiled with cURL.
```

Of course curl is installed and working on my system so this looks like an issue with the build script.

Book-reader:
fixes #2460 and allows the linux binaries in releases to use `vendor-fetch`

I've tested this with the `build-with-docker.sh` script, but it should work the same for the github actions scripts as well

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -236,7 +236,7 @@ jobs:
       - uses: actions/checkout@v4
       - name: Install common deps
         run: |
-          sudo apt-get install zlib1g zlib1g-dev python3 ninja-build curl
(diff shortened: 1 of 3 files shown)


lerno:
Seems good!

PR #2467, opened 2025-09-06
Find cl and set INCLUDE env var outside of a visual studio command promt 

JoNil:
If we are running outside a visual studio command prompt find an appropriate cl and the set the INCLUDE env var.

This allowed me to build the https://github.com/floooh/sokol-c3 examples in a git bash terminal on windows.

I referenced https://github.com/rust-lang/cc-rs/tree/main/find-msvc-tools to figure out what includes to set

### DIFF
--- a/src/compiler/compiler.c
+++ b/src/compiler/compiler.c
@@ -1754,7 +1754,7 @@ const char *default_c_compiler(void)
 		return cc;
 	}
 #if PLATFORM_WINDOWS
-	return cc = windows_get_cl();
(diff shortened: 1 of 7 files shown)


JoNil:
The build error should be fixed now. But I'm not entirely sure if I'm doing the correct thing.

The vs_path for example is allocated by scratch_buffer_copy() so i can reference it when building paths with that as a base. I think that ends up on the char_arena so there is nothing that i should free?

lerno:
Thank you!

Issue #2427, opened 2025-08-21
enhancement request: allow doc comments on individual struct members


Book-reader:
it allows doc generators to include information from places like here if these were doc comments instead:
<img width="633" height="213" alt="Image" src="https://github.com/user-attachments/assets/2b8f4044-9742-45fb-b22b-b17306313d15" />

lerno:
How would it look like?

NotsoanoNimus:
One suggestion might look like:
```swift
<*
 A C struct for time and time things.

 @member tm_sec `Seconds after the minute [0-60]`
 @member tm_min `Minutes after the hour [0-59]`
...
*>
struct Tm
{
    CInt tm_sec;
    CInt tm_min;
    // ...
}
```

Book-reader:
I was thinking something like this:
```cpp
<*
 The libc time format.
 Describes time, broken down into distinct components.
*>
struct Tm
{
  <* Seconds after the minute [0-60] *>
  CInt tm_sec;
  <* Minutes after the hour [0-59] *>
  CInt tm_min;
  ...
}
```
It might be a bit easier to read the definition having the docs directly above each member compared to having to look at each `@member` to find the docs.
It also could be a bit simpler for a doc generator to parse since it would only need to check if there is a `doc_comment` before each member instead of having to parse the doc comment above the struct to look for them.

lerno:
Both are good points. How do we select one?

NotsoanoNimus:
Why not use both? I certainly agree with the suggestion on locality of the comment being important, but `@member` could be supported by the language for whatever it needs syntactically (if further `struct` contracts develop in the future).

Doc generators can be free to parse it as well as a `<* doc comment block *>` above each member or after it on the same line.

Now for "standardizing" how a doc generator should handle that, I'm not sure. Perhaps both locations could be checked, with the "inline" form overriding or appending to the `@member` text. Or maybe the "inline" is output by the generator first as a bold annotation and the `@member` text follows.

Book-reader:
Having doc comments above each member declaration would be consistent with how it's done for interfaces:
https://github.com/c3lang/c3c/blob/f7143c185248a48105b768d7983ef83250dc5628/lib/std/core/mem_allocator.c3#L33-L62

NotsoanoNimus:
[...]

PR #2473, opened 2025-09-11
Update stdlib to use struct member docs from #2427 and other small changes

Book-reader:
I've also moved many single-line comments in places like the libc docs to be doc comments and slightly reworded the docs in `siphash.c3`, let me know if you have any problems/feedback with the changes.

### DIFF
--- a/lib/std/hash/siphash.c3
+++ b/lib/std/hash/siphash.c3
@@ -1,11 +1,9 @@
 // Copyright (c) 2025 Zack Puhl <github@xmit.xyz>. All rights reserved.
 // Use of this source code is governed by the MIT license
 // a copy of which can be found in the LICENSE_STDLIB file.
-//
 
-// These provide typical 64-bit hash results.
 <*
- A common variant of SipHash that is widely used. Best for performance-critical applications.
+ Best for performance-critical applications.
  See std::hash::siphash for more information.
 *>
 module std::hash::siphash24;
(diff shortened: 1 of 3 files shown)


ManuLinares:
I prefer the comments be in the same line like:

```c
const int SO_TYPE         = 0x1008; /* get socket type */
const int SO_NETPROC      = 0x1020; /* multiplex; network processing */
const int SO_RTABLE       = 0x1021; /* routing table to be used */
```

or

```c
const int SO_TYPE         = 0x1008; <* get socket type *>
const int SO_NETPROC      = 0x1020; <* multiplex; network processing *>
const int SO_RTABLE       = 0x1021; <* routing table to be used *>
```

lerno:
What should we do with this?

Book-reader:
I'll revert the large changes on the constants/functions for now, but I think the other things are fine to merge

lerno:
What do you say @ManuLinares

ManuLinares:
> What do you say @ManuLinares

I still prefer more packed comments in the same line because I can easily see a struct members. For example I don't like this change:

image

This is about striking a balance in readability and style.

lerno:
There is not really any way to retain reasonable semantics for <* *> if end of line comments after `;` are treated as belonging to the previous line unless we make eol meaningful for the semantics. I mean theoretically it should be possible to place them behind and then add them I guess, because it's still not on the same line. Doubtful semantically but still.

lerno:
We can perhaps revisit this later, it would need some special line comment style if we had the comments to the right. And it's still possible to have the `//` comments, it's just that it's not picked up by doc parsers.

lerno:
Thank you @Book-reader for the PR and @ManuLinares for your thoughts on this. Let's perhaps continue the discussion in a new issue or on Discord.

PR #2475, opened 2025-09-11
Generic inference

lerno:


data-man:
IMHO, but `QuaternionNumber` isn't good name, sorry.   
What about `QuaternionType`?

lerno:
I didn't like QuaternionType, it felt too... uh... generic. I originally just went for Quatern. It's not really a type that should be used outside of the aliasing so it's a bit weird. It's consistent with ComplexNumber at least.

lerno:
Neither is intended to be used except through their aliases.

lerno:
@data-man If you have any other alternative I'm open to change QuaternionNumber to that.

data-man:
Some projects uses `Intl` or `Internal` suffixes for hidden types.   
So, maybe `QuaternionInternal` would be fine.

lerno:
We can fix the naming later perhaps @data-man, I'll merge this now, but feel free to post a PR with a better name

PR #2481, opened 2025-09-14
Add Gentoo installation instructions to README

chrischtel:
Added Gentoo installation instructions for c3c.

I am also maintaining the package


lerno:
Thank you!

Issue #2478, opened 2025-09-13
[Feature Request] Implement write_stdin() in std::os::process


EchoPouet:
Add the ability to write to a process's stdin (std::os::process).

This will be useful for sending commands to a terminal program, such as in my use case, a Docker container instance or SFTP communication.

Thank you

velikoss:
Created a PR regarding that

velikoss:
It's now merged into main. Have a look, @EchoPouet

EchoPouet:
Thank you very much @velikoss.

PR #2482, opened 2025-09-15
Implement write_to_stdin() in std::os::process

velikoss:
Implementing feature `write_to_stdin` described in #2478


lerno:
Looks good!

PR #2487, opened 2025-09-17
[STD LIB] Add macro return types

BWindey:
This PR adds return types to macros where they make sense in order to make them easier to read and understand at a glance. 

It also fixes a function-call with too many modules in front of it, and some stray spaces got pruned by my formatter (no actual lines were removed).


lib/std/core/builtin.c3
```diff
@@ -430,7 +430,7 @@ macro swizzle2(v, v2, ...) @builtin
  @require types::is_int($typeof($value)) : "Input value must be an integer"
  @require $sizeof($value) * 8 <= 128 : "Input value must be 128 bits wide or lower"
 *>
-macro @clz($value) @builtin @const
+macro usz @clz($value) @builtin @const
```
lerno: Change this to `uint` to make the need for casts less.

lib/std/core/slice2d.c3
```diff
@@ -33,7 +33,7 @@ fn usz Slice2d.count(&self)
 <*
  Step through each element of the slice.
 *>
-macro void Slice2d.@each(&self; @body(usz[<2>], Type))
+macro Slice2d.@each(&self; @body(usz[<2>], Type))
```
lerno: No need to remove the void

lib/std/core/slice2d.c3
```diff
@@ -47,7 +47,7 @@ macro void Slice2d.@each(&self; @body(usz[<2>], Type))
 <*
  Step through each element of the slice *by reference*
 *>
-macro void Slice2d.@each_ref(&self; @body(usz[<2>], Type*))
+macro Slice2d.@each_ref(&self; @body(usz[<2>], Type*))
```
lerno: Nor here

lib/std/core/slice2d.c3
BWindey: Oh, I must have stripped them when I stripped other changes I made. My bad.

lerno:
Thank you!

PR #2496, opened 2025-09-24
Add AES algorithm

konimarti:
This is an implementation of the AES algorithm with the ECB, CTR and CBC
modes. The key size can be chosen among AES128, AES192, AES256.

Ported from github.com/kokke/tiny-aes-c


lerno:
I've played around with the API a bit. I am assuming CTR is the normally used one, which is why I went and created some `aes128::encrypt` functions for the general case. What do you think?

lerno:
Thank you! Let's see if we can polish the API further, but for now I'm happy.

PR #2508, opened 2025-10-01
add Matrix4x4.rotate_axis + small fix

any1here:
Small fix so it is consistent with the other mul methods

Edit: Also added Matrix4x4.rotate_axis


lerno:
We're replacing the whole matrix library with this: https://github.com/m0tholith/c3math but it still has some way to go.

any1here:
Closed in favor of https://github.com/m0tholith/c3math/pull/5

PR #2511, opened 2025-10-04
Make test runner print logs using `std::core::log` of all levels

m0tholith:


lerno:
Should this not be added on a per-case basis?

lerno:
Maybe add a compile time argument for the test runner to set this?

lerno:
Are you working on an update?

m0tholith:
Master branch got polluted... I'll fix it on another PR

m0tholith:
Moved to #2560

PR #2529, opened 2025-10-14
fixed not compiling filenames with single char stem

### LINKED ISSUE #2528: Compiling source code with filenames that have a single character stem errors out with "Failed to split SHD"
When running `c3c compile` on a source code with single character stem eg: `a.c3`, the compiler errors out with the message `Failed to split SHD`. But it compiles fine when renaming the source code to 'ab.c3'.

Example:
```
$ cat a.c3
import std::io;

fn void main() {
        io::printn("Hello world");
}
$ c3c compile a.c3
Failed to split SHD
$ cp a.c3 ab.c3
$ c3c compile ab.c3
Program linked to executable './ab'.
$ ./ab
Hello world

```
The same thing happens with files with a single character module:
```
$ cat b.c3
module b;
import std::io;

fn void main() {
        io::printn("Hello world");
}
$ c3c compile b.c3
Failed to split SHD
$ cp b.c3 c.c3
$ cat c.c3
module bc;
import std::io;

fn void main() {
        io::printn("Hello world");
}
$ c3c compile c.c3
Program linked to executable './bc'.
$ ./bc
Hello world

```

giuliano-macedo:
Fixes #2528

There was an edge case for `file_namesplit`, where if `path` contains a single character after a relative directory (eg: './a'), the function was incorrectly
identifing as a single '.' and returning `false` thats because it should check for the cases `.` and `..` by checking the total path len instead of `file_len`. 

Also, in `file_create_folder` when the function `file_namesplit` the compiler was printing a function pointer of the function `filename` as a string on `error_exit`, so I changed to use path instead.

### DIFF
--- a/src/utils/file_utils.c
+++ b/src/utils/file_utils.c
@@ -197,6 +197,8 @@ bool file_namesplit(const char *path, char** filename_ptr, char** directory_ptr)
 {
 	size_t len = strlen(path);
 	if (len == 0) return false;
+	if (len == 1 && path[0] == '.') return false;
+	if (len == 2 && path[0] == '.' && path[1] == '.') return false;
 	size_t found_at = (size_t)-1;
 	for (size_t i = len - 1; i > 0; i--)
 	{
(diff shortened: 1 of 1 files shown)


giuliano-macedo:
This was solved on 7c8e3dd4f, my local branch was outdated, sorry.

PR #2530, opened 2025-10-14
Fixed incorrect format strings when using `error_exit`.

giuliano-macedo:
Fixed string formatting when using `error_exit` by adding [__attribute__((format))](https://clang.llvm.org/docs/AttributeReference.html#format) in the function declaration.

Also changed `FATAL_ERROR` message to be less repetitive ('so that we can get it fixed' was duplicated) and I've embed the string format inside the function call so that `TEST_ASSERTF` could be simplified.

### DIFF
--- a/src/utils/common.h
+++ b/src/utils/common.h
@@ -81,18 +81,35 @@
 #define UNUSED __attribute__((unused))
 #define NORETURN __attribute__((noreturn))
 #define INLINE __attribute__((always_inline)) static inline
+#define FORMAT_STR
+#define FORMAT(__X__, __Y__) __attribute__((format (printf, __X__, __Y__)))
 #elif defined(_MSC_VER)
 #define FALLTHROUGH ((void)0)
 #define INLINE static __forceinline
 #define NORETURN __declspec(noreturn)
 #define UNUSED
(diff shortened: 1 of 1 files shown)


lerno:
Looking good, just fix the `error_exit` cases there with an `error_exit(error)` to `error_exit("%s", error)`

giuliano-macedo:
@lerno Fixed! my bad, I compiled with clang20. A little weird that I'm not getting these error on 20 :thinking:

lerno:
It needs to compile with MSVC as well, so you need to define the attribute through a macro so that it does nothing on MSVC

giuliano-macedo:
@lerno Done! thanks! sorry for taking so long, I've made it compatible with msvc if the /analyze flag is enabled in the future by using [SAL Annotations](https://learn.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/ms235402(v=vs.100)?redirectedfrom=MSDN), more info on this [here](https://stackoverflow.com/questions/2354784/attribute-formatprintf-1-2-for-msvc)

PS: I don't have strong opinions on this, if you think using SAL would add unnecessary complexity I think it would be understandable.

lerno:
Thank you!

PR #2531, opened 2025-10-15
fix overriding with custom patches in the nix build

Book-reader:
The previous method of setting the correct date in the nix build was to override the `patchPhase` and substitute them there, the problem is that this will completely remove the ability to patch the compiler at build time through this method:
```nix
pkgs.c3c.overrideAttrs {
  patches = [ ./my_custom_c3c_patch.diff ];
}
```
following the advice at https://github.com/jtojnar/nixpkgs-hammering/blob/main/explanations/patch-phase.md I have moved the date substitution to the `postPatch` phase and tested that it still correctly sets the date at build time.
I have also added calls to the `preCheck` and `postCheck` hooks, which will also allow them to be overridden if needed.

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -9,13 +9,13 @@
   rev,
   debug ? false,
   checks ? false,
-}: let 
+}: let
   inherit (builtins) readFile elemAt;
   # inherit (lib.sources) cleanSourceWith cleanSource; 
   inherit (lib.lists) findFirst;
   inherit (lib.asserts) assertMsg;
   inherit (lib.strings) hasInfix splitString removeSuffix removePrefix optionalString;
-in llvmPackages.stdenv.mkDerivation (_: 
+in llvmPackages.stdenv.mkDerivation (_:
 {
   pname = "c3c${optionalString debug "-debug"}";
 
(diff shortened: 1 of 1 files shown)


lerno:
Do you have any suggestions who could review this?

Book-reader:
Maybe @vssukharev could give it a go since iirc he uses nix. Here's a simple patch that can be used to test that patches apply successfully with this change (and that they don't without):
```patch
diff --git a/src/main.c b/src/main.c
index b56ead3d..5a2d3696 100644
--- a/src/main.c
+++ b/src/main.c
@@ -28,6 +28,7 @@ const char *compiler_exe_name;

 int main_real(int argc, const char *argv[])
 {
+       printf("test\n");
        srand((unsigned int)time(NULL));
        compiler_exe_name = argv[0];
 #ifdef __OpenBSD__
```

vssukharev (approved):
Thank you for mastering up the nix build! Yeah, eveything is working and also fixes the previous problems.

lerno:
Then it's merged! Thank you for the fix and than you @vssukharev for helping to check it.

PR #2545, opened 2025-10-25
Dev

lerno:


ManuLinares:
~~This work is amazing, you telling me that I can typedef MyFloatx4 = float[<4>] @simd; and the compiler will use simd for all ops?~~

I just saw the stream where you talk about this-

lerno:
No vectors are always SIMD, it's just that `typedef MyFloatx4 = float[<4>] @simd;` would be stored with simd alignment and passed by as a SIMD vector in the API. And the innovation is actually that `float[<4>]` is passed and stored as an array.

PR #2554, opened 2025-10-30
Add $schema to project_default_keys

dazoe:
As noted in #2066 adding $schema to project.json causes compiler to emit a WARNING.
Adding $schema to avoid compiler WARNING about an unknown parameter.


lerno:
Thank you

PR #2557, opened 2025-11-01
Add default exception handler for win32

TechnicalFowl:
This adds a default UnhandledExceptionFilter to the stdlib for win32. This will be called when an exception (such as a seg fault) reaches the top-level unhandled exception filter and the application is not being debugged. Currently the application will just silently die in these cases necessitating breaking out the debugger but this will report a convenient backtrace.

It also registers a panic function to allow it to track if a panic has occurred to prevent it double reporting these when the panic function calls `$$trap()`.

Example:
```
import std::io;

fn void main()
{
    io::printfn("Foo");
    uint* p = (uint*) 0x13370;
    *p = 0;
    io::printfn("Bar");
}
```
Previous output:
```
Foo
```
New output:
```
Foo

ERROR: 'Access Violation at address 0x7ff6fe370f8a'
  in main (C:\Data\test\src\main.c3:52) [C:\Data\test\build\test.exe]
  in _$main (C:\Data\test\src\main.c3:47) [C:\Data\test\build\test.exe]
  in invoke_main (D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:78) [C:\Data\test\build\test.exe]
  in __scrt_common_main_seh (D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:288) [C:\Data\test\build\test.exe]
  in __scrt_common_main (D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:330) [C:\Data\test\build\test.exe]
  in mainCRTStartup (D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_main.cpp:16) [C:\Data\test\build\test.exe]
  in BaseThreadInitThunk (source unavailable) [C:\WINDOWS\System32\KERNEL32.DLL]
  in RtlUserThreadStart (source unavailable) [C:\WINDOWS\SYSTEM32\ntdll.dll]
  ```

### DIFF
--- a/lib/std/os/win32/exception.c3
+++ b/lib/std/os/win32/exception.c3
@@ -6,118 +6,118 @@ const EXCEPTION_MAXIMUM_PARAMETERS = 15;
 
 struct ExceptionRecord
 {
-    Win32_DWORD exception_code;
-    Win32_DWORD exception_flags;
-    ExceptionRecord* exception_record;
-    void* exception_address;
-    Win32_DWORD number_parameters;
-    Win32_ULONG_PTR[EXCEPTION_MAXIMUM_PARAMETERS] exception_information;
+	Win32_DWORD exception_code;
+	Win32_DWORD exception_flags;
+	ExceptionRecord* exception_record;
+	void* exception_address;
+	Win32_DWORD number_parameters;
+	Win32_ULONG_PTR[EXCEPTION_MAXIMUM_PARAMETERS] exception_information;
 }
 
 struct ExceptionContext @align(16)
 {
-    Win32_DWORD64 p1_home;
-    Win32_DWORD64 p2_home;
-    Win32_DWORD64 p3_home;
-    Win32_DWORD64 p4_home;
-    Win32_DWORD64 p5_home;
-    Win32_DWORD64 p6_home;
+	Win32_DWORD64 p1_home;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: lib/std/os/win32/exception.c3 [hunk 1]
```diff
@@ -0,0 +1,188 @@
+module std::os::win32 @if(env::WIN32);
+
+import std::io;
+
+const EXCEPTION_MAXIMUM_PARAMETERS = 15;
+
+struct ExceptionRecord
+{
+    Win32_DWORD exception_code;
```
lerno: Could you do a quick space -> tab pass on this one, so that it follows the same format as the rest of the stdlib?

lib/std/os/win32/exception.c3
```diff
@@ -0,0 +1,188 @@
...
+        };
+    }
+    if (previous_filter)
+    {
+        return previous_filter(exception_info);
+    }
+    return EXCEPTION_EXECUTE_HANDLER;
+}
+
+fn void panicTracker(String message, String file, String function, uint line)
```
lerno: Can you change this name style to the standard as well, so for this one it'd be `panic_tracker`

lerno:
Thank you!

PR #2560, opened 2025-11-04
Add test option `--test-log-level` to choose tests' log level

m0tholith:


m0tholith:
Made argument passed to `--test-log-level` case-insensitive (so: `debug`, `Debug`, `DEBUG` all work). Also made error messages more helpful.

m0tholith:
(bump @lerno)

src/build/build_options.c
```diff
@@ -784,6 +785,12 @@ static void parse_option(BuildOptions *options)
 				options->test_filter = next_arg();
 				return;
 			}
+			if (match_longopt("test-log-level"))
```
lerno: Better use `match_argopt` with the different options here.

lerno: You can see from the other examples how it works.

lib/std/core/runtime_test.c3
```diff
@@ -215,11 +218,36 @@ fn bool run_tests(String[] args, TestUnit[] tests) @private
 				}
 				context.test_filter = args[i + 1];
 				i++;
+			case "--test-log-level":
+				if (i == args.len - 1)
+				{
+					io::printn("Missing log level for argument `--test-log-level`.");
+					return false;
+				}
+				args[i + 1].convert_to_lower();
```
lerno: If you convert it to upper case instead, you can use `enum_by_name(LogPriority, args[i + 1])` and drop the switch.

lib/std/core/runtime_test.c3
```diff
@@ -215,11 +218,36 @@ fn bool run_tests(String[] args, TestUnit[] tests) @private
...
+				{
+					case "verbose": context.log_level = LogPriority.VERBOSE;
+					case "debug": context.log_level = LogPriority.DEBUG;
+					case "info": context.log_level = LogPriority.INFO;
+					case "warn": context.log_level = LogPriority.WARN;
+					case "error": context.log_level = LogPriority.ERROR;
+					case "critical": context.log_level = LogPriority.CRITICAL;
+					default:
+						io::print("Log level given to `--test-log-level` is not one of: ");
+						$foreach $level : LogPriority.names[:^1]:
```
lerno: Since LogPriority will not change, just write this as a hard coded description rather than grabbing the names.

lerno:
I'll look at this on Friday. Remind me if I forget.

lerno:
Thank you!

PR #2561, opened 2025-11-04
Make expected error in `test::@error` macro optional

m0tholith:


lib/std/core/test.c3
```diff
@@ -77,6 +79,33 @@ macro @check(#condition, String format = "", args...)
...
+ Check if function returns error
+
+ @param #funcresult : `result of function execution`
+ @require runtime::test_context != null : "Only allowed in @test functions"
+*>
+macro @catch(#funcresult)
+{
+	if (catch err = #funcresult) return;
+	print_panicf("`%s` unexpectedly did not return error.", $stringify(#funcresult));
+}
```
BWindey: Looks a lot like the `@error()`, would it be better to merge those two together?

m0tholith: My idea was that it would be similar to the normal `@ok`, `@catch`.

m0tholith:
Added the proposed changes.

m0tholith:
`foo()!!` is already an equivalent for `test::@ok(foo())`, so I removed that. I also cleaned up the empty-slot code in `@error`.

lerno:
Thank you!

Issue #1456, opened 2024-09-16
The big 1.0 


lerno:
I'm collecting small pieces of things needed for 1.0 that aren't really ready to go into issues yet into this single issue.

- [ ] Full BSD-family support in the standard library
- [ ] C-to-C3 tool available
- [ ] C3docs available
- [ ] Full standard library available on the web
- [ ] No known bugs in the compiler
- [ ] A secondary backend for fast compilations
- [ ] Automatic static / dynamic library .c3l generation
- [ ] A completely fleshed out standard library
- [ ] Support for most common embedded targets (when 1.0 is released!)
- [ ] Full Android / iOS support
- [ ] Downloadable binaries for all Linux distros, MacOS, Windows, all BSD variants.
- [ ] A good experience cross compiling to Linux
- [ ] Cross compiling to MacOS & Windows with automatic download of required supporting libraries

Nice to haves:
- [ ] A C output backend

Caleb-o:
Very exciting!

Real-Packet:
ZIP compression? *yoinks C# ZIP library implementation*

Real-Packet:
> I'm collecting small pieces of things needed for 1.0 that aren't really ready to go into issues yet into this single issue.
> 
>     * [ ]  Full BSD-family support in the standard library
> 
>     * [ ]  C-to-C3 tool available
> 
>     * [ ]  C3docs available
> 
>     * [ ]  Full standard library available on the web
> 
>     * [ ]  No known bugs in the compiler
> 
>     * [ ]  A secondary backend for fast compilations
> 
>     * [ ]  Automatic static / dynamic library .c3l generation
> 
>     * [ ]  A completely fleshed out standard library
> 
>     * [ ]  Support for most common embedded targets (when 1.0 is released!)
> 
>     * [ ]  Full Android / iOS support
> 
>     * [ ]  Downloadable binaries for all Linux distros, MacOS, Windows, all BSD variants.
> 
>     * [ ]  A good experience cross compiling to Linux
> 
>     * [ ]  Cross compiling to MacOS & Windows with automatic download of required supporting libraries
> 
>     * [ ]  High quality LSP
> 
>     * [ ]  Easily available editor plugins
> 
>     * [ ]  High quality IntelliJ plugin
> 
> 
> Nice to haves:
> 
>     * [ ]  A C output backend
> 
> 
> Stdlib features
> 
>     * [ ]  Zip compression
> 
>     * [ ]  Regex library

[...]

PR #2564, opened 2025-11-05
Added "implement-libc" feature.

8lall0:
This feature adds an option from project.json and build options to disable libc unreachable statements in order to be able to implement one for yourself. Useful feature for bare metal developing.

### DIFF
--- a/releasenotes.md
+++ b/releasenotes.md
@@ -4,6 +4,7 @@
 ## 0.7.9 Change list
 
 ### Changes / improvements
+- Add `--custom-libc` option for custom libc implementations.
 
 ### Fixes
 - Regression with npot vector in struct triggering an assert #2219.
--- a/src/build/build_options.c
+++ b/src/build/build_options.c
@@ -167,7 +167,7 @@ static void usage(bool full)
 		PRINTF("");
 		print_opt("--use-stdlib=<yes|no>", "Include the standard library (default: yes).");
 		print_opt("--link-libc=<yes|no>", "Link libc other default libraries (default: yes).");
-		print_opt("--custom-libc=<yes|no>", "Enable your custom libc implementation (default: no).");
+		print_opt("--custom-libc=<yes|no>", "Set to true if a custom libc implementation is provided (default: no).");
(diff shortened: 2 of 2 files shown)


lib/std/libc/libc.c3
```diff
@@ -258,7 +258,7 @@ macro CFile stdin() { return (CFile*)(uptr)STDIN_FD; }
 macro CFile stdout() { return (CFile*)(uptr)STDOUT_FD; }
 macro CFile stderr() { return (CFile*)(uptr)STDERR_FD; }
 
-module libc @if(!env::LIBC);
+module libc @if(!env::LIBC && !env::IMPLEMENT_LIBC);
```
lerno: Rather than doing this, let's change this to `env::NO_LIBC` and then only set env::NO_LIBC is neither exists. Also look at the ones doing `@if(env::LIBC)` to see if they should be included when LIBC is implemented by the custom libc

lerno:
Rename `implement-libc` to `custom-libc`

lib/std/libc/libc.c3
8lall0: NO_LIBC is set by "link-libc", so are you suggesting to use that mechanism?

Book-reader: I think he means changing the definition of `NO_LIBC` to something like `const bool NO_LIBC = !LIBC && !CUSTOM_LIBC`

lerno: Yes exactly.

lerno:
Ping!

8lall0:
> Ping!

I'm dealing with a heavy flu right now xD sorry for the delay, i'm unable to do anything atm

lerno:
It's fine, I hope you get well soon!

LukeGary462:
excited to see some movement on this!
https://github.com/c3lang/c3c/issues/1456#issuecomment-2761869972

lerno:
Thank you

8lall0:
> Thank you

Happy to contribute <3

8lall0:
> Thank you

Happy to contribute <3

PR #2570, opened 2025-11-08
Add lengthSqr

any1here:


tonis2:
@any1here  Isn't the `sq_magnitude` function already doing same thing ?

https://github.com/c3lang/c3c/blob/master/lib/std/math/vector.c3#L6-L7

lerno:
Yes there are those functions that already exist. So maybe we could close this? For future PRs, note that methods also use snake case style, so that the method would have to be `length_sqr`.

any1here:
> Yes there are those functions that already exist. So maybe we could close this? For future PRs, note that methods also use snake case style, so that the method would have to be `length_sqr`.

Is there a reason why the length method is in math.c3 and sq_magnitude in vector.c3?
Also, the naming is confusing since one uses length while the other uses magnitude.

PR #2572, opened 2025-11-11
[stdlib] macOS - Add objc and core foundation types and enums.

terids:
Add additional objc types:
- CGFloat, CGPoint, CGSize, CGRect
- WindowCollectionBehavior, WindowLevel, WindowTabbingMode

Change EventMask to be a `ulong` to better align with objc's `unsigned long long`.

Change existing enum backing types to NSInteger/NSUInteger for 32/64 bit platform compatibility, and to better align with the objc implementations.

Add unit tests for the objc and core_foundation modules.

### DIFF
--- a/lib/std/os/macos/core_foundation.c3
+++ b/lib/std/os/macos/core_foundation.c3
@@ -33,9 +33,6 @@ struct CGRect
 	CGSize size;
 }
 
-const CGFloat NS_VARIABLE_STATUS_ITEM_LENGTH = -1.0;
-const CGFloat NS_SQUARE_STATUS_ITEM_LENGTH = -2.0;
-
 extern fn ZString CFString.getCStringPtr(&self, CFStringEncoding encoding) @cname("CFStringGetCStringPtr");
 extern fn ZString CFString.getCString(&self, char* buffer, usz len, CFStringEncoding encoding) @cname("CFStringGetCString");
 
--- a/lib/std/os/macos/objc.c3
+++ b/lib/std/os/macos/objc.c3
@@ -71,7 +71,10 @@ extern fn ObjcIvar getInstanceVariable(ObjcId id, ZString name, void* outValue) 
 extern fn ObjcIvar setInstanceVariable(ObjcId id, ZString name, void* value) @cname("object_setInstanceVariable");
 extern fn ObjcClass allocateClassPair(ObjcClass cls, ZString name, uint extraBytes) @cname("objc_allocateClassPair");
 
-enum StatusItemLength : (double val) @deprecated("Use NS_VARIABLE_STATUS_ITEM_LENGTH and NS_SQUARE_STATUS_ITEM_LENGTH from std::os::macos::cf.")
+module std::os::macos::objc::ns @if(env::DARWIN) @link(env::DARWIN, "CoreFoundation.framework");
+import std::os::macos::cf;
+
+enum StatusItemLength : (double val) @deprecated("Use NSStatusItemLength.")
 {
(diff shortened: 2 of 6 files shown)


terids:
I was originally going to change them to const, but decided to keep with the style of the existing implementation as this would be a breaking change. I'd be happy to make those changes as I agree it would be much nicer to interface with and keeps it more in line with objc. The unit tests will also need to be changed alongside.

If I make this change, existing code will need to:
- Drop `.val` from enum members
- Add a cast where an `int` or `NSInteger` is expected, OR..
   Change `int` or `NSInteger` to the specific enum type

Any concerns about this? If not I can work on these changes later today for this PR.

Book-reader:
> Add a cast where an int or NSInteger is expected

For this you can do `: const inline NSInteger` which will make the enum convert to `NSInteger` without needing a cast.
<br/>
As for the breaking change part, it *should* be fine since it isn't likely a frequently used part of the stdlib and it would make it a bit more convenient to use, but @lerno should probably weigh in on this in case I'm giving wrong advice.

lerno:
Feel free to convert it to const enums. That's what I was going to suggest anyway.

terids:
@lerno Are you OK with `const inline NSInteger` for the enums? It means the EventType can be used to bitshift the EventMask without casting and follows the NSEvent.h api more closely.

I've tested the changes locally in a separate project.

lerno:
A problem is that we can't have breaking changes in a 0.7.x release. On the other hand, the enums are incorrectly named I think, if we want to follow the actual naming? If so we can retain the older enums and use @deprecated on them, and for the new ones use NS-prefix? For the functions we can see if we can keep both a version using the deprecated enums, and one with the new enums?

terids:
Are the msvc/msys2 build errors something I need to worry about?

```
- 1096/0 test_suite\statements\foreach_zstring.c3: FAILED

 Unexpected compilation errors:
 ------------------------------
  1. Error at foreach_zstring.c3:6: "'std::os::posix::environ' could not be found, did you spell it right?
```

Edit: Disregard, I see a fix has already been pushed in [20dfdf5](https://github.com/c3lang/c3c/commit/20dfdf5c5de8c64a80b814aa57507ce949bc7c84)

lerno:
Thank you. Sorry I made you wait, I've had very little time to sit down for PRs. I made some small adjustments. I think it's probably better to move all "ns" into an "ns" namespace. So I did that change, and modified the constants for status item to use a const enum as well.

lerno:
In the long run the organization of this should be updated. But that is a task to tackle later. Thank you for the contribution!

terids:
@lerno The NS separation is a good shout.

My reasoning for changing StatusItemLength from enum to const was to match the objc library. Just wanted to mention for awareness since everything else is usually an enum. [StatusBarItemLength](https://developer.apple.com/documentation/appkit/status-bar-item-length?language=objc).

I agree the organisation of this could be updated in future. Currently `objc.c3` file contains a mixture of runtime, ns, and appkit bindings.
My thoughts would be to create `runtime.c3` and `message.c3` files that contain all of the raw `extern` bindings from the obj-c runtime, then the `objc.c3` file would contain the friendly helpers for msg_send etc.
At the moment, `msg_send` has a nice macro that unwraps the selector with `sel_getUid` automatically so the caller just needs to pass a string. The other functions, such as `class_getClassMethod` and `class_addMethod` don't have this and need to be converted manually, so it feels a little inconsistent.

Let me know if you'd like to discuss further, I'd be happy to help is this area.

PR #2579, opened 2025-11-15
add join for ThreadPool without destroying the threads

sandervdbosch:
Add the ability to wait for the pool to finish all tasks without destroying all the threads in the pool.
This results in the new `ThreadPool.join()` function, while increasing the ThreadPool object by 56 bytes.

### DIFF
--- a/lib/std/threads/os/thread_posix.c3
+++ b/lib/std/threads/os/thread_posix.c3
@@ -224,11 +224,21 @@ fn bool NativeThread.equals(thread, NativeThread other)
 	return (bool)posix::pthread_equal(thread.pthread, other.pthread);
 }
 
-fn int? NativeThread.join(thread)
+<*
+ @return "the return value of the thread"
+ @return? thread::THREAD_NOT_FOUND : "If the thread was not running"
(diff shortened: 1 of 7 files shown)


lerno:
Why did you mark it a draft?

sandervdbosch:
Yes, I noticed that the thread sleep excessively when using the join function. I will try to fix it and update the pull request.

lerno:
Can you add some additional tests? I commented out some code that you added but all tests still passed, so see if you can add some more asserts on the behaviour.

sandervdbosch:
There are no tests for the FixedThreadPool, when I tried to add them it failed on memory leaks that I was not able to trace down.
Commenting out the changes I made does break functionality, although this is hard to test for the mutex locking and unlocking due to the limitation that the maximum queue size is the same as the amount of threads in the pool.

lerno:
As I was looking through this I did a cleanup of the thread api.

lerno:
Thank you!

sandervdbosch:
You're welcome and thank you too.

PR #2586, opened 2025-11-21
Check for `LLVM_HAS_XTENSA_TARGET` instead of llvm major version when enabling Xtensa target.

Book-reader:
This fixes compiling with https://github.com/espressif/llvm-project since it has an older version of LLVM while still supporting xtensa and also fixes any errors from builds of llvm that aren't compiled with xtensa support, since from what I can tell it's still experimental even in llvm 21+ and unlikely to be enabled in most builds:
image

`LLVM_HAS_XTENSA_TARGET` is set in https://github.com/llvm/llvm-project/blob/bf4dc96bc15464fc4af9dde3a5a686a129abfdbd/llvm/include/llvm/Config/Targets.h.cmake#L88

### DIFF
--- a/lib/std/core/builtin.c3
+++ b/lib/std/core/builtin.c3
@@ -503,7 +503,7 @@ macro void? @try(#v, #expr) @builtin @maydiscard
   {
 	char[] data;
 	// Read until end of file
-	if (@try_catch(data, load_line(), io::EOF)) break;
+	if (@try_catch(data, load_line(), io::EOF)!) break;
 	.. use data ..
   }
 
--- a/lib/std/core/runtime_test.c3
+++ b/lib/std/core/runtime_test.c3
@@ -222,6 +222,7 @@ $endif
 			case "--test-noleak":
 				check_leaks = false;
 			case "--test-nocapture":
+			case "--test-show-output":
 				context.is_no_capture = true;
 			case "--noansi":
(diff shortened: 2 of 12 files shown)


Book-reader:
well that doesn't seem right
edit: right, I forgot that it's set to a bool and not just defined when it's enabled

Book-reader:
for some reason `LLVM_HAS_XTENSA_TARGET` is being defined as `1` for the llvm 18 builds despite the fact it isn't enabled
image

I'll have to look into this a bit more

Book-reader:
it apparently is enabled in version 18, and the other functions do exist, but `LLVMInitializeXtensaAsmPrinter` specifically doesn't
Should I add a check that disables xtensa on versions less than 19 or if the `LLVMInitializeXtensaAsmPrinter` function is missing or something else? Maybe it shouldn't be enabled automatically and instead use a `-DENABLE_XTENSA` so people who know that they need it can enable it

lerno:
Yeah, add the -DXTENSA_ENABLED for now.

Book-reader:
I've added a `-DXTENSA_ENABLE` flag, added the change to the release notes, and added a mention of the Xtensa target and `-DXTENSA_ENABLE` to the readme

lerno:
Thank you!

PR #2596, opened 2025-11-25
Add $foreach_r

data-man:
I remember asking for this a couple of years ago. So I decided to do it myself. :)


lerno:
This is not how we do language feature additions. I don't believe there's even an issue for it.

PR #2598, opened 2025-11-26
Add NSApplicationTerminateReply and registerClassPair to objc stdlib

terids:
The missing `registerClassPair` function is required to register a class and make it active with objc after allocating with the existing `allocateClassPair` function.


lerno:
Thank you!

PR #2620, opened 2025-12-06
Matrix math cleanup

TechnicalFowl:
This change cleans up a few sticking points with the stdlib matrix type.
1. corrects the translation component of the ortho projection matrix to be row major to match the rest of the module.
2. transposes the rotate_y matrix so that the rotation in a consistent direction with rotate_x and rotate_z (clockwise if you are looking in the direction of the axis).
3. Adds a matrix4_look_at and matrix4f_look_at alias in line with the other generic functions in the module.
4. Adds an operator overload attribute for the scalar multiply method.

I also expanded the unit tests for these functions.

### DIFF
--- a/test/unit/stdlib/math/matrix.c3
+++ b/test/unit/stdlib/math/matrix.c3
@@ -13,6 +13,36 @@ fn void test_mat4_translate()
 	assert(translated.m == mat.m);
 }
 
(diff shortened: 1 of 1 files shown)


lerno:
Thank you

PR #2622, opened 2025-12-06
fix ByteBuffer.grow

hwchen:
ByteBuffer.grow calculated its next buffer capacity using just the number of bytes to add; now fixed by using curr len + the number of bytes to add.


lerno:
Thank you!

PR #2629, opened 2025-12-08
Improve the error message for miscasing types in the last argument of a method

cheese3660:
Improves the error message generated in this case
```c3
 1: module main;
 2: import std::io;
 3: 
 4: fn int main(string[] args)
                      ^
(/*...*/main.c3:4:19) Error: Expected ')'.
```
Instead becomes

```c3
1: import std::io;
2: fn int main(string[] args)
               ^^^^^^
(/*...*/test.c3:2:13) Error: Expected typename here. ( '[' in a parameter list must follow a type)
```

### DIFF
--- a/src/compiler/parse_global.c
+++ b/src/compiler/parse_global.c
@@ -1736,6 +1736,12 @@ CHECK_ELLIPSIS:
 		vec_add(params, param);
 		if (!try_consume(c, TOKEN_COMMA)) break;
 	}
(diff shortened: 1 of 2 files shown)


cheese3660:
Oh
I forgot to put the original error message back in, thats easy enough to fix

ManuLinares:
I would drop the "usually".

```c
PRINT_ERROR_LAST("Expected typename here. ('%s' in a parameter list must follow a type).",token_type_to_string(c->tok));

```

parsers should sound authoritative

cheese3660:
Good point, I'll change that

lerno:
Thank you!

PR #2636, opened 2025-12-10
add `--linux-libc=host` option and set default to `host`

DylanDoesProgramming664:
This is a minor enhancement to the `linux-libc` compiler flag and project parameter, adding the `host` option to go with the host's `default_libc` as specified in `src/build/builder.c`.

`LINUX_LIBC_HOST` and `extern LinuxLibc default_libc;` have been added to `src/build/build.h`, and `src/compiler/linker.c` now uses `compiler.build.linuxpaths.libc` instead of `compiler.platform.environment_type` in `find_linux_ld`.


lerno:
Thank you!

DylanDoesProgramming664:
You're most welcome. Happy to help.

PR #2639, opened 2025-12-11
Add Poly1305 Universal Hashing Algorithm (MAC)

NotsoanoNimus:
Adding in another standalone hashing type to create one-time Message Authentication Codes, outside the typical HMAC construct: [Poly1305](https://en.wikipedia.org/wiki/Poly1305).

The included tests for this addition are [as thorough as Google's](https://github.com/google/boringssl/blob/main/crypto/poly1305/poly1305_tests.txt) for this construct.

While this can probably stay in `lib/std/hash`, it's open for discussion as to whether we should move the module to `lib/std/crypto/mac`, or something similar, instead. The purpose for adding this is to later support _Authenticated Encryption with Associated Data_ (i.e., AEAD) structures and functionality in the standard library.

(<sub>by the by, my C3 skills have atrophied a tad - if there's anything to clean up, please do!</sub>)

### DIFF
--- a/lib/std/hash/poly1305.c3
+++ b/lib/std/hash/poly1305.c3
@@ -18,11 +18,15 @@ const KEY_SIZE = 32;
 
 struct Poly1305
 {
-    ulong[3]        h;
-    uint128         r;
-    uint128         nonce;
-    char[TAG_SIZE]  temp;
-    usz             num;
+    ulong[3]        h; // hash internal state
+    uint128         r; // secret portion of key
+    uint128         nonce; // initialization vector, derived from key
+    char[TAG_SIZE]  temp; // last partial ingestion state
+    usz             num; // index into last partial ingestion
+    // Additional, cached state information:
+    ulong           r0;
+    ulong           r1;
+    ulong           s1;
 }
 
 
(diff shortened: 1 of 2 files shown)


ManuLinares:
You are performing two unaligned loads of the Poly1305 r-key (r0, r1) inside every call to blocks() and since r never changes after initialization, these loads and the s1 computation are pure overhead. Moving r0, r1, and s1 into the context during init() removes repeated memory accesses and saves cycles inside the hottest part of the MAC.

example:
```c
...
struct Poly1305
{
    ulong[3]        h;
    uint128         r;
    uint128         nonce;
    ulong           r0;
    ulong           r1;
    ulong           s1;
    char[TAG_SIZE]  temp;
    usz             num;
}

...

fn void Poly1305.init(&self, char[KEY_SIZE] key)
{
    uint128 r128   = @unaligned_load(*(uint128*)&key[0], 1)
                   & 0x0ffffffc_0ffffffc_0ffffffc_0fffffff;

    uint128 nonce  = @unaligned_load(*(uint128*)&key[16], 1);

    // Write to state
    self.r     = r128;
    self.nonce = nonce;

    // Pre-extract low/high limbs
    self.r0 = @unaligned_load(((ulong*)&r128)[0], 1);
    self.r1 = @unaligned_load(((ulong*)&r128)[1], 1);

    // Precompute s1 = r1 + (r1 >> 2)
    self.s1 = self.r1 + (self.r1 >> 2);

    // Clear h & counters
    self.h   = {0U, 0U, 0U};
    self.num = 0;
}

...

fn void Poly1305.blocks(&self, char[] input, ulong pad_bit = 1) @local
{
    // Use cached r0, r1, s1 (precomputed in init)
    ulong r0 = self.r0;
    ulong r1 = self.r1;
    ulong s1 = self.s1;

    ulong h0 = self.h[0];
    ulong h1 = self.h[1];
    ulong h2 = self.h[2];

    for (; input.len >= BLOCK_SIZE; input = input[BLOCK_SIZE..])
...

```

NotsoanoNimus:
> You are performing two unaligned loads of the Poly1305 r-key (r0, r1) inside every call to blocks() and since r never changes after initialization, these loads and the s1 computation are pure overhead. Moving r0, r1, and s1 into the context during init() removes repeated memory accesses and saves cycles inside the hottest part of the MAC.

Excellent observation, @ManuLinares, totally an oversight on my part.

Thanks for taking the time to go through the code. I've incorporated your feedback into the PR.

lerno:
Thank you!

PR #2640, opened 2025-12-11
Add a leniency flag to String.unescape(), and fix a memory leak in String.(un)escape()

cheese3660:
Adds a parameter `lenient` to String.unescape() which changes how it handles unknown escape characters by just resolving them to themselves rather than returning `INVALID_ESCAPE_SEQUENCE`

Also fixes memory leaks in the string.escape and string.unescape methods, by freeing the allocated dstring using a `defer`, as on success these strings are copied.

### DIFF
--- a/lib/std/core/string_escape.c3
+++ b/lib/std/core/string_escape.c3
@@ -1,4 +1,4 @@
-// Copyright (c) 2024 Christoffer Lerno. All rights reserved.
+// Copyright (c) 2024-2025 Christoffer Lerno. All rights reserved.
 // Use of this source code is governed by the MIT license
 // a copy of which can be found in the LICENSE_STDLIB file.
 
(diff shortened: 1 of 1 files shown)


ManuLinares:
mmm, maybe it oughta be just a flag to the original `String.unescape`?

```c
fn String? String.unescape(String s, Allocator allocator, bool allow_unquoted = false, bool lenient = false)
```

and

```c
default:
    if (lenient)
        result.append_char(escape_char);
    else
        return INVALID_ESCAPE_SEQUENCE?;
```

cheese3660:
Hmm, perhaps that may be better

cheese3660:
what did I mess up there lol, I'm tired, oh a typo

lerno:
Thank you.

PR #2643, opened 2025-12-14
Implement ChaCha20 Crypto in stdlib

NotsoanoNimus:
Complete with some tests and a benchmark unit.

This can be used with in-place buffers (the `_mut` suffix on these can be change if desired) and with cloned buffers for read-only inputs.

### DIFF
--- a/lib/std/crypto/chacha20.c3
+++ b/lib/std/crypto/chacha20.c3
@@ -130,49 +130,39 @@ fn void ChaCha20.transform(&self, char[] data)
 	if (!data.len) return;
 
 	usz original_length = data.len;
-	char[] key_stream = @as_char_view(self.key_stream); // shorthand
-	$typefrom(env::ARCH_64_BIT ??? ulong : uint) _a = 0; // try to use the ideal size for the current architecture
+	char[] key_stream = @as_char_view(self.key_stream);
 
-	// Align the incoming data stream to an 8-byte boundary by scrolling along it byte-by-byte.
-	for (; data.len && ((usz)data.ptr % $sizeof(_a)); data = data[1..])
+	// 1. Process remaining bytes in the current keystream block.
+	if (self.position < BLOCK_SIZE)
 	{
-		self.check_position(); // be sure to watch for keystream mutations
-		data[0] ^= key_stream[self.position++];
+		usz len = data.len < (BLOCK_SIZE - self.position) ? data.len : (BLOCK_SIZE - self.position);
+		for (usz i = 0; i < len; i++)
+		{
+			data[i] ^= key_stream[self.position + i];
+		}
+		self.position += len;
+		data = data[len..];
 	}
 
-	// Now that we're aligned to a 64-bit address, we should be able to quickly apply the XORs on chunks at a time.
-	while (data.len >= $sizeof(_a))
+	// 2. Process full blocks at a time, word by word according to the system's architecture.
+	for (; data.len >= BLOCK_SIZE; data = data[BLOCK_SIZE..])
 	{
-		self.check_position(); // mutate when required
-		*($typeof(_a)*)data.ptr ^= @unaligned_load(*($typeof(_a)*)&key_stream[self.position], 1); // aligned assignment
-		self.position += $sizeof(_a); // we've read the X bytes from the key_stream
-		data = data[$sizeof(_a)..]; // we've also mutated X bytes from the data buffer
-		// Annoyingly, there are times when 'position' might fall to something like '61', meaning +sizeof will run out of bounds on the next pass.
-		//   The 'position' variable cannot skip around; i.e., we can't +4 to 61, let it roll back to 0, and consider all ok.
-		//   In these cases, it's best to just consume byte-wise again up until the BLOCK_SIZE, then realign the data buffer pointer again.
-        //   Despite it seeming fruitless, this has almost double performance of the transform.
-		if (@unlikely(self.position + $sizeof(_a) > BLOCK_SIZE && self.position != BLOCK_SIZE))
+		self.mutate_keystream();
+		for (usz i = 0; i < BLOCK_SIZE / usz.sizeof; i++)
 		{
-			for (; data.len && self.position < BLOCK_SIZE; data = data[1..])
-			{
-				// NOTE: no need to check_position here because we know we're creeping up to BLOCK_SIZE already, and not over it.
-				data[0] ^= key_stream[self.position++];
-			}
-			// Then, once again course-correct the data buffer's alignment, because now that's out of sync.
-			//   This time the position value DOES need to be checked (because we crawled up to BLOCK_SIZE, so >= 64 is always true here).
-			self.check_position();
-			for (; data.len && ((usz)data.ptr % $sizeof(_a)); data = data[1..])
-			{
-				data[0] ^= key_stream[self.position++];
-			}
+			((usz*)data.ptr)[i] ^= ((usz*)&self.key_stream)[i];
 		}
 	}
 
-	// Walk up the final bytes, if there are any left.
-	for (; data.len; data = data[1..])
+	// 3. Process any remaining bytes.
+	if (data.len > 0)
 	{
-		self.check_position();
-		data[0] ^= key_stream[self.position++];
+		self.mutate_keystream();
+		for (usz i = 0; i < data.len; i++)
+		{
+			data[i] ^= key_stream[i];
+		}
+		self.position = data.len;
 	}
 
 	// All done. Capture the transformed length of data and check limits.


lib/std/crypto/chacha20.c3
```diff
@@ -0,0 +1,201 @@
...
+	x[c] += x[d]; x[b] = (x[b] ^ x[c]).rotl(7);
+}
+
+
+<* Process the next (or final) chunk of ingested data. *>
+fn void ChaCha20.next_block(&self) @local
+{
+	self.key_stream[..] = self.state[..];
+
+	for (usz i = 0; i < 10; ++i) // unrolling this does not improve performance measurably
```
lerno: Use `i++` for consistent `for` loops in the stdlib.

NotsoanoNimus: All set!

A quick project-wide search shows this `++i` style in numerous stdlib places. I can go through and quickly patch them up in a separate PR if you'd like?

ManuLinares:
Hi, I found in benchmarking that processing in 64bit chunks and then the remaining gives a measurable performance boost, like a ~14% speedup

edit: Sorry I forgot to account for alignment, I've added a fix
<details>
<summary>diff</summary>

```diff
diff --git a/lib/std/crypto/chacha20.c3 b/lib/std/crypto/chacha20.c3
index f082a957..796737e7 100644
--- a/lib/std/crypto/chacha20.c3
+++ b/lib/std/crypto/chacha20.c3
@@ -111,14 +111,43 @@ fn void ChaCha20.transform(&self, char[] data)
 {
 	if (!data.len) return;
 
-	for (usz i = 0; i < data.len; i++)
+	char* p = data.ptr;
+	char* end = data.ptr + data.len;
+	char* key_stream_bytes = (char*)&self.key_stream;
+
+	while (p < end)
 	{
-		if (self.position >= 64) 
+		if (self.position >= BLOCK_SIZE)
 		{
 			self.next_block();
 			self.position = 0;
 		}
-		data[i] ^= @as_char_view(self.key_stream)[self.position++];
+
+		usz remaining_in_block = BLOCK_SIZE - self.position;
+		usz remaining_in_data = end - p;
+		usz chunk_size = $$min(remaining_in_block, remaining_in_data);
+
+		usz i = 0;
+		// Process byte-by-byte until the *p is aligned to 8 bytes.
+		for (; i < chunk_size && ((usz)p + i) % (ulong).alignof != 0; i++)
+		{
+			p[i] ^= key_stream_bytes[self.position + i];
+		}
+
+		// Process in 8-byte chunks
+		for (; i + (ulong).sizeof <= chunk_size; i += (ulong).sizeof)
+		{
+			*(ulong*)(p + i) ^= *(ulong*)(key_stream_bytes + self.position + i);
+		}
+
+		// Process any remaining bytes that did not fit in an 8-byte chunk.
+		for (; i < chunk_size; i++)
+		{
+			p[i] ^= key_stream_bytes[self.position + i];
+		}
+
+		p += chunk_size;
+		self.position += chunk_size;
 	}
 
 	self.bytes_processed += data.len;
```

</details>

NotsoanoNimus:
> Hi, I found in benchmarking that processing in 64bit chunks and then the remaining gives a measurable performance boost, like a ~14% speedup

Hey @ManuLinares, awesome find here, thank you for the feedback.

I do agree that processing this according to the native alignment of the running machine's architecture is the best option. However, wouldn't this synchronization of alignment cause issues with the `key_stream_bytes` if `self.position` falls into an odd number?

For example, let's say I call `transform` first on a 3-byte string. Now `self.position` is at 3. Then I call it again with the next arbitrary-length string input (let's say like 2 KiB), and for the sake of simplicity that's already aligned properly. Then, the portion `*(ulong*)(key_stream_bytes + self.position + i)` will load that `key_stream_bytes` character array with an offset of `[base_pointer] + 3 + 0`, which will likely result in alignment issues during casting without `@unaligned_load` (which defeats optimization AFAIK).

It's possible I need to step deeper into your changes and that I'm mistaken. What do you think?

ManuLinares:
You are right. I've updated the loop to align based on the current keystream position (self.position) instead of the data pointer, then using @unaligned_load/@unaligned_store for both sides xor. it still gives a solid speedup ~12-15% in my benchmarks on x86-64

I've also added two new unit tests `chacha20_chunked_unaligned_keystream` and `chacha20_chunked_keystream`.

Both pass cleanly now, here is the updated diff:

<details>
<summary>updated diff</summary>

```diff
diff --git a/lib/std/crypto/chacha20.c3 b/lib/std/crypto/chacha20.c3
index f082a957..542dff65 100644
--- a/lib/std/crypto/chacha20.c3
+++ b/lib/std/crypto/chacha20.c3
@@ -111,14 +111,46 @@ fn void ChaCha20.transform(&self, char[] data)
 {
 	if (!data.len) return;
 
-	for (usz i = 0; i < data.len; i++)
+	char* p = data.ptr;
+	char* end = data.ptr + data.len;
+	char* key_stream_bytes = (char*)&self.key_stream;
+
+	while (p < end)
 	{
-		if (self.position >= 64) 
+		if (self.position >= BLOCK_SIZE)
 		{
 			self.next_block();
 			self.position = 0;
 		}
-		data[i] ^= @as_char_view(self.key_stream)[self.position++];
+
+		usz remaining_in_block = BLOCK_SIZE - self.position;
+		usz remaining_in_data = end - p;
+		usz chunk_size = $$min(remaining_in_block, remaining_in_data);
+
+		usz i = 0;
+
+		// Align keystream position to 8 bytes
+		for (; i < chunk_size && ((self.position + i) % (ulong).alignof) != 0; i++)
+		{
+			p[i] ^= key_stream_bytes[self.position + i];
+		}
+
+		// Process in 8-byte chunks - safely
+		for (; i + (ulong).sizeof <= chunk_size; i += (ulong).sizeof)
+		{
+			ulong val = @unaligned_load(*(ulong*)(p + i), ulong.sizeof);
+			val ^= @unaligned_load(*(ulong*)(key_stream_bytes + self.position + i), ulong.sizeof);
+			@unaligned_store(*(ulong*)(p + i), val, ulong.sizeof);
+		}
+
+		// Process any remaining bytes that did not fit in an 8-byte chunk.
+		for (; i < chunk_size; i++)
+		{
+			p[i] ^= key_stream_bytes[self.position + i];
+		}
+
+		p += chunk_size;
+		self.position += chunk_size;
 	}
 
 	self.bytes_processed += data.len;
diff --git a/test/unit/stdlib/crypto/chacha20.c3 b/test/unit/stdlib/crypto/chacha20.c3
index ac174f44..9bce1400 100644
--- a/test/unit/stdlib/crypto/chacha20.c3
+++ b/test/unit/stdlib/crypto/chacha20.c3
@@ -367,3 +367,45 @@ fn void boringssl_borrowed_suite_counter_overflow() => @pool()
 		test::@check(mutable == K_OVERFLOW_OUTPUT[:i], "Failed to get the correct crypt result (mutated - index %d).", i);
 	}
 }
+
+fn void chacha20_chunked_transform() @test => @pool()
+{
+	char[] key = "12345678901234567890123456789012";
+	char[] nonce = "abcdefghijkl";
+
+	char[] original_text = "This is a test string that is longer than a single block.";
+	char[] encrypted_once = chacha20::tencrypt(original_text, key[..], nonce[..]);
+
+	char[] text_to_transform = @tclone_slice(encrypted_once);
+
+	ChaCha20 c @noinit;
+	defer c.destroy();
+	c.init(key[..], nonce[..]);
+
+	c.transform(text_to_transform[:3]);
+	c.transform(text_to_transform[3..]);
+
+	test::@check(original_text == text_to_transform, "Chunked decryption did not produce the original plaintext.");
+}
+
+fn void chacha20_chunked_unaligned_keystream() @test => @pool()
+{
+	char[] key = "12345678901234567890123456789012";
+	char[] nonce = "abcdefghijkl";
+
+	char[] original_text = "This is a test string that is longer than a single block and will be used for this test.";
+	char[] encrypted_once = chacha20::tencrypt(original_text, key[..], nonce[..]);
+
+	char[] unaligned_buffer_storage = mem::talloc_array(char, original_text.len + 1);
+	char[] text_to_transform = unaligned_buffer_storage[1..];
+	text_to_transform[..] = original_text[..];
+
+	ChaCha20 c @noinit;
+	defer c.destroy();
+	c.init(key[..], nonce[..]);
+
+	c.transform(text_to_transform[:3]);
+	c.transform(text_to_transform[3..]);
+
+	test::@check(text_to_transform == encrypted_once, "Chunked encryption with unaligned keystream failed.");
+}
```

</details>

worth testing on other platforms though

lerno:
Do you all want to continue optimizing this?

NotsoanoNimus:
After all's said and done, my benchmark on my dev server went from ~4.3 milliseconds to ~2.2 milliseconds. Incredible gains!

Thank you, @ManuLinares, for the contribution to this. If there's anything about what I submitted that looks a bit strange or should be fixed, we can patch it again.

@lerno I'm also using the a possibly strange construct in the `transform` method of the library code (switching types based on `ARCH_64_BIT`). For the sake of optimization, this is mainly about 32- versus 64-bit systems. For outliers (128-bit, 8-bit, et al.), perhaps we can use a better `env::` variable here to swap the type when that's available. If that looks fine to you, I think everything here is ready to go. :+1:

ManuLinares:
> After all's said and done, my benchmark on my dev server went from ~4.3 milliseconds to ~2.2 milliseconds. Incredible gains!
> 
> Thank you, @ManuLinares, for the contribution to this. If there's anything about what I submitted that looks a bit strange or should be fixed, we can patch it again.

Nice, glad to hear it. I'm reviewing your changes, I think it could be simplified maintaining correctness and same speedup. This removes a lot of branching and alignment edge cases, I've dropped check_position() macro

```c3
<*
 Transform some input data using the current context structure.

 @param[inout] data : `The data to transform (encrypt or decrypt).`
*>
fn void ChaCha20.transform(&self, char[] data)
{
	if (!data.len) return;

	usz original_length = data.len;
	char* data_ptr = data.ptr;
	usz remaining = data.len;
	char* key_stream_ptr = (char*)&self.key_stream;

	// 1. Process remaining bytes in the current keystream block
	if (self.position < BLOCK_SIZE)
	{
		usz len = remaining < (BLOCK_SIZE - self.position) ? remaining : (BLOCK_SIZE - self.position);
		for (usz i = 0; i < len; i++)
		{
			data_ptr[i] ^= key_stream_ptr[self.position + i];
		}
		self.position += len;
		data_ptr += len;
		remaining -= len;
	}

	// 2. Process full blocks
	while (remaining >= BLOCK_SIZE)
	{
		self.mutate_keystream();
		// XOR 64 bytes at a time, word by word.
		// The compiler should be able to vectorize this.
		ulong* data_words = (ulong*)data_ptr;
		ulong* key_words = (ulong*)&self.key_stream;
		for (usz i = 0; i < BLOCK_SIZE / ulong.sizeof; i++)
		{
			data_words[i] ^= key_words[i];
		}

		data_ptr += BLOCK_SIZE;
		remaining -= BLOCK_SIZE;
	}

	// 3. Process any remaining bytes
	if (remaining > 0)
	{
		self.mutate_keystream();
		for (usz i = 0; i < remaining; i++)
		{
			data_ptr[i] ^= key_stream_ptr[i];
		}
		self.position = remaining;
	}

	// All done. Capture the transformed length of data and check limits.
	self.bytes_processed += original_length;
	if (@unlikely(self.bytes_processed >= CHACHA20_NONCE_REUSE_LIMIT && !permit_overflow))
	{
		abort(
			"ChaCha20 transform limit (~256 GiB) exceeded. You can set 'chacha20::permit_overflow = true;' at"
			" runtime to disable this panic, but you accept the terrible SECURITY IMPLICATIONS of doing so."
		);
	}
}
```

lerno:
Thank you!

PR #2645, opened 2025-12-14
Initial changes to get c3c compiling and running on NetBSD aarch64 and x64

### LINKED ISSUE #2285: Fails to build on NetBSD.
1. NetBSD supports POSIX API and so this change skips the pragma warning.

```
elf$ git diff
diff --git a/dependencies/miniz/miniz.c b/dependencies/miniz/miniz.c
index e6501010..a60ca1af 100644
--- a/dependencies/miniz/miniz.c
+++ b/dependencies/miniz/miniz.c
@@ -3165,7 +3165,7 @@ static int mz_stat64(const char *path, struct __stat64 *buffer)
 #define MZ_FREOPEN(p, m, s) freopen64(p, m, s)
 #define MZ_DELETE_FILE remove

-#elif defined(__APPLE__) || defined(__FreeBSD__)
+#elif defined(__APPLE__) || defined(__FreeBSD__) || defined(__NetBSD__)
 #ifndef MINIZ_NO_TIME
 #include <utime.h>
 #endif
elf$
```

2.  With the above change this error occurs.  I have not looked into at this time.

```
[ 71%] Building C object CMakeFiles/c3c.dir/src/utils/whereami.c.o
/usr/local/src/c3c/src/utils/whereami.c: In function 'get_executable_path_raw':
/usr/local/src/c3c/src/utils/whereami.c:404:17: error: 'free' called on unallocated object 'buffer1' [-Werror=free-nonheap-object]
  404 |                 free(path);
      |                 ^~~~~~~~~~
/usr/local/src/c3c/src/utils/whereami.c:353:14: note: declared here
  353 |         char buffer1[PATH_MAX];
      |              ^~~~~~~
cc1: all warnings being treated as errors
```

limit-ordinal:
Fixes #2285

These are some changes necessary to get c3c building and running on NetBSD (10.1) on aarch64 and x64. More work is required to get a NetBSD port to the same level as even the OpenBSD port, but I figured it would be better to make those other changes as their own PRs when they are ready. In the meantime, this should allow a NetBSD user on an aarch64 or x64 system to be able to build the compiler from source and start hacking on it.

One thing I noticed while making these changes was that in the standard library, `Off_t` was defined as `usz` for anything other than WIN32. [According to POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_types.h.html), it's supposed to be signed. I changed it to `isz`, but let me know if there's a better option or if there's some reason it was defined as `usz` to begin with.


lerno:
Can you create CI for netbsd please? Otherwise it will be hard to verify that it works.

limit-ordinal:
I could do that, but if I'm going to add CI, do you think it would be better for me to just hold off until I get all tests passing? I thought it might be desirable to just submit these changes for now to avoid saddling you with a big PR all at once and to allow someone else to download the source and at least get the compiler running if they wanted. But, although c3c can already compile many of the examples with these changes, there is a good amount of work to do to get all tests passing still (the main issue seems to be that compatibility versions of a lot of libc functions get incorrectly linked with on NetBSD).

So I could add CI that leaves out the units tests for now, or I could just resubmit this later with the changes necessary to get all the tests passing.  What do you think?

limit-ordinal:
Nevermind, I made a bunch of progress on fixing compiling and linking issues that were causing tests to fail / compiler features not to function, so I figure at this point I may as well just finish the work. I'm closing the pull request until I've got everything passing. I'll leave a comment in the issue with the branch I'm working on in case anyone happens to want to collaborate.

PR #2648, opened 2025-12-15
Implement BLAKE2 Hashing in stdlib

NotsoanoNimus:
Adds [BLAKE2](https://datatracker.ietf.org/doc/html/rfc7693) hashing (also see [blake2.net](https://www.blake2.net/)) to the C3 standard library, along with various thorough unit tests.

The creation of this module was a bit unconventional, due to the various renditions this hash type can assume. While generic modules were an option, I chose to limit code generation while imposing a specified compile-time "length" in the `init` and a matching compile-time `final` length. However, _this requirement is cleanly masked_ when just calling the different aliases that read easily.

```c3
import std::hash::blake2;
// Just from reading left-to-right, it's very clear that this is BLAKE2s-256...
//   No extra frivolities required.
char[] my_hash = blake2::s_256("A typical type to call!", "my_OPTIONAL_key");
// OR:
char[] my_hash_same = blake2::s(256, "A typical type to call!", "my_OPTIONAL_key");
```

To demonstrate, considering the compile-time size agreements is only required when using a hash streaming context:
```c3
import std::hash::blake2;

Blake2b myctx @noinit;
defer myctx.destroy();

myctx.init(blake2::SIZE_384, "my_OPTIONAL_secret");

while (/* read some data from the network or other buffer */)
{
    myctx.update(read_content_bytes);
}

if (expected_hash != myctx.final(blake2::SIZE_384)) return SOME_ERROR?;
```

Additionally, something may be necessary for some of the 'error' cases in the code. I chose not to implement any faultdefs for the module at this time, because it would complicate invocations for the sake of unlikely edge cases.

`BLAKE2b` is optimized for 64-bit platforms. `BLAKE2s` for 32-bit. HOWEVER, these should not be limited or swapped by Arch type, since they produce distinct hashes for the same inputs. These are, to my knowledge, the much more popular variants of BLAKE2 that are out there.

### DIFF
--- a/lib/std/hash/blake2.c3
+++ b/lib/std/hash/blake2.c3
@@ -146,6 +146,8 @@ macro @common_final(#instance, $output_length) @local
 	mem::zero_volatile(#instance.buf[#instance.buflen..]); // pad buffer with zeroes
 	#instance.compress(#instance.buf);
 
+    defer mem::zero_volatile(@as_char_view(*#instance)); // destroy the current context implicitly
+
 	result[:#instance.outlen] = @as_char_view(#instance.h)[:#instance.outlen];
 	return result;
 }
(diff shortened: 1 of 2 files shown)


LowByteFox:
Really exciting PR I must say, only matter of time till Argon2 lands in the stdlib, Poly1305 with ChaCha20_Poly1305 and I can even drop the dependency for WolfSSL!

I would help but my cryptographics skills are rough haha, can possibly look into Argon2 once this gets merged 👀

lib/std/hash/blake2.c3
```diff
@@ -0,0 +1,400 @@
...
+const BLAKE2S_KEYBYTES		@local = BLAKE2B_KEYBYTES 		/ 2;
+const BLAKE2S_SALTBYTES		@local = BLAKE2B_SALTBYTES 		/ 2;
+const BLAKE2S_PERSONALBYTES	@local = BLAKE2B_PERSONALBYTES 	/ 2;
+const uint[8] BLAKE2S_IV @local = {
+	0x6A09E667, 0xBB67AE85, 0x3C6EF372, 0xA54FF53A,
+	0x510E527F, 0x9B05688C, 0x1F83D9AB, 0x5BE0CD19
+};
+const char[16][10] BLAKE2S_SIGMA @local = {
+	{  0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11, 12, 13, 14, 15 },
+	{ 14, 10,  4,  8,  9, 15, 13,  6,  1, 12,  0,  2, 11,  7,  5,  3 },
```
lerno: It would be more elegant to use lines of hex literals for this perhaps? Like:
```
x"000102030405060708090a0b0c0d0e0f",
x"0e0a0408090f0d06010c00020b070503",
```
What do you think?

NotsoanoNimus: Good idea, I went through and cleaned that up. While I was there, I also made `destroy()` implicit, as there's never a use-case to use the hash context again after finalization.

lerno:
Thank you!

PR #2649, opened 2025-12-15
Add Elf32_Shdr and Elf64_Shdr to std::os::linux

mchernigin:
I needed `Elf32_Shdr` and `Elf64_Shdr` for my own, so though I could also contribute. Nothing serious, just copied both structs from `<linux/elf.h>` and fixed formatting.


lerno:
Thank you!

PR #2653, opened 2025-12-17
Linux: implement signal stacktrace using SA_SIGINFO + sigaltstack

ManuLinares:
Adds proper signal handlers for SIGSEGV/SIGBUS/SIGILL on Linux, enables backtraces from signal context, and exits with correct POSIX signal codes (128+signal). Fixes missing "signal stacktrace" support. Just the basics

Test:
```c3
module crash;
import std;

fn void main()
{
	asm {
		movq $rax, 0;
		movl [$rax], 1;  // write to NULL -> SIGSEGV
	}
}
```

before:
```
Illegal instruction        (core dumped) ./crash
exit code: 132
```

with this:
```
exit code: 139
```


ManuLinares:
a better test for this is to cause a stack overflow, as without handling it would cause a coredump

```c3
module crash;
import std;

fn void s() => 	s();
fn void main() => s();
```

lerno:
SA_ONSTACK doesn't work properly on MacOS. It completely erases the stack trace instead. The only place it's guaranteed to work properly is on Linux, so it shouldn't be used as is. `$$trap` is used in order for debuggers to stop there.

lerno:
Can you have a look at those issues?

ManuLinares:
> Can you have a look at those issues?

I guarded the changes with `$if env::LINUX:` and only tested on **Linux x86_64** so far. SA_ONSTACK is not reliable on macOS, I can't test it.. :(

Using $$trap() inside a signal handler is undefined behavior on POSIX: it raises SIGILL while already handling a fatal signal. 
Debuggers already stop at the faulting instruction when the signal is delivered (SIGSEGV/SIGBUS/SIGILL). After printing the backtrace, terminating with _exit(128 + signunber) preserves correct debugger behavior without re-faulting.

If I get time I'll also look at Linux aarch64 and see whether ucontext_t handling needs adjustments there.

ManuLinares:
~I've noticed that when a panic occurs (example from `unreachable()`), the program prints two backtraces. The first is the expected panic message, but the subsequent $$trap() call in default_panic generates a SIGILL, which is then caught by sig_illegal_instruction, leading to a second, redundant backtrace.~

~This can be a bit confusing. Would it be better for default_panic and abort to use libc::abort() instead of $$trap()? This would still cause an abnormal termination and allow for a coredump, but it would likely avoid triggering the custom SIGILL handler, resulting in a cleaner single backtrace for panics.~

~I don't like coredumps, but maybe this is wanted for when you call `unreachable()` or `unsupported()`??~

~I don't know what is the behaviour expected here.~

edit: On linux now the behaviour is always print 1 backtrace, exit with the correct signal code, and debuggers stop correctly.

lib/std/os/posix/process.c3
```diff
@@ -116,24 +134,57 @@ struct PosixUContext_t @if(env::DARWIN)
 	__Darwin_sigaltstack    uc_stack;       /* stack used by this context */
 	PosixUContext_t*        uc_link;        /* pointer to resuming context */
 	__Darwin_size_t         uc_mcsize;      /* size of the machine context passed in */
-	__Darwin_mcontext64*    uc_mcontext;   /* pointer to machine specific context */
+	__Darwin_mcontext64*    uc_mcontext;    /* pointer to machine specific context */
+}
+
+/* taken from sys/ucontext.h */
+typedef Greg_t = long;
```
lerno: Place these either in std::os::linux, or guard these with env::linux && env::x64, including the MContext_t here

lib/std/os/posix/process.c3
```diff
@@ -116,24 +134,57 @@ struct PosixUContext_t @if(env::DARWIN)
...
+struct LibcFPState { char[512] __data; }
+
+struct MContext_t @align(16)
+{
+	Greg_t[23] gregs;
+	LibcFPState* fpregs;
+	ulong[8] __reserved1;
+}
+
+/* struct MContext_t @align(16) @if(env::AARCH64)
```
lerno: Let's avoid commented out things. If this is correct, then hide it behind env::AARCH64 && env::LINUX. I know I left some things commented out, but those were intended as hints for people implementing the stack pointer.

lib/std/os/posix/process.c3
```diff
@@ -116,24 +134,57 @@ struct PosixUContext_t @if(env::DARWIN)
...
-			return uc.uc_mcontext.__ss.__rip;
+			return (void*)uc.uc_mcontext.__ss.__rip;
 		$case env::LINUX && env::X86:
-			return uc.uc_mcontext.gregs[REG_EIP];
+			return uc.uc_mcontext.gregs[REG_EIP];*/
 		$case env::LINUX && env::X86_64:
-			return uc.uc_mcontext.gregs[REG_RIP];
-		$case env::LINUX && env::AARCH64:
-			return uc.uc_mcontext.pc;
+			return (void*)uc.uc_mcontext.gregs[REG_RIP];
```
lerno, line 179: Did you check that this ends up on the correct line? Sometimes you need to do a +1 to make it end up on a valid instruction (you need to test some permutations of instructions to make sure it works)

lib/std/os/posix/process.c3
```diff
@@ -85,22 +94,31 @@ fn CInt backtrace(void** buffer, CInt size) @if(!env::OPENBSD)
...
+	if (libc::setjmp(&backtrace_jmpbuf) == 0)
+	{
+		for (i = 0; i < size; i++)
+		{
+			void* addr = builtin::get_returnaddress(i);
+			if (!addr) break;
+			buffer[i] = addr;
+		}
+	}
+/* 	for (i = 0; i < size; i++)
```
lerno: Why is this commented out?

lerno:
I tested on MacOS, it works fine there, so just these questions.

lib/std/os/posix/process.c3
ManuLinares: on macos yes but on linux this is the correct way. gregs[REG_RIP] (index 16) holds the exact return address, points to the instruction **after** the call to get context or the interrupted point.

lib/std/os/posix/process.c3
ManuLinares: this is correct, I'll uncomment and guard

lib/std/os/posix/process.c3
ManuLinares: I'll guard this for now

lib/std/os/posix/process.c3
ManuLinares: I'll remove the comment, I refactored this loop because setjmp should establish a single recovery point, not be repeatedly executed inside loops. The single setjmp outside the loop is the correct pattern on both Linux and macOS and doing this 128 times adds unwanted overhead, is fragile and especially unsafe on macOS due to clang optimization and weaker signal guarantees from my understanding.

lerno:
Something broke, see the CI logs.

lerno:
Thank you!

PR #2656, opened 2025-12-17
Fix ChaCha20 Alignment Issues

NotsoanoNimus:
Noticed some ChaCha20 unit tests breaking earlier after the merge was completed. This was happening non-deterministically, based on memory alignments and allocations of the various GitHub runners. This can also be confirmed by copying the updated unit test to the current rendition of the code and playing with both the PRNG seed and the `k` value in its nested loops.

My updated unit test was catching on the current rendition of the code, causing alignment access issues, which couldn't merely be solved by `@unaligned_load`.

Spent all day poring over this to figure out how to maintain our efficiency gains while correcting unaligned access once and for all. The required changes weren't much, but the algorithm was difficult to work out without a full whiteboarding session.

Happy to note that the drop in efficiency for safety is almost negligible, and the issue appears to be fixed by this PR.

### DIFF
--- a/benchmarks/stdlib/crypto/chacha20.c3
+++ b/benchmarks/stdlib/crypto/chacha20.c3
@@ -2,6 +2,7 @@
 // Use of this source code is governed by the MIT license
 // a copy of which can be found in the LICENSE_STDLIB file.
 module chacha20_benchmarks;
+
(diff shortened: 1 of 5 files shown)


lerno:
Thank you! Nice corrections too.

PR #2657, opened 2025-12-18
Fix: Correct precision calculation for floating point formatting

ManuLinares:
added extra unit tests


ManuLinares:
I got two edge cases I don't now how to fix:

```python
>>> print(f"{12345.9:.3e}")
1.235e+04
>>> print(f"{0.00087654:.3e}")
8.765e-04
```

but c3
```c
fn void main() {
    double t1 = 12345.9d;
    double t2 = 0.00087654d;

    String s = string::format(mem, "%.3e %.3e", t1, t2);
    io::printn(s);
    //prints    1.234e+04 1.000e-03
    //expected  1.235e+04 8.765e-04
}
```

ManuLinares:
https://git.musl-libc.org/cgit/musl/tree/src/stdio/vfprintf.c

fixed and added more unit tests for the edge cases

lerno:
I'll add the release notes for it.

lerno:
Thank you!

PR #2659, opened 2025-12-18
Add Streebog-256 and -512 Hashing (sometimes aka "GOST-12 hash")

NotsoanoNimus:
Title. This one was interesting to write!

Naming of the hash derived from GOST standards is a bit tricky to pin down. But this one is indeed colloquially known as [Streebog](https://en.wikipedia.org/wiki/Streebog) and is not the same as the original [GOST-94) hash](https://en.wikipedia.org/wiki/GOST_(hash_function)) (though I plan to add that one too, at some point, for completeness :wink:).

Due to the slowness of this hashing algorithm - it's not _awful_, but slower - some SIMD instructions in the code may or may not help it to be at least on-par with Whirlpool. But that's not inherently necessary at the moment; you should see how slow Python's implementation is.


ManuLinares:
Interesting, never seen this hash fn. I think xoring and state storage can be SIMD with refactoring with something like `typedef State512 = ulong[<8>] @simd;`

`const State512[12] ITERATION_CONSTANTS` and `const State512[256] TR` need a little multiline magix to convert, I don't even know if this works but we may do operator overloading for the xoring

`fn State512 State512.xor(self, State512 other) @operator(^) => { (State512)((ulong[<8>])self ^ (ulong[<8>])other) }`

edit: Testing this I got negligible performance speedup with higher optimization levels. The compiler knows best in my testing. :(

NotsoanoNimus:
It's definitely one of the stranger hashes I've implemented for C3.

Very interesting that the performance impacts are negligible when trying to use the SIMD typedefs. I might poke around a bit more later on with some of the inspiration you've given me. If I can find some optimization with enough benefit, I'll certainly publish another PR.

Thank you for looking into it!

lerno:
Make sure you're using the right settings to enable better vector ops and do investigate the resulting asm

lerno:
Do you want to delay adding this?

NotsoanoNimus:
> Do you want to delay adding this?

I think it can be merged for now. Later when I work on GOST94 hashes, I'll revisit this code to see if I can optimize it any more.

lerno:
Thank you!

PR #2661, opened 2025-12-19
Add NetBSD Support

### LINKED ISSUE #2285: Fails to build on NetBSD.

limit-ordinal:
Fixes #2285

Resubmitting this now that all compiler and unit tests are passing and CI has been added.

Changes include:
- Hints for the linker to find several libc functions. On NetBSD, several of the default libc function names point to compatibility versions, leading to some unexpected behavior unless specific C header files are imported. NetBSD C headers contain macros that will translate references to standard libc function names into calls to the modern (renamed) functions on modern systems, but since we lack access to this from within c3 code, I included the modern function names in `@cname` attributes when compiling for NetBSD.
- Struct and constant definitions for sockets, polling, errno, etc.
- Changes to the linker code to work around some quirks in the NetBSD dynamic linker. In particular, LLVM will sometimes generate an output executable / library with more than two PT_LOAD segments, which the dynamic linker on NetBSD does not support. I added flags when compiling / linking on NetBSD that restrict LLVM to generating two segments.
- A target triple for netbsd aarch64 so llvm builds/links the compiler properly on this platform

### DIFF
--- a/test/test_suite/functions/splat_initializer.c3t
+++ b/test/test_suite/functions/splat_initializer.c3t
@@ -36,8 +36,8 @@ fn void main()
 
 @"$ct.test.ParentType" = linkonce global %.introspect { i8 10, i64 0, ptr null, i64 16, i64 0, i64 4, [0 x i64] zeroinitializer }, comdat, align 8
 @"$ct.test.ChildType" = linkonce global %.introspect { i8 10, i64 ptrtoint (ptr @"$ct.test.ParentType" to i64), ptr null, i64 36, i64 0, i64 4, [0 x i64] zeroinitializer }, comdat, align 8
-@test.ABC = local_unnamed_addr constant %ParentType { i32 1, i32 7, i32 12312, i32 -1 }, align 4
-@test.ABC2 = local_unnamed_addr constant %ChildType { %ParentType { i32 1, i32 7, i32 12, i32 0 }, i32 8, i32 0, [10 x i8] zeroinitializer }, align 4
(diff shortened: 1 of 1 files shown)


lerno:
Thank you!

PR #2663, opened 2025-12-19
Add RIPEMD Hashing to stdlb

NotsoanoNimus:
Title. Adds all well-known variants of [RIPEMD](https://en.wikipedia.org/wiki/RIPEMD) to the standard library, except the very first iteration (which is entirely useless/broken).
  - RIPEMD-128
  - RIPEMD-160 (most common by far)
  - RIPEMD-256
  - RIPEMD-320

Extremely easy to use:
```c3
char[*] myhash = ripemd::hash{160}("C3 programming");
```

The macros which outline the permutations within each computation lane are not really something that should be summarized any further. I tried it, and it's not worth it. There are plenty of hashing algorithms with _giant_ S-box structures - like WHIRLPOOL - which take up thousands of lines of code. Having this arrangement allows developers to clearly observe what each digest size does, even if they're very similar to one another.

For a good example of why we _shouldn't_ try to summarize these any further, check out the `RustCrypto/hashing` library to understand why. To each their own.

### DIFF
--- a/test/unit/stdlib/hash/ripemd.c3
+++ b/test/unit/stdlib/hash/ripemd.c3
@@ -62,10 +62,10 @@ fn void basic_streamed()
 }
 
 
-fn void large_input() @if(!env::WIN32) => @pool()
+fn void large_input()
 {
 	char[] one_mb = mem::talloc_array(char, 1024*1024);
-	one_mb = { [0..(1024*1024)-1] = 0xA5 };
+	one_mb[..] = 0xA5;
 
 	test::@check(ripemd::hash{128}(one_mb) == x'e8b57fa5f8415f7a17043b67509fc182');
 	test::@check(ripemd::hash{160}(one_mb) == x'edac5dd2281c86b49b362269636ef1680283d291');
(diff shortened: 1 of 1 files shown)


lib/std/hash/ripemd.c3
```diff
@@ -0,0 +1,575 @@
...
+	{
+		@as_char_view(self.buffer)[($bufsz - avail) : data.len] = data[..];
+		return;
+	}
+
+	@as_char_view(self.buffer)[($bufsz - avail) : avail] = data[:avail];
+	self.transform(self.buffer);
+	data = data[avail..];
+
+	for (bool is_aligned = 0 == (usz)data.ptr % uint.sizeof; data.len >= $bufsz; data = data[$bufsz..])
```
ManuLinares: Maybe deciding alignment once, then running one of two tight loops, instead of branching every block could make a small improvement on some optimization levels.

```
	bool aligned = ((usz)data.ptr & (uint.sizeof - 1)) == 0;

	if (aligned) // when aligned, this optimization is a ~10-20% performance boost
	{
		for (; data.len >= 64; data = data[64..])
		{
			self.transform(((uint*)data.ptr)[:16]);
		}
	}
	else
	{
		char[] buf = @as_char_view(self.buffer);

		for (; data.len >= 64; data = data[64..])
		{
			buf[:64] = data[:64];
			self.transform(self.buffer);
		}
	}
```

lerno:
Fails on Win32 for some reason.

NotsoanoNimus:
> Fails on Win32 for some reason.

Can only assume, with the way it exited the test on that platform, that Win32 just isn't happy about repeated 1MiB `talloc` calls. I tried swapping this 1MiB to a const, putting it in as part of the compilation unit, but `c3c` _really_ didn't seem to like that so I backed off.

So... rather than omitting these, I think I'll just add `@if(!env::WIN32)` or such to the `large_` tests.

lerno:
Still fails..?

NotsoanoNimus:
> Still fails..?

Don't ask me, I just love using CI time. :hammer:

lerno:
Thank you!

PR #2664, opened 2025-12-20
Improve android support & add CI

### LINKED ISSUE #2469: Support outputing correct executables for Termux (a Linux shell for Android)
Running `c3c run` on Termux gives the following error when executing the output file:

```
error: ".....": executable's TLS segment is underaligned: alignment is 8, needs to be at least 64 for ARM64 Bionic
```

Book-reader:
this attempts to fix #2469 (it works for me, but I'd like to get other people test it as well)

I've marked this as draft since I'll also work on adding CI & releases for termux (I can also do it in a separate PR if needed)

The patch can also be reduced down to `if (compiler.platform.os == OS_TYPE_ANDROID) LLVMSetTargetMachineEmulatedTLS(result, true);`, I went with this way in case there are other situations where it could be enabled in the future (eg. a compiler flag that explicitly enables/disables it) but currently that isn't needed.


lerno:
Can we get a termux CI, that would be awesome?

Book-reader:
I've added an action that runs tests & builds releases for termux on aarch64 and x86_64. Neither of them pass currently because some unit tests fail, so I'll look into that

Book-reader:
all stdlib unit tests pass on my phone, so it should all pass in the github actions as well

lerno:
Is it safe to merge as is?

Book-reader:
Yes, all the immediate issues have been fixed and if there is anything else it can be done in another PR

Book-reader:
Nevermind then, it seems like it's a bit unstable as https://github.com/c3lang/c3c/actions/runs/20395562326/job/58610567085?pr=2664 has now failed with a memory error despite no code changing. i'll have to look at that in the morning. This shouldn't be merged until I can figure out what's going on & fix it

Book-reader:
Seems like something I've done has broken the commit history a bit, I'll rebase it on master once I get everything working again

Book-reader:
Just an update on what I have so far: the random CI failures are some sort of race condition that on my phone manifests as a hang on one of the worker threads with the following backtrace:
```cpp
#0  0x0000005559fd43b8 in llvm::MetadataTracking::untrack(void*, llvm::Metadata&) ()
#1  0x00000055593e3ad0 in (anonymous namespace)::AArch64InstructionSelector::~AArch64InstructionSelector() ()
#2  0x00000055593b5928 in llvm::AArch64Subtarget::~AArch64Subtarget() ()
#3  0x000000555940ff24 in llvm::AArch64TargetMachine::~AArch64TargetMachine() ()
#4  0x000000555940fe8c in llvm::AArch64leTargetMachine::~AArch64leTargetMachine() ()
#5  0x00000055591d1fb8 in gencontext_destroy (context=0xb400007f2cafb660) at /data/data/com.termux/files/home/c3c/src/compiler/llvm_codegen.c:85
#6  0x00000055591d1d38 in llvm_codegen (context=0xb400007f2cafb660) at /data/data/com.termux/files/home/c3c/src/compiler/llvm_codegen.c:1128
#7  0x0000005559117484 in thread_compile_task_llvm (compile_data=0xb400007f4c91ea50) at /data/data/com.termux/files/home/c3c/src/compiler/compiler.c:147
#8  0x00000055591bb2a8 in taskqueue_thread (data=0x7fffff4460) at /data/data/com.termux/files/home/c3c/src/utils/taskqueue.c:27
#9  0x0000007fe5af1650 in __pthread_start(void*) () from /apex/com.android.runtime/lib64/bionic/libc.so
#10 0x0000007fe5ae3aac in __start_thread () from /apex/com.android.runtime/lib64/bionic/libc.so
```

Disabling threading fixes it, but since compiling on aarch64 CI is already slower than x64 and I don't want to slow CI down too much, I'll look to see if I can figure out what's causing the race condition and fix it first and if that doesn't work I'll disable threading on android builds.

lerno:
This might be c3c destroying things in the wrong order.

lerno:
In `dev` branch I have changed the order of LLVM dispose commands. Does that fix the situation?

lerno:
I realize that there's a big diff to merge too, if you just want to test it you can change gencontext_destroy to:

```c
static void gencontext_destroy(GenContext *context)
{
	ASSERT(llvm_is_global_eval(context));
	LLVMDisposeBuilder(context->global_builder);
	LLVMDisposeTargetData(context->target_data);
	LLVMDisposeTargetMachine(context->machine);
	if (!context->shared_context) LLVMContextDispose(context->context);
	free(context);
}
```

Book-reader:
That has fixed it, thank you! I'm away from my laptop rn but once I get back I'll rebase this pr onto that

lerno:
This is ready now?

Book-reader:
Yes it is

lerno:
Thank you for all the work on this

PR #2667, opened 2025-12-21
Add BLAKE3 Hashing/XOF to stdlib

NotsoanoNimus:
Adds BLAKE3 hashing/XOF support to the standard library.

```c3
char[*] my_hash = blake3::hash("C3 programming");
// or
char[*] my_keyed_hash = blake3::hash("C3 programming", some_exactly_32_byte_key);
// or
char[*] my_other_keyed_hash = blake3::ctx_hash("C3 programming", "my optional, variable-length key");
// and there are other XOF/seek options for more advanced usage
```

If you want to see how the XOF facet of BLAKE3 is used in C3, check out the unit tests. I tried to keep that as easy as possible.

This whole thing will be optimized with vectorization _in a future second pass_ of all hashing algorithms. For now, this is mostly a 1-to-1 reference implementation translation with a lot of C3-esque summarizations. Majority of the testing is an exact copy of the BLAKE3 Team's unit tests.

For the SIMD scaffolding I've set up, I followed the reference implementation to determine the SIMD degrees. If we want to split some of these `const` values out to `env.c3`, I'd be happy to do so @lerno.

### DIFF
--- a/lib/std/hash/blake3.c3
+++ b/lib/std/hash/blake3.c3
@@ -1,4 +1,4 @@
-// Copyright (c) 2025 Zack Puhl <github@xmit.xyz>. All rights reserved.
+// Copyright (c) 2025-2026 Zack Puhl <github@xmit.xyz>. All rights reserved.
 // Use of this source code is governed by the MIT license
 // a copy of which can be found in the LICENSE_STDLIB file.
 //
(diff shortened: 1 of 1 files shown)


lerno:
We should probably have the same for aarch64, but I haven't had the time to fix it (cpu properties). This is good for now! Thank you

PR #2672, opened 2025-12-22
Benchmark Runner QoL Improvements

NotsoanoNimus:
I work with `compile-benchmark` quite often - only fair I help with its upkeep. Slapped this together in a couple hours, because I really need more consistent results and a way to visualize them, and the benchmark runner is a bit of a rat's nest.

Changes, in no particular order:
- Add a MEDIAN metric to the results
  - The _mean_ gives us throughput information, but it's too heavily skewed by a set's outliers.
  - The _median_ gives us an idea of performance expectations with a 50% probability. Put another way, the tested function has performed at or better than the median in 50% of samples taken.
  - Caveat: a sorted set is required. For high-volume iterations, this can get expensive.
- Improve the output units and refactor that into `NanoDuration`, so it can be used elsewhere as desired
- Provide some pretty colors, oooooo
- Added CSV reporting option to get a resultant data-set
  - Not the most useful thing, but makes it easy to plug benchmark outputs into other software. Could probably be improved upon. <sub>and yeah I'm a boomer who likes CSV, sue me</sub>
- Restructure the benchmark runtime to be more like the tester runtime
- Fixed a divide-by-zero crash when the benchmark iteration count was <100
- Rudimentary command-line options to adjust benchmark options


BWindey:
While trying to review I tried running the code but got an error:
```
Program linked to executable './benchmarkrun'.
Launching ./benchmarkrun --benchmark-no-median
--------- BENCHMARKS ---------

ERROR: 'Negative array indexing (index was -2147483648)'
  in std.time.NanoDuration.unit_str (~/.../lib/std/time/time.c3:129) [~/tmp/benchmarkrun]
  in std.time.NanoDuration.tunit_str (~/.../lib/std/time/time.c3:136) [~/tmp/benchmarkrun]
  in std.core.runtime.run_benchmarks (~/.../lib/std/core/runtime_benchmark.c3:231) [~/tmp/benchmarkrun]
  in std.core.runtime.default_benchmark_runner (~/.../lib/std/core/runtime_benchmark.c3:279) [~/tmp/benchmarkrun]


Program completed with exit code 132.
```

The file:
```c3
fn void foo() @benchmark {
	String s = string::format(mem, "Hello %s", "World");
	free(s);
}
```

There appears to have broken something with timing as I'm getting negative values for cpu-time.



I applied my suggestion that I gave in the reviews below for the `NanoDuration.unit_str()` method and I'll paste the output of running the benchmarks with the code from this PR and the one from current main:
```
$ c3c compile-benchmark test.c3
Program linked to executable './benchmarkrun'.
Launching ./benchmarkrun --benchmark-no-median
--------- BENCHMARKS ---------
Benchmarking test::foo ..  -9223372036854775808.00ns,       -nan CPU clocks, 0 iters in 0.00ns

1 benchmark run.

$ c3vm use --from-source -- compile-benchmark test.c3
Program linked to executable './benchmarkrun'.
Launching ./benchmarkrun
--------- BENCHMARKS ---------
Benchmarking test::foo .. [COMPLETE] 1.26 microseconds, 4042.42 CPU clocks, 10000 iterations (
runtime 12.55 milliseconds)

1 benchmark run.

```

benchmarks/stdlib/hash/ripemd.c3
```diff
@@ -6,7 +6,8 @@ module ripemd_bench;
 fn void initialize_bench() @init
 {
 	set_benchmark_warmup_iterations(3);
-	set_benchmark_max_iterations(256);
+	set_benchmark_max_iterations(1024);
+    set_benchmark_report_median(true);
```
BWindey: Inconsistent whitespace / mix tabs and spaces

benchmarks/stdlib/hash/streebog.c3
```diff
@@ -7,6 +7,7 @@ fn void initialize_bench() @init
 {
 	set_benchmark_warmup_iterations(3);
 	set_benchmark_max_iterations(256);
+    set_benchmark_report_median(true);
```
BWindey: Again spaces used while tabs are norm

lib/std/core/runtime_benchmark.c3
```diff
@@ -19,80 +37,109 @@ fn BenchmarkUnit[] benchmark_collection_create(Allocator allocator)
 	foreach (i, benchmark : fns)
 	{
 		benchmarks[i] = { names[i], fns[i] };
-		if (!bench_fn_iters.has_key(names[i])) bench_fn_iters[names[i]] = benchmark_max_iterations;
+		if (!bench_context.fn_iters.has_key(names[i])) bench_context.fn_iters[names[i]] = bench_context.max_iterations;
```
BWindey: Maybe I'm alone in this, but I think that this line is way too long now?

lib/std/core/runtime_benchmark.c3
```diff
@@ -19,80 +37,109 @@ fn BenchmarkUnit[] benchmark_collection_create(Allocator allocator)
...
-uint benchmark_warmup_iterations @private = DEFAULT_BENCHMARK_WARMUP_ITERATIONS;
-uint benchmark_max_iterations @private = DEFAULT_BENCHMARK_MAX_ITERATIONS;
+<*
+ @require $defined((double)set[0])
+ @require lengthof(set) > 0
+*>
+macro get_median(set) @local
+	=> lengthof(set) % 2 != 0
+		? (double)set[lengthof(set) >> 1]
+		: ((double)(set[lengthof(set) >> 1 - 1] + set[lengthof(set) >> 1]) / 2.0);
```
BWindey: I get that `>> 1` is technically faster then dividing by 2, but it's less readable and one of the first compiler optimizations, so I would still write `/ 2` instead of `>> 1`. Note that you already wrote a `/ 2.0`.

lib/std/core/runtime_benchmark.c3
```diff
@@ -19,80 +37,109 @@ fn BenchmarkUnit[] benchmark_collection_create(Allocator allocator)
...
-	foreach (&unit : benchmarks)
+	for (usz i = 1; i < args.len; i++)
 	{
-		if (max_name < unit.name.len) max_name = unit.name.len;
+		switch (args[i])
+		{
+			case "--noansi":
+				ctx.has_ansi_codes = false;
+			case "--useansi":
+				ctx.has_ansi_codes = true;
```
BWindey: I would prefer a simple `--noansi` or `--ansi=on|off` instead of having two seperate flags.
Also, what's the default?

lib/std/core/runtime_benchmark.c3
```diff
@@ -101,86 +148,133 @@ fn bool run_benchmarks(BenchmarkUnit[] benchmarks)
...
 	name.clear();
 
-	foreach (unit : benchmarks)
+	DString csv_report;
+	csv_report.init(mem);
+	defer csv_report.free();
+
+	csv_report.append("Function Name,Iterations,Clock Cycles,Average,Average Nanoseconds");
+	if (ctx.report_median) csv_report.append(",Median,Median Nanoseconds");
+	csv_report.append("\r\n");
```
BWindey: I'm not sure if we have a policy about `\r\n` vs `\n`? Not implying this is wrong, I just don't know if the standard library or the compiler ever writes to a file anywhere else and if they use the carriage return there or not.

lib/std/time/time.c3
```diff
@@ -114,6 +114,28 @@ fn NanoDuration Duration.to_nano(td) => (NanoDuration)td * 1_000;
...
+	while (reduced_value > 1_000) reduced_value /= 1_000;
+
+	DString out_str;
+	out_str.appendf(
+		use_short_units ? "%.2f%s" : "%.2f %s",
+		reduced_value,
+		(use_short_units ? short_units : units)[math::min(3, (int)math::floor(math::log((double)*self, 1_000)))]
+	);
+	return out_str.str_view();
+}
```
BWindey, line 133: Calculating the unit could be cleaner if you have a `unit_idx` which you increment each iteration of the `while-loop`, and then you can also have an extra condition inside the while-loop to stop on seconds:

```c3
fn String NanoDuration.unit_str(&self, Allocator allocator, bool use_short_units = true)
{
	static String[] units = { "nanoseconds", "microseconds", "milliseconds", "seconds" };
	static String[] short_units = { "ns", "us", "ms", "s" }; // using Mu here causes spacing oddities, just using 'us'

	double reduced_value = (double)*self;
	uint units_idx = 0;
	while (reduced_value > 1_000 && units_idx + 1 < units.len)
	{
		reduced_value /= 1_000;
		units_idx += 1;
	}

	if (use_short_units) {
		return string::format(allocator, "%.2f%s", reduced_value, short_units[units_idx]);
	} else {
		return string::format(allocator, "%.2f %s", reduced_value, units[units_idx]);
	}
}
```

BWindey: Furthermore the passed `Allocator` was not used inside the `.unit_str()` function.

lib/std/core/runtime_benchmark.c3
BWindey: I just realised this is set by the compiler when invoking `compile-benchmark`. Not sure how that would be best then?

lib/std/core/runtime_benchmark.c3
NotsoanoNimus: I think that's fair. While this is writing only to `stdout` with everything else, I added CRLF in case the user is running in a terminal that requires it and they're trying to capture that output to a text file. However, since I don't know much about that sort of thing, I think it's safe to just adjust it back to `\n` given the lack of CRLF anywhere else in the code.

However, one thing to note is maybe we should eventually add an `env::NEWLINE` or similar, based on the runtime operating system.

BWindey: Good idea, I'll tag @lerno for that.

lib/std/core/runtime_benchmark.c3
NotsoanoNimus: The runtime determines ANSI support dynamically and will default to that determination if an explicit parameter is not set. I swapped this out now to behave closer to the compiler flags.

```c3
	for (usz i = 1; i < args.len; i++)
	{
		switch (args[i])
		{
			case "--ansi":
				if (catch @try(ctx.has_ansi_codes, get_yes_no(args[i..]))) return false;
				i++;
			case "--median":
				if (catch @try(ctx.report_median, get_yes_no(args[i..]))) return false;
				i++;
			case "--csv-report":
				ctx.generate_csv = true;
			default:
				io::printfn("Invalid argument: %s", args[i]);
				return false;
		}
	}
```

These are parameters you can give to the generated `benchmarkrun` executable file after compiling it. This is useful for doing repeat runs of a bench without recompiling. But, yep, when you invoke the benchmark through `c3c`, these parameters are proxied anyway.

Also, fixed a bug with not setting the `median` switch to its `NOT_SET` default. Push coming soon.

lib/std/time/time.c3
NotsoanoNimus: Fixed, great idea - thank you! Push coming soon.

data-man:
> Caveat: a sorted set is required. For high-volume iterations, this can get expensive.

Sorry, as the author of initial implementation of benchmarking, I strongly object to this change.

BWindey:
> > Caveat: a sorted set is required. For high-volume iterations, this can get expensive.
> 
> Sorry, as the author of initial implementation of benchmarking, I strongly object to this change.

Can you provide info why?

NotsoanoNimus:
> Sorry, as the author of initial implementation of benchmarking, I strongly object to this change.

I respect your opinion about this, but a median is necessary for the reasons I detailed in the initial PR. The call to quicksort happens after all samples are already taken, if and only if the median report will be given.

If you don't want to use a median, it's disabled by default. You can also turn it off permanently for your bench programmatically in a `bench_setup` type `@init` function, or there's a command-line argument to turn it off (both for `c3c` and the bench executable).

lerno:
@data-man would you like to explain why?

data-man:
> @data-man would you like to explain why?

It's already 6 a.m. for me, I'll write tomorrow if possible.

data-man:
So, here are some thoughts:
1. `median` is slightly useless because it's **always** very close to `average`.
For example `$ c3c --benchmark-median=yes compile-benchmark chacha20.c3`:
```Launching ./benchmarkrun --median yes
-------------------------- BENCHMARKS ---------------------------
Benchmarking chacha20_benchmarks::gogo_chacha20 ............   42.55ms avg,  42.22ms med, 119474064.00 CPU clocks, 1024 iters in 43.57s
Benchmarking chacha20_benchmarks::gogo_chacha20_unaligned ..   42.86ms avg,  42.54ms med, 120355856.00 CPU clocks, 1024 iters in 43.89s
```
2. > and yeah I'm a boomer who likes CSV, sue me

I think most users would prefer Markdown. But ok, MD support can be added later.

3. > Restructure the benchmark runtime to be more like the tester runtime

The compiler does not need to be patched for this. I propose to add `bench_config` struct and `set_benchmark_config` function. And `DEFAULT_BENCHMARK_CONFIG` const, of course, with `warmup_iterations` and `max_iterations` fields. This is a more flexible approach in the long term.

4. Do you know about this switch `--benchfn <name>           - Override the benchmark runner function name.`?

lerno:
Can we have a follow up to those concerns @NotsoanoNimus ?

NotsoanoNimus:
> [...] `median` is slightly useless because it's **always** very close to `average`.
>        For example `$ c3c --benchmark-median=yes compile-benchmark chacha20.c3`:

Perhaps "slightly" useless, but not entirely so. The fact of the matter is that a median metric from the runtime will objectively pardon its results from any significant outliers, which will skew an average value. I don't see why a metric which describes "at least 50% of runs hit this performance mark" isn't useful.

> I think most users would prefer Markdown. But ok, MD support can be added later.

Totally agree with this. CSV was a quality-of-life, might-as-well addition to allow at least _some_ automated parsing of runtime results if one so desires.

Given that a Markdown table format is just as simple as combining commas a la CSV, I would have neither difficulty nor objection to adding this in as another option.

> The compiler does not need to be patched for this. I propose to add `bench_config` struct and `set_benchmark_config` function. And `DEFAULT_BENCHMARK_CONFIG` const, of course, with `warmup_iterations` and `max_iterations` fields. This is a more flexible approach in the long term.

This isn't a bad idea, and I wouldn't have a problem changing this around some more to accommodate it. A struct-splat with `DEFAULT_BENCHMARK_CONFIG` would make it very easy to change runtime options for the runner indeed.

I'd have to double-check that nothing is _required_ before the runtime gets to an `@init` method, but this should be fine.

> [...] Do you know about this switch `--benchfn <name>           - Override the benchmark runner function name.`?

For sure. Rather than create a new one or maintain my own, I thought it better to show some love to the current runtime without changing what it actually does too drastically (despite shifting inner its workings).

-----

Thanks for your honest feedback, @data-man. Ultimately, my heart won't be broken if this PR isn't merged. But before that's determined, let me run this TODO list back and make some changes for you to review further.

lerno:
Should this be a draft until you update?

lerno:
What's the current status?

data-man:
@lerno, I'd like to know your opinion about a progress bar. Is it really necessary?

lerno:
To me a progress bar is not necessary, esp given that it can interfere with measurements. I think the final list is the most important.

NotsoanoNimus:
> To me a progress bar is not necessary, esp given that it can interfere with measurements. I think the final list is the most important.

The progress bar does not affect the sampling of each run, at least not explicitly (meaning there might be some other cache/performance aspect outside of the code).

Regardless, I don't think I'm going to continue working on these improvements at this time anyhow: the further changes I've made have deviated so far from the original PR that it's not really sensible to push without opening another one sometime in the future. And I don't really have time to iterate on this right now.

~~I will open a separate PR just to patch the bug mentioned in the PR description.~~ Nevermind, this was already fixed separately by @ManuLinares. :+1:

lerno:
Wait, wasn't this more than the progress bar? There's already one progress bar that we have. It's the one that is not necessary for me, although I can understand that seeing it might be nice. Just a spinner might be sufficient though. And didn't this one also have a CSV output?

lerno:
This one was merged as PR #3004

PR #2677, opened 2025-12-23
Github CI: fixes

ManuLinares:
~~This PR doesn't fix the BSD errors, it just shows them.~~

- Group each BSD test in its own sub-shell. _Without this, if an error occurs, it exits, but we don't know "where" it failed_
- Some test_suite_runners weren't correctly printing. added --no-terminal
- Split runs
- openbsd and netbsd now passes all tests (except dynlib for openbsd)
- added win-llvm 21.1.8 and fixed CMakeLists.txt
- added full unit tests and suite for build-msys2-mingw and build-msys2-clang.

### DIFF
--- a/wrapper/src/wrapper.cpp
+++ b/wrapper/src/wrapper.cpp
@@ -359,7 +359,7 @@ void LLVMBuilderSetFastMathFlags(LLVMBuilderRef Builder, FastMathOption option)
 bool LLVMHasUseList(LLVMValueRef value)
 {
 	llvm::Value *val = llvm::unwrap<llvm::Value>(value);
-	return val->hasUseList();
+	return !llvm::isa<llvm::ConstantData>(*val);
 }
 
 LLVMValueRef LLVMConstBswap(LLVMValueRef ConstantVal)


Book-reader:
Another way you can group them is by splitting them into separate `run:` sections like the other OS: https://github.com/c3lang/c3c/blob/master/.github/workflows/main.yml#L38-L82
It would make it easier to read and maintain as well

ManuLinares:
> Another way you can group them is by splitting them into separate `run:` sections like the other OS: https://github.com/c3lang/c3c/blob/master/.github/workflows/main.yml#L38-L82 It would make it easier to read and maintain as well

I don't think that works with the current runner based on virtualbox, but I may have outdated info. 

Tomorrow I will investigate further alternatives.

https://github.com/vmactions/netbsd-vm

limit-ordinal:
As an aside, I'm working on fixes for the NetBSD build / test errors right now. I don't have an OpenBSD system, so it's less easy for me to fix the OpenBSD errors, but the fixes for those will probably be similar. When I'm done with my changes I can open a PR as well.

limit-ordinal:
I copied your changes over to my fork and it helped me fix the build errors on NetBSD, so nice work. One issue that came up after I fixed some of the errors was that the `run -vvv` command prints so much text that it seems like it crashes the VM.

So in `.github/workflows/main.yml`, changing line 1215 (in the build testproject section) to

```
../../build/c3c run -vv --trust=full
```

and line 1223 (in the testproject direct linker section) to

```
../../build/c3c run -vv --linker=builtin --trust=full
```

(in both cases, just changing the `-vvv` flag to `-vv`) should prevent that. (EDIT: a similar change to the OpenBSD CI is probably also needed.)

Anyway, with the updated CI yaml file I've got all steps for NetBSD completing successfully on my fork. I'm not sure about OpenBSD. Some of the changes will be very similar to the changes necessary for NetBSD, but it's throwing errors that I didn't see for NetBSD and I'm just less familiar with it in general. I'll do some research and see if I can fix those too, but in any case I have NetBSD working with the updated CI.

LowByteFox:
I wouldn't really rely on `sshfs` as on some of my projects it used to fail.

Probably was fixed.

ManuLinares:
CI: I've separated the runs.

@limit-ordinal you say it needs `--linker=builtin` in some places? If you can test this on your own please make a pull request to my https://github.com/ManuLinares/c3c/tree/fix_bsd_ci_workflow

_I was only trying to fix the CI because I noticed it wasn't reporting the build failures, so nobody was looking at this._

.github/workflows/main.yml
```diff
@@ -956,95 +952,128 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-      - name: OpenBSD VM
-        uses: vmactions/openbsd-vm@v1.2.5
+      - name: NetBSD VM
+        uses: vmactions/openbsd-vm@v1
```
lerno, line 1103: NetBSD?

ManuLinares: Oops, 👍

limit-ordinal:
@ManuLinares Sorry I wasn't clear. The issue was actually the verbosity of the commands in the testproject checks. Commands with the `-vvv` flag were causing the runners to fail, whereas changing the flag to `-vv` would get them to work. But I think that was an artifact of the previous way you structured the checks, because with your recent rework, **this is no longer an issue**. So this CI looks fine to me now.

After this gets merged in I can submit a separate PR with my code changes to fix the NetBSD build. I am _almost_ done with that, I think.

ManuLinares:
@lerno What do you think about separating msvc-debug into separate workflow

Runs msvc-debug only on maintainer pushes to any branch, keep CI fast and still catches
errors.

ManuLinares:
gentle ping @lerno

lerno:
It still needs to run before pull requests are accepted. But we could also take a look at win-llvm and update that one. That would speed up things and probably give the same amount of checks.

Issue #2678, opened 2025-12-23
Comments in project.json


cadamsdev:
**Problem**
When I create a new project using `c3c init` it creates comments in the `project.json` file. Since JSON doesn't support comments I think it would be better to use a different file extension like `.jsonc` or don't add comments instead.
<img width="475" height="307" alt="Image" src="https://github.com/user-attachments/assets/ba5b34ba-1f17-4415-b648-6fcac5ccb17f" />

**Expected**
Don't add comments or use a different file extension e.g `.jsonc` instead.

lerno:
Rename it to `project.json5` if this is a problem.

cadamsdev:
> Rename it to `project.json5` if this is a problem.

Sure but I think it would be better to have `c3c` do that by default..

lerno:
More likely project.json will have comments removed.

waveproc:
@cadamsdev Here is a vscode extension I found for json5 code highlighting, formatting, etc: https://marketplace.visualstudio.com/items?itemName=blueglassblock.better-json5

Honestly, I like json5 better than json. The comments in the file are super helpful since you don't have to look at the documentation for simple things as a beginner. And changing the filename is easy enough. But yes, I usually remove some of the comments anyway to remove the visual noise.

ManuLinares:
in vscode you can add this line to your `settings.json`

```json
"files.associations": {
    "*.json": "jsonc"
}
```

I don't have a strong opinion on "changing the extension of project.json", it's a non-issue for me.

cadamsdev:
> in vscode you can add this line to your `settings.json`
> 
> ```json
> "files.associations": {
>     "*.json": "jsonc"
> }
> ```
> 
> I don't have a strong opinion on "changing the extension of project.json", it's a non-issue for me.

Right but it's kinda tedious having to change the editor settings to get c3 working right. Like Lerno said I think it would be better to remove the comments instead of changing the extension. You don't need to add documentation to the project.json that's what the doc site is for.

waveproc:
I don't agree with removing the comments from project.json, because other build system files like cmake and buck/bazel allow comments. The compiler already allows the comments, just not the code editor. However, I like how simple the project.json setup is, even if it's not a replacement for a more advanced build system. As of right now, the "[Project Configuration](https://c3-lang.org/build-your-project/project-config/)" section in the sidebar of the docs is kinda hidden, but it still has comments in between each line, because that is simply easier to read than the comments in a separate list below.

[...]

PR #2679, opened 2025-12-23
change default project file name to project.json5; json5 spec allows comments

waveproc:
Fixes this issue: https://github.com/c3lang/c3c/issues/2678

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1071,7 +1071,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: NetBSD VM
-        uses: vmactions/netbsd-vm@v1
+        uses: vmactions/netbsd-vm@v1.2.6
         with:
           prepare: |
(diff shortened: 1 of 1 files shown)


waveproc:
Haven't tested the fix yet.

lerno:
No, I'm sorry, we're not going to change this. Possibly removing comments from the json, but not changing it to json5.

Book-reader:
Maybe we should go with `project.xml` instead 😆.
I am curious though, is there any specific reason for not going json5 (complex to implement, lack of editor support, etc)? I don't have any major problems with the current format, but without comments in the template it would be harder for a beginner to figure out what things do without finding it in the docs (which, as we've seen from the youtube videos, they don't always do).
Would you consider migrating to a different config format if there was something that has comments, supports everything that the current json format has, and isn't crazily verbose? (As long as it isn't yaml lol, whitespace sensitive config files 😬)

Issue #2638, opened 2025-12-11
macOS install instructions incomplete?


moomerman:
I'm looking forward to trying out C3.  I just followed the [macOS install instructions](https://c3-lang.org/getting-started/prebuilt-binaries/#installing-on-mac-arm64) from the website that state:

```
Installing on Mac Arm64
* Make sure you have XCode with command line tools installed.
* [Download the C3 compiler. ](https://github.com/c3lang/c3c/releases/latest/download/c3-macos.zip)Or the [debug build ](https://github.com/c3lang/c3c/releases/latest/download/c3-macos-debug.zip).
* Unzip executable and standard lib.
* Run ./c3c.
```

However following these instructions results in:

```
./c3c
dyld[28179]: Library not loaded: /opt/homebrew/opt/llvm@18/lib/libunwind.1.dylib
  Referenced from: <5CFAB7E2-3E1C-3826-A8C9-D5360C535C17> /Users/richard/Downloads/macos/c3c
  Reason: tried: '/opt/homebrew/opt/llvm@18/lib/libunwind.1.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/opt/homebrew/opt/llvm@18/lib/libunwind.1.dylib' (no such file), '/opt/homebrew/opt/llvm@18/lib/libunwind.1.dylib' (no such file)
fish: Job 1, './c3c -v' terminated by signal SIGABRT (Abort)
```

`brew install llvm@18` fixes the problem

lerno:
Also related: #1388 #1585. Short story is that it builds by grabbing the *static* LLVM libraries from Homebrew. But those are idiotically built with *dynamic dependencies* to other homebrew packages. The solution is for the CI to use its own compiled LLVM static libraries. There is only a little work that has been done on that so far.

lfnoise:
I also hit this. `brew install llvm@18` worked.

fireinrain:
can macos intel version works?

shakfu:
I hit this as well, and I have to `brew install llmv@18` (even though I already had `llvm` via home-brew).

I searched for the build-from-source instructions, and I was able to build via the latest `llvm` (and delete llvm@18) using the following:

```sh
git clone --depth=1 https://github.com/c3lang/c3c.git
cd c3c
cmake -B build -S . -DLLVM_DIR=`brew --prefix llvm`/lib/cmake/llvm
cmake --build build
cd build
cmake --install . --prefix <your-prefix-path>
```

[...]

PR #2684, opened 2025-12-26
Update MacOS install instructions in readme.md

waveproc:
This addresses https://github.com/c3lang/c3c/issues/2638


lerno:
I understand, but that would just patch a broken situation. The actual solution is fixing the CI, which in turn means creating our own build of LLVM. This hasn't been done yet because I've not had enough time. But using win-llvm as a base it could be done fairly easily.

waveproc:
Okay. The temporary fix that worked for me was just building with Nix:
1. [Install Determinate Nix](https://docs.determinate.systems/)
2. `git clone https://github.com/c3lang/c3c`
3. `cd c3c`
4. `nix develop`
5. `cmake -B build -S . $C3_CMAKE_FLAGS`
6. `cmake --build build`
7. add c3c to path

waveproc:
How would I link the dynamic libraries this way, using llvm/lld (and the other deps) from devbox instead of brew? Note that this is only for the precompiled c3c.
```
% ./c3c
dyld[10387]: Library not loaded: /opt/homebrew/opt/zstd/lib/libzstd.1.dylib
  Referenced from: <C22D4466-6944-370C-AD47-4BAEA5CEAAD8> /Users/user/Downloads/c3c_macos/c3c
  Reason: tried: '/opt/homebrew/opt/zstd/lib/libzstd.1.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/opt/homebrew/opt/zstd/lib/libzstd.1.dylib' (no such file), '/opt/homebrew/opt/zstd/lib/libzstd.1.dylib' (no such file)
zsh: abort      ./c3c
```

lerno:
This should work now with the new CI flow.

PR #2688, opened 2025-12-26
Merge AsciiCharset CT/non-CT Functions

NotsoanoNimus:
Combines functionality of compile-time and runtime variants of the same operation with AsciiCharset, so constant folding can occur automatically.

Also adds in some extra constant sets to use.

Related, basic unit tests included.

### DIFF
--- a/lib/std/core/ascii.c3
+++ b/lib/std/core/ascii.c3
@@ -114,33 +114,36 @@ const char[256] TO_LOWER @private  = { ['A'..'Z'] = 'a' - 'A' };
 
 typedef AsciiCharset = uint128;
 
-macro AsciiCharset create_set(String string)
+macro AsciiCharset @create_set(String $string) @const
+{
+	AsciiCharset $set;
+	$foreach $c : $string:
+		$set |= 1ULL << $c;
+	$endforeach
+	return $set;
+}
+
+fn AsciiCharset create_set(String string)
 {
 	AsciiCharset set;
 	foreach (c : string) set |= (AsciiCharset)1ULL << c;
 	return set;
 }
 
-macro AsciiCharset @create_set(String $string) @const
-{
-	AsciiCharset $set;
-	$foreach $c : $string: $set |= 1ULL << $c; $endforeach
-	return $set;
-}
-
 macro bool AsciiCharset.@contains($set, char $c) @const => !!($c < 128) & !!($set & (AsciiCharset)(1ULL << $c));
 
-macro AsciiCharset combine_sets(AsciiCharset #left, ...) @safemacro
+macro AsciiCharset @combine_sets(AsciiCharset $first, AsciiCharset... $sets) @const
 {
-	$if $defined(var $left = #left, AsciiCharset[] $others = { $vasplat }):
-		var $left = #left; var $others = { $vasplat };
-		$foreach $s : $others: $left |= (uint128)$s; $endforeach
-		return $left;
-	$else
-		var left = #left; AsciiCharset[] others = { $vasplat };
-		foreach (c : others) left |= (AsciiCharset)c;
-		return left;
-	$endif
+	var $res = $first;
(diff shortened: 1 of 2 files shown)


lib/std/core/ascii.c3
```diff
@@ -114,25 +114,45 @@ const char[256] TO_LOWER @private  = { ['A'..'Z'] = 'a' - 'A' };
...
 }
 
-macro bool AsciiCharset.contains(set, char c) => !!(c < 128) & !!(set & (AsciiCharset)(1ULL << c));
+macro bool AsciiCharset.@contains(#set, char #c)
+{
+	$if $defined(var $set = #set, var $c = #c):
+		var $set = #set; var $c = #c;
+		return !!($c < 128) & !!($set & (AsciiCharset)(1ULL << $c));
+	$else
+		return !!(#c < 128) & !!(#set & (AsciiCharset)(1ULL << #c));
```
Book-reader:
```suggestion
		var set = #set; var c = #c;
		return !!(c < 128) & !!(set & (AsciiCharset)(1ULL << c)); 
```

In the runtime case you evaluate `#c` twice which might cause side effects, it would probably be good to store it in a variable like in the constant case.
Since there would be no double evaluation or other unexpected behaviour it may be fine to add `@safemacro` so it can continue to be called `.contains`

lib/std/core/ascii.c3
```diff
@@ -114,25 +114,45 @@ const char[256] TO_LOWER @private  = { ['A'..'Z'] = 'a' - 'A' };
...
+		$foreach $c : $string: $set |= 1ULL << $c; $endforeach
+		return $set;
+	$else
+		AsciiCharset set;
+		foreach (c : string) set |= (AsciiCharset)1ULL << c;
+		return set;
+	$endif
 }
 
-fn AsciiCharset create_set(String string)
```
Book-reader, line 126: Generally if you remove a function you should add an `@deprecated` forwarding function to keep backwards compatibility

lib/std/core/ascii.c3
```diff
@@ -114,25 +114,45 @@ const char[256] TO_LOWER @private  = { ['A'..'Z'] = 'a' - 'A' };
...
 {
-	AsciiCharset set;
-	foreach (c : string) set |= (AsciiCharset)1ULL << c;
-	return set;
+	$if $defined(var $left = #left, var $others = { $vasplat }):
+		var $left = #left; var $others = { $vasplat };
+		$foreach $s : $others: $left |= (uint128)$s; $endforeach
+		return $left;
+	$else
+		foreach (c : { $vasplat }) #left |= (uint128)c;
```
Book-reader: This would modify the original set passed in as `#left` in the runtime case (and evaluate it multiple times), you should probably have `var left = #left` and do the operations on that instead

NotsoanoNimus:
Thank you, @Book-reader, for the helpful review feedback - I learned a few things. Please let me know if you notice anything else that should be changed.

lerno:
1. If this is deprecating the old one then the old still needs to be retained with a deprecation and forwarding to the new function.

2. But what is the point of merging them? The reason why there is no implicit folding in C3 is to communicate more clearly if constant evaluation happens or not. While in some cases we want to blur this, I don't see the point in this case...?

NotsoanoNimus:
Seems sensible to me - I merged back a few of the functions that were there, but kept `combine_sets` neutral to either type of invocation (see the unit tests for a possibly-good reason).

I will keep in mind going forward that `@const` and runtime stuff should continue to have a more explicit delineation in the stdlib. Thanks!

Edit: I actually think `contains` might warrant the dynamic approach still, the more I think about it. Because it's such a small operation that one doesn't really need to explicitly change between compile-time and runtime at all. What do you think, @lerno?

lerno:
Thank you!

PR #2701, opened 2026-01-01
Alphabetise import statements

mikomikotaishi:
The import statements in `lib/std` are not alphabetised, and some are all on one line while others are on a separate line. This PR addresses this by alphabetising the statements, which is hopefully easier to maintain.

### DIFF
--- a/lib/std/atomic.c3
+++ b/lib/std/atomic.c3
@@ -1,7 +1,7 @@
 // Copyright (c) 2023-2025 Eduardo José Gómez Hernández. All rights reserved.
 // Use of this source code is governed by the MIT license
 // a copy of which can be found in the LICENSE_STDLIB file.
-module std::atomic::types{Type};
+module std::atomic::types {Type};
 
 struct Atomic
 {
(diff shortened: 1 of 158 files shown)


lerno:
Import statements should in general be on a single line.

mikomikotaishi:
I'm not sure why the CI tests are failing, seeing as I didn't make any changes to the compiler itself, only the standard library. The errors don't seem to mention anything relating to that either.

Also, the source files originally didn't have them all on one line, which is why I moved them into one import per line.

ManuLinares:
> I'm not sure why the CI tests are failing, seeing as I didn't make any changes to the compiler itself, only the standard library. The errors don't seem to mention anything relating to that either.
> 
> Also, the source files originally didn't have them all on one line, which is why I moved them into one import per line.

I've run the test suite locally with this pull, it fails. It touched 158 files, it's up to you to fix/test before/after pull. The test failures stem from nondeterministic LLVM debug info, maybe you can update the tests #expect, and also what lerno said.

lerno:
I'll have to reject this one. Although there are some useful formatting improvements, it operates under the faulty assumption that it is desirable to expand import statements on multiple lines and don't have them single line. This is wrong, so consequently everything then fails on that.

PR #2718, opened 2026-01-06
Fixed macro hash_vec(vec): it was checking only the first bytes (4 or 8, not sure)

mikelgarai:
Hi!

I was trying to use the hash of a ulong[<4>] for a custom compact hashmap I was doing and noticed the hash values were terrible, I did a bit of testing and came across this. I believe the len is not suppose to be the size of the type returned by the operation, but the operation itself.

Please let me know if I'm mistaken.

Also, I'm just realizing, do you want me to add some unitary tests? I'm not sure how to run them though (very new to C3, just learning for fun by doing AoC 2022 xD )

the reproducer is as easy as 
```	
ulong[<6>] a = {1,2,3,4,5,6};
ulong[<6>] b = {1,2,3,4,6,6};
assert(a.hash() != b.hash());
```

Cheers!

Mikel

### DIFF
--- a/releasenotes.md
+++ b/releasenotes.md
@@ -57,6 +57,7 @@
 - Creating recursive debug info for functions could cause assertions.
 - bitorder::read and bitorder::write may fail because of unaligned access #2734.
 - Fix `LinkedList.to_format` to properly iterate linked list for printing.
+- Hashing a vector would not use the entire vector in some cases.
 
 ### Stdlib changes
 - Add `ThreadPool` join function to wait for all threads to finish in the pool without destroying the threads.


lib/std/core/builtin.c3
```diff
@@ -700,7 +700,7 @@ macro uint hash_array(array_ptr) @local
 *>
 macro uint hash_vec(vec) @local
 {
-	var $len = $sizeof(vec.len * $typeof(vec).inner.sizeof);
+	var $len = vec.len * $typeof(vec).inner.sizeof;
```
lerno: The reason for this one is that there was padding for old simd vectors, so that an `int[<3>]` would have an extra int at the end. This doesn't happen anymore, so just `$sizeof(vec)` should be fine

mikelgarai:
Done! I tested it (only with `ulong[<6>]` though) and it works fine.

I tried finding unit tests for this but didn't find any, let me know if you want me to add some.

lerno:
Unit tests would be great as it prevents errors when refactoring.

mikelgarai:
should I add the tests to `test/unit/stdlib/core/builtintests.c3` -> `test_hash()`, `test_hash_repeat()` or create a new function?

also, I see there are a lot of reds on the CI, but they don't seem to be related to my change (at least to the best of my knowledge) (didn't click "all" the reds though) the ones I checked are in some wrapper.cpp file. Should I rebase my branch or something like that?

lerno:
Add a new test function in there which tests this.

mikelgarai:
first time using macros, so something might be off. I did do some debugging by adding an assert(false) and printing and I think it does what it should do, but still, macros are confusing to me :-D

EDIT: what it tries to do is compare 2 vectors that differ only in a single bit in a single element, and does that for each element of the vector, to vectors of all types and vectors of many sizes (didn't do all of them because it was probably too much time, just powers of 2 up to a point and some interesting odd and prime values) while testing I noticed that for vectors < 4 bytes it was also failing before my original change with hash() that was undefined.

mikelgarai:
BTW, not experienced at all on github ettiquete , do I have to be one submitting or you'll do it? (not sure if I even can do it, is it the "update branch" button?)

lerno:
Thank you!

PR #2720, opened 2026-01-06
Ignore ctags and gtags files in .gitignore

frapank:
This PR updates the .gitignore file to exclude automatically generated tags files from ctags and gtags.

Specifically, it ignores:
- ctags output file: tags
- GNU Global output files: GTAGS, GRTAGS, GPATH


lerno:
Thank you

Issue #2723, opened 2026-01-07
`libc` is missing signals


jawlms:
`libc` is currently missing at least these signals:

- `SIGEMT` (BSD)
- `SIGINFO` (BSD)
- `SIGIO` (Linux, BSD)
- `SIGIOT` (Linux, BSD)
- `SIGPOLL` (Linux)
- `SIGPROF` (Linux, BSD)
- `SIGPWR` (Linux)
- `SIGSTKFLT` (Linux)
- `SIGTHR` (BSD)
- `SIGTTIN` (POSIX)
- `SIGTTOU` (POSIX)
- `SIGUNUSED` (Linux)
- `SIGUSR1` (POSIX)
- `SIGUSR2` (POSIX)
- `SIGVTALRM` (POSIX)
- `SIGWINCH` (POSIX)
- `SIGXCPU` (POSIX)
- `SIGXFSZ` (POSIX)

lerno:
These are now merged.

PR #2724, opened 2026-01-07
Rework `libc` signal constants

### LINKED ISSUE #2714: `libc` signals need restructuring
Currently, all signal constants (`libc::SIG*`) are defined unconditionally in [libc.c3](https://github.com/c3lang/c3c/blob/master/lib/std/libc/libc.c3#L60-L80).

Since the availability of signals and their values are OS-dependent, it would make sense to move these definitions into the respective `@if(env::OS)` sections of module `libc`, possibly into the corresponding files in [libc/os/](https://github.com/c3lang/c3c/tree/master/lib/std/libc/os), and update their values accordingly.

jawlms:
Fixes #2714 and #2723.

The signal constants have been split up to conditionally compiled sections of `libc`, so that for every operating system only available signals are defined. Missing signals have been added for POSIX, Linux and BSD systems.

### DIFF
--- a/lib/std/libc/libc.c3
+++ b/lib/std/libc/libc.c3
@@ -57,6 +57,13 @@ alias Fd = CInt;
 alias Fpos_t @if(env::WIN32) = long;
 alias SignalFunction = fn void(CInt);
 
+// Common signals for all targets
+const CInt SIGINT = 2;
+const CInt SIGILL = 4;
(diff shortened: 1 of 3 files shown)


lerno:
I prefer if the common ones are shared in libc. That way it's easier to see what's common, so I moved them back.

lerno:
Thank you!

PR #2725, opened 2026-01-07
fix dwarf crash

ManuLinares:
```c3
fn void main(String[] args){
    LoadFn lfn;
}
alias LoadFn     = fn Result();
alias CompleteFn = fn void (Result result);
typedef Result = CompleteFn;
```

llvm keeps the pointer we give it, we reused and modified that memory later on recursion. llvm then read garbage and crashed. Now each function gets its own memory, so nothing gets corrupted.


ManuLinares:
fixed in https://github.com/c3lang/c3c/commit/8fa59cbb432f0e2d15dad44cd963c8df0140de5a

PR #2729, opened 2026-01-08
Require parenthesized assignment expressions in condition of 'if' statements #2716

### LINKED ISSUE #2716: Require `if (a = b)` to have extra parens
Require `if ((a = b))` in conds

VarunVF:
Fixes #2716.

I've added a small test to cover both the error case and allowed case.


lerno:
Change the test file to .c3, .c3t is usually for files that are expected to compile.

lerno:
I moved the analysis to the semantic checker which also handles while, and added a note in the release notes and fixed the broken tests.

PR #2749, opened 2026-01-12
Fix minor typos in README.md

Darvisim:
There were few typos I spotted in the README. Also, I have updated all the fenced code blocks containing C3 code by replacing cpp or c++ with c3.


lerno:
Thank you

PR #2751, opened 2026-01-13
Add $$VERSION and $$PRERELEASE

lyranie:


lerno:
There is no need to implement them like this. Look at the compile function in compiler.c. It defines a bunch of built in constants:
```c
	setup_int_define("C_SHORT_SIZE", compiler.platform.width_c_short, type_int);
	setup_int_define("C_INT_SIZE", compiler.platform.width_c_int, type_int);
	setup_int_define("C_LONG_SIZE", compiler.platform.width_c_long, type_int);
	setup_int_define("C_LONG_LONG_SIZE", compiler.platform.width_c_long_long, type_int);
```
Use this instead. It's a single line to add a constant. Also, please add something to the release notes.

PR #2752, opened 2026-01-13
small fix to generated C header global

ManuLinares:
test:
```c3
module my_module;

alias Something = fn void(int);
Something something @export("something") = &something_else;
fn void something_else(int a) => {};
```

would generate
```c
/* GLOBALS */
my_module__Somethingextern  something;
```

now correctly generates
```c
/* GLOBALS */
extern my_module__Something  something;
```


lerno:
Thank you!

PR #2753, opened 2026-01-13
collections: fix LinkedList.to_format

konimarti:
For-loop in LinkedList.to_format goes brrrrrr and prints the value of the first node an infinte number of times because it does not properly iterate the linked list.

Fix the list iterations and add the missing string format specifier.

Bug can be reproduced with:
```
import std;
fn void main() => @pool()
{
	io::printfn("%s", linkedlist::@tnew{usz}({1,2,3}));
}
```


lerno:
Thank you!

PR #2760, opened 2026-01-15
Add mem_allocator `realloc_array` Macros

NotsoanoNimus:
Noticed these were missing when trying to slightly grow a statically-allocated array at runtime.

### DIFF
--- a/lib/std/collections/enummap.c3
+++ b/lib/std/collections/enummap.c3
@@ -1,5 +1,6 @@
 <*
  @require Enum.kindof == TypeKind.ENUM : "Only enums may be used with an enummap"
+ @require Enum.values.len > 0 : "Only non-empty enums may be used with enummap"
 *>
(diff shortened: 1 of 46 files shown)


lerno:
Did you verify that realloc to zero length works?

NotsoanoNimus:
> Did you verify that realloc to zero length works?

Is the expected behavior to essentially free the array and return `null`? After writing a unit test, which I can push depending on your reply, I can confirm that's what happens.

Edited to add: Wait, I just checked under `realloc_try` and I see:
```c3
	if (!new_size)
	{
		free(allocator, ptr);
		return null;
	}
```
So looks to be following the expected behavior.

lerno:
Can you update the releasenotes as well?

NotsoanoNimus:
> Can you update the releasenotes as well?

All set. :+1:

lerno:
Thank you!

PR #2762, opened 2026-01-15
std/io/os/temp_directory.c3: fix INVALID_PATH in Win32 native_temp_directory

ManuLinares:
- Use the actual length from GetTempPathW for Windows temp path slice
- We can remove the workaround in the test_suite_runner for WIN32 and create all directories in %temp% now

### DIFF
--- a/releasenotes.md
+++ b/releasenotes.md
@@ -60,6 +60,7 @@
 - bitorder::read and bitorder::write may fail because of unaligned access #2734.
 - Fix `LinkedList.to_format` to properly iterate linked list for printing.
 - Hashing a vector would not use the entire vector in some cases.
+- Fix to `temp_directory` on Windows #2762.
 
 ### Stdlib changes
 - Add `ThreadPool` join function to wait for all threads to finish in the pool without destroying the threads.


lib/std/io/os/temp_directory.c3
```diff
@@ -122,9 +122,10 @@ fn Path? native_temp_directory(Allocator allocator) @if(env::WIN32) => @pool()
 {
 	Win32_DWORD len = win32::getTempPathW(0, null);
 	if (!len) return io::GENERAL_ERROR?;
-	Char16[] buff = mem::talloc_array(Char16, len + (usz)1);
-	if (!win32::getTempPathW(len, buff)) return io::GENERAL_ERROR?;
-	return path::new(allocator, string::tfrom_utf16(buff[:len]));
+	Char16[] buff = mem::talloc_array(Char16, len);
```
lerno, line 125: Does len include the zero terminator?

lerno: I checked the docs and it does so, good.

lerno:
Thank you for fixing this!

PR #2765, opened 2026-01-16
Fix function signature for free to return void

Darvisim:
`free` expects a return type `void*` but it should return `void` instead of `void*`


lerno:
Thank you!

PR #2768, opened 2026-01-17
use libc `stat` on more platforms to get file sizes

Book-reader:
Suggested by @nomota on the discord. Linux, Android, and BSD all support using `stat` to get the file size


lerno:
Great!

PR #2772, opened 2026-01-18
Add maximum memory usage tracking to tracking allocator

m0tholith:
I think this is a helpful metric.
image
I would have liked to have the option to collect more data (perhaps a graph for memory usage over time) but it's kind of overkill for a single PR. I could do it but I'd like to have more thought poured into it. I'll open an issue about it at a later time.

### DIFF
--- a/lib/std/collections/enummap.c3
+++ b/lib/std/collections/enummap.c3
@@ -1,5 +1,6 @@
 <*
  @require Enum.kindof == TypeKind.ENUM : "Only enums may be used with an enummap"
+ @require Enum.values.len > 0 : "Only non-empty enums may be used with enummap"
 *>
 module std::collections::enummap <Enum, ValueType>;
 import std::io;
--- a/lib/std/compression/qoi.c3
+++ b/lib/std/compression/qoi.c3
@@ -292,17 +292,17 @@ fn char[]? decode(Allocator allocator, char[] data, QOIDesc* desc, QOIChannels c
 	if (bswap(header.be_magic) != 'qoif') return INVALID_DATA~;
 
 	// copy header data to desc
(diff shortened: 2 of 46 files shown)


lerno:
Please update the releasenotes as well?

lib/std/core/allocators/tracking_allocator.c3
```diff
@@ -88,27 +95,34 @@ fn void*? TrackingAllocator.acquire(&self, usz size, AllocInitType init_type, us
...
 	self.mem_total += size;
+	self.usage += size;
+	if (self.usage > self.max_usage) self.max_usage = self.usage;
 	return data;
 }
 
 fn void*? TrackingAllocator.resize(&self, void* old_pointer, usz size, usz alignment) @dynamic
 {
 	void* data = self.inner_allocator.resize(old_pointer, size, alignment)!;
+	self.usage -= self.map[(uptr)old_pointer]!!.size;
```
data-man, line 106: I think `-=` is incorrect here.

m0tholith: How come? The old pointer gets resized from `A` bytes to `B` bytes. The usage already tracks the `A` bytes, so that gets subtracted, and later in the function the new size of `B` bytes gets added.

m0tholith:
Added release notes. No issues

lib/std/core/allocators/tracking_allocator.c3
data-man: Resizing != decreasing only.

m0tholith: I add the new size at line 112

lerno:
Thank you!

Issue #2727, opened 2026-01-08
env missing SSE4_2


nomota:
X86 after 2008 supports SSE4.2
which is used by Windows11 and high-speed databases, which is for fast data/string comparison.

env:: module doesn't support env::SSE4_2

on linux,
```
grep -o 'sse4_2' /proc/cpuinfo | head -1
```

ManuLinares:
https://github.com/c3lang/c3c/blob/master/lib/std/core/private/cpu_detect.c3#L249

```c3
import std::core::cpudetect;
import std::io;

macro bool has_feat(f) => (cpudetect::x86_features & (1ULL << f.ordinal)) != 0;

fn void main()
{
    // Initialize the feature detection
    cpudetect::x86_initialize_cpu_features();

	switch {
		case has_feat(X86Feature.FMA4):
		  io::printn("FMA4: Supported");
		case has_feat(X86Feature.SSE4_2):
		  io::printn("SSE4_2: Supported");
	}
}
```

I really don't understand the filing of this issue. The above code is an example of how to detect a feature. If your question is if the compiler supports optimizing your code, yes depending on the optimization level you provide with `-O` flag, llvm detects your cpu capabilities and optimizes.

I once again invite you to the discord channel for questions regarding c3, and only file __issues__ that pertain to the compiler or standard library itself.

nomota:
That's a runtime detection.

What's needed is a compiletime FLAG for conditional compilation.

ManuLinares:
> That's a runtime detection.
> 
> What's needed is a compiletime FLAG for conditional compilation.

Why would this need to be compile-time?

If the goal is to force a binary to be built with or without a specific CPU feature, C3 already supports that.

For compile-time conditional code, you should use feature flags:

- $if $feature("…") { ... }
- Features can be added/removed passing **-D** / **-U** to the compiler: [source](https://c3-lang.org/generic-programming/compiletime/#feature)

If the goal is to control what LLVM is allowed to emit, that's handled by compiler flags, not the env module, for example:

- --x86cpu=baseline : Disables SSE2+ (lowest x86-64 target)
- --x86vec=none : Prevents SIMD codegen entirely (no SSE/AVX)
- **--cpu-flags=-sse4.2** (or -sse2, etc.) : Explicitly removes individual CPU features <-------- this is likely what you want

Could you provide an example of your use case? Your issue description is lacking in detail

ManuLinares:
[...]

PR #2773, opened 2026-01-19
Add Argon2 memory-hard hashing and encoding

NotsoanoNimus:
Implements Argon2 in accordance with RFC 9106 and establishes a canonical way to encode/decode Argon2 hash strings with their included options. This is a known standard, the [PHC-SF](https://github.com/P-H-C/phc-string-format/blob/e8fbd333dcc9a8b0843fac6b33371cf157e91a48/phc-sf-spec.md).

Also includes a small rework of the inflexible way BLAKE2 hashing worked, in order to accommodate _runtime_ hash lengths. This should be acceptable without deprecation notices since this change comes within the same compiler/stdlib version as BLAKE2's implementation.

If you want to know how to use Argon2, please see the related unit-test file.

Future versions of this software should aim to make use of SIMD optimizations for each parallel processing lane, and for the underlying BLAKE2-like round function. At the moment, this sort of optimization is left for the C3 compiler to take care of.

### DIFF
--- a/releasenotes.md
+++ b/releasenotes.md
@@ -25,6 +25,7 @@
 - Add `to_format` functionality for DateTime.
 - `SubProcess`/`process::create`/`process::execute_stdout_to_buffer` deprecated, replaced by `Process`/`process:spawn`/`process::run_capture_stdout`.
 - Add support for AES-encrypted Zip files (AE-1 and AE-2 formats).
+- Add `Argon2` memory-hard hashing with associated tests. #2773
 
 ### Fixes
 - `@deprecated` in function contracts would be processed twice, causing a compilation error despite being correct.
@@ -70,7 +71,6 @@
 - Typekind enums are changed CONST_ENUM -> CONSTDEF, DISTINCT -> TYPEDEF.
 
 ### Stdlib changes
-- Add `Argon2` memory-hard hashing with associated tests. #2773
 - Summarize sort macros as generic function wrappers to reduce the amount of generated code. #2831
 - Remove dependency on temp allocator in String.join.
 - Remove dependency on temp allocator in File.open.


ManuLinares:
Nice, I'll test this later.

Just a reminder that https://github.com/c3lang/c3c/blob/master/lib/std/hash/blake3.c3#L67 isn't working and to replace it with something like https://github.com/c3lang/c3c/issues/2727#issuecomment-3721625487

NotsoanoNimus:
Great! Thanks in advance, @ManuLinares.

And I appreciate the reminder: I have that code change hanging around in my SHA-3 branch but I haven't pushed it up yet. My next target is to finally finish that PR and open it for review, so it should be fixed then.

lib/std/hash/blake2.c3
```diff
@@ -132,13 +132,15 @@ macro @common_update(#instance, $block_size, char[] data) @local
...
+<*
+ @param [out] into : "The destination for the resultant hash. Its length must be at least the size of the hash's initialized size."
 
+ @require into.len >= #instance.outlen : "The result slice must be large enough to accept the full hash digest."
+*>
+macro @common_final(#instance, char[] into) @local
+{
 	@add_ctr(#instance, #instance.buflen);
-	if (#instance.f[0]) return result; // technically an error return
+	// TODO:  if (#instance.f[0]) return result; // technically an error return
```
lerno: TODOs make me worried. Keep or not?

NotsoanoNimus: Whereas before this would return an empty array, it now just doesn't error out. However, this isn't a concern since the condition could only ever be `true` in the event that the instance is uninitialized or there's some other context corruption. The `init` method explicitly writes zeroes to the instance anyway.

I think it's safe to delete this line, so I will do so. I marked it TODO for my review and got pulled into the rest of the vortex.

lerno:
We got a bunch of tests that are failing because unfortunately they are comparing with code that depends on the line numbers of certain functions, the right way to fix this is to cut away the line number parts of the comparison text.

lerno:
When compiled with `--print-large-functions`:

```
   14009 instructions found in std::crypto::argon2:decode (/Users/lerno/Projects/c3c/lib/std/crypto/argon2.c3) - function is very long.
   24059 instructions found in std::crypto::argon2:compress (/Users/lerno/Projects/c3c/lib/std/crypto/argon2.c3) - function is very long.
```
10 instructions are about 1 LOC, so those are 1400 and 2400 lines long functions. Very long functions are relatively slow to codegen in LLVM and may end up less efficient due to cache concerns.

lerno:
This unfortunately now have some merge conflicts.

NotsoanoNimus:
> This unfortunately now have some merge conflicts.

👍 Will resolve, along with reducing instruction counts, when I return to my workstation in a few hours.

ManuLinares:
[argon2.patch](https://github.com/user-attachments/files/25472515/argon2.patch)

one more thing you could do, extract the logic for parsing a single parameter into a separate helper

lerno:
Can there be some update on this one?

lerno:
Tests are not passing on MacOS.

NotsoanoNimus:
@lerno Unit tests for macOS should now be fixed. Alignment bug turned out to be an artifact of my very first implementation of the Argon2 algorithm.

I've also incorporated one more refactor courtesy of @ManuLinares. This is now ready to merge. :crossed_fingers:

lerno:
Thank you!

Issue #2808, opened 2026-01-23
Cryptographic randomness source


laura240406:
As far as I can tell, C3 doesn't have a way to read cryptographically secure random bytes right now.

Here is how you would do it on every supported platform:
Windows: CryptGenRandom or BCryptGenRandom
MacOS, iOS, Linux, FreeBSD, NetBSD, OpenBSD: read /dev/random or /dev/urandom
ELF, WASM, MCU: no idea

ManuLinares:
Do you mean in the stdlib? Maybe you could sketch a `native_random()` macro

This is a minimal example for linux x86_64:

```c3
import std;

fn int main()
{
	char[32] buf;
	isz result = $$syscall(318, (uptr)&buf, buf.len, 0);
	if (result != buf.len) return 1;
	io::printn(buf);
	return 0;
}
```

lerno:
There was a PR for it but it was abandoned.

lerno:
This is now merged.

PR #2812, opened 2026-01-23
Add cryptographically secure random bytes source

laura240406:
My proposal implementation for #2808. The WIN32 version is untested as it seems to be impossible for me to obtain the Windows SDK as it's buried in layers of MSIs that Wine can't deal with.


laura240406:
you can test this with

```c3
import std::io;
import std::crypto::random;

fn void main()
{
    char [64] key;
    random::crypto_secure_random(&key)!!;

    io::printn(key);
}
```

ManuLinares:
fopen, fread, fclose is not a good way to get a random. Have you taken a look at my example in https://github.com/c3lang/c3c/issues/2808, syscall is much faster.

👎

laura240406:
It's much more portable as it also works on the BSDs and Darwin and the syscall number of getrandom() depends on the architecture.

laura240406:
idk why it just closed the PR but I've added the getrandom() call for Linux/x86_64

laura240406:
It makes sense to add OS/Architecture specific cases to use the syscall but the file method should remain as a fallback.

laura240406:
The current implementations are:

Linux/Android x86_64/Aarch64: syscall
Linux x86: syscall
Win32: call RtlGenRandom (basically like a syscall)
Android/Linux/*BSD/Darwin: read /dev/random or /dev/urandom if the first two didn't work

laura240406:
I'm also currently installing Windows 10 in a VM to test the Win32 implementation.

Book-reader:
The stdlib has a random interface that can be used through functions like `random::next_float(&my_rand)`: https://github.com/c3lang/c3c/blob/396263f5c3ef4c933f5bf6d82154788636838fd3/lib/std/math/random.c3#L165-L174
Maybe this could be added under https://github.com/c3lang/c3c/tree/master/lib/std/math/random as a type that implements this interface? Though I don't know if this form of random would be able to fulfil the expectations of the `Random` interface (specifically setting the random seed) so it's up to you if you want to do that.

laura240406:
No the idea is to use this to derive cryptographically strong keys.
We could, however, add some helper that initializes such a Random instance from a strong seed.
I could think of something like RSA key generation where we read a cryptographically strong seed, put it into some Random implementation and use that to then sample primes until we have 2.

I originally started this since I wanted to implement RSA for the stdlib but then I realized there is no way to generate those random primes safely.

laura240406:
I wonder if we should add an option to use `RDRAND` on x86 platforms that support it since it should be faster than getrandom() and still produce adequate randomness.

laura240406:
Okay from my tests, the Win32 version works on Windows 10 and under Wine.

ManuLinares:
@laura240406 Pull requests aren't meant for ongoing experimentation. Please keep PRs focused on a specific requested change, or a complete implementation. For discussion or exploration, let's use issues, discussions or discord chat instead.

laura240406:
okay but this PR as it currently stands should be done

it has syscall implementations for the main platforms of Linux/Android (x86_64/x86/Aarch64) and the library implementation of Win32 and the fallback of /dev/urandom

a later PR could add *BSD specific getrandom() implementations and x86 RDRAND implementations but that's out of scope rn

laura240406:
I've also tested it on OpenBSD (uses /dev/urandom) and it works

lib/std/crypto/random.c3
```diff
@@ -0,0 +1,95 @@
+module std::crypto::random;
+import std::io::file;
+import std::os::win32;
+
+faultdef SECURE_RANDOM_UNAVAILABLE;
+
+alias RtlGenRandom = fn CInt(void *, ulong);
+RtlGenRandom rtlgenrandom @if(env::WIN32) @local;
+
+fn void crypto_init_win32_random() @if(env::WIN32) @init
```
lerno, line 10: This should probably not be `@init`, since that will make ALL win32 apps do this. Instead, the correct way is to run this once. You can use OnceFlag.call_once in std::thread to ensure it's done once across all threads.

lerno:
Can you add a test for this to the unit tests, and then add a note in the releasenotes as well please?

lerno:
Ping

lerno:
I still can't merge this due to lack of testcases.

lerno:
Also it should be added to the releasenotes

lerno:
If no one has any interests in adding test cases and releasenotes I'll close this PR without merging.

lerno:
Closing due to lack of activity and requested test cases

PR #2670, opened 2025-12-22
add Win32_CODEPAGES enum and enable utf8 console output on win32

Book-reader:
this adds an enum of all the win32 codepages from https://learn.microsoft.com/en-us/windows/win32/intl/code-page-identifiers and sets the output codepage to utf8 with an `@init` function so that utf8 characters can be printed correctly (see #2668)

PR #2814, opened 2026-01-24
also set utf8 codepage on win32 when the main function takes no args

Book-reader:
in https://github.com/c3lang/c3c/pull/2670#issuecomment-3766496033 the reading input and printing utf8 tests work apart from 1 and 4, which are the `fn void main()` and `fn int main()` tests from https://github.com/c3lang/c3c/pull/2670#issuecomment-3764518823.

in sema_decls.c it calls the `@main_to_int_main` and `@main_to_void_main` macros when the main function has no args and isn't `@winmain`, so this PR also sets the utf8 codepage in these macros and fixes printing utf8 text for main functions with no args:
https://github.com/c3lang/c3c/blob/396263f5c3ef4c933f5bf6d82154788636838fd3/src/compiler/sema_decls.c#L4136-L4154

it would also be possible to add `@wmain_to_void_main` and `@wmain_to_int_main` macros that get called instead of  `@main_to_int_main` and `@main_to_void_main` on win32, I can do that instead if you prefer.

### DIFF
--- a/lib/std/core/private/main_stub.c3
+++ b/lib/std/core/private/main_stub.c3
@@ -77,6 +77,23 @@ macro win32_set_utf8_codepage() @local
 	win32::setConsoleOutputCP(UTF8);
 }
 
+macro int @wmain_to_err_main(#m, int, Char16**)
+{
+	win32_set_utf8_codepage();
+	if (catch #m()) return 1;
+	return 0;
+}
+macro int @wmain_to_int_main(#m, int, Char16**)
+{
+	win32_set_utf8_codepage();
+	return #m();
+}
+macro int @wmain_to_void_main(#m, int, Char16**)
+{
+	win32_set_utf8_codepage();
+	#m();
+	return 0;
+}
 
 extern fn Char16** _win_command_line_to_argv_w(ushort* cmd_line, int* argc_ptr) @cname("CommandLineToArgvW");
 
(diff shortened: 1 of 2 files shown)


lerno:
Better have it completely separate for windows.

lerno:
So add the @wmain_to_void_main and @wmain_to_int_main macros please

lerno:
I did some updates to master, can you check if this actually covers your cases?

Book-reader:
I've been able to borrow a windows PC to test these changes myself.
On master (120f56ac5fe3bf9a2c18336431c963564ad80759) `fn void main(String[] args) => io::printn("你好世界！");` prints correctly, but `fn void main() => io::printn("你好世界！");` does not. Using this PR both print correctly.

lerno:
I made a mistake. Try it now please.

Book-reader:
On master (21758476d43ba894e5489802608f3cc88c084921) it now works for everything except
```c3
fn int main()
{
io::printn("你好世界！");
return 0;
}
```

lerno:
Missed that case. Now?

Book-reader:
It now works for every `main` type except `fn int main(int argc, char** argv)` since there is no forwarding function for it, which might be fine because it's the raw libc main type?.

lerno:
Yes, I think there should be a way on windows to avoid the forwarding and be able to compile with --use-stdlib=no

Book-reader:
Then I can close this

PR #2831, opened 2026-01-24
[stdlib] Reduce inline code volume from sorting macros

NotsoanoNimus:
Reinvent `EmptySlot` but specifically for stdlib sorting modules. Still permits the eventual removal of the stdlib `EmptySlot` builtin later on.

This change allows the various permutations of missing parameters to still apply in wrapper macros, while also not letting macros run wild within dependent code. While each parameterized combination of sorting modules might still generate different outputs, repeated calls to each generated sorting function will no longer be placed inline.


lerno:
Thank you!

PR #2834, opened 2026-01-24
Added Mt19937 random number generator.

laura240406:


ManuLinares:
as far as I can tell, MT19937 is already obsolete compared to modern algorithms like Xoshiro, a broken, non-compliant implementation serves no purpose here.

also, this seems like ai slop, idk

laura240406:
> as far as I can tell, MT19937 is already obsolete compared to modern algorithms like Xoshiro, a broken, non-compliant implementation serves no purpose here.

okay fair, I will try to implement xoshiro tomorrow

> also, this seems like ai slop, idk

what, that is a mean accusation
the code is just a C3 translation of [the Wikipedia implementation](https://en.wikipedia.org/wiki/Mersenne_Twister#C_code)

ManuLinares:
My apologies for the 'slop' comment; some of the logic errors made it look like generated code rather than a manual port.

laura240406:
all good :)

PR #2839, opened 2026-01-25
stdlib: optimize math::next_power_of_2 to O(1) using hardware CLZ

ManuLinares:
- updated DString.reserve to use this

### DIFF
--- a/lib/std/core/dstring.c3
+++ b/lib/std/core/dstring.c3
@@ -647,8 +647,8 @@ fn void DString.reserve(&self, usz addition)
 	usz len = data.len + addition;
 	if (data.capacity >= len) return;
 	usz new_capacity = data.capacity * 2;
-	if (new_capacity < len)          new_capacity = math::next_power_of_2(len);
 	if (new_capacity < MIN_CAPACITY) new_capacity = MIN_CAPACITY;
+	if (new_capacity < len) new_capacity = math::next_power_of_2(len);
 	data.capacity = new_capacity;
 	*self = (DString)allocator::realloc(data.allocator, data, StringData.sizeof + new_capacity);
 }


lib/std/math/math.c3
```diff
@@ -997,11 +997,19 @@ macro bool is_power_of_2(x)
...
+ Uses hardware instructions (CLZ) for O(1) performance.
+
+ @require types::is_int($typeof(x)) : "Input must be an integer"
+*>
 macro next_power_of_2(x)
 {
-	$typeof(x) y = 1;
-	while (y < x) y += y;
-	return y;
+	if (x <= 1) return 1;
```
lerno, line 1008: What about the classic solution 

```c
x -= 1;
x |= (x >> 1);
x |= (x >> 2);
x |= (x >> 4);
x |= (x >> 8);
x |= (x >> 16);
```

BTW `if (x <= 2) return x` might take one more?

ManuLinares: maybe something like this?

```c
<*
 Returns the next power of two that is greater than or equal to x.

 @require types::is_int($typeof(x)) : "Input must be an integer type"
*>
macro next_power_of_2(x)
{
    if (x <= 1) return 1;
    if (x == 2) return 2;

    $typeof(x) v = x - 1;
    
    v |= v >> 1;
    v |= v >> 2;
    v |= v >> 4;

    $if ($sizeof(x) >= 2):  v |= v >> 8;  $endif;
    $if ($sizeof(x) >= 4):  v |= v >> 16; $endif;
    $if ($sizeof(x) >= 8):  v |= v >> 32; $endif;
    $if ($sizeof(x) >= 16): v |= v >> 64; $endif;

    return v + 1;
}
```

lerno: Something like that yes. But I think `if (x <= 2) return x;` would be shorter when optimized? Since it just returns the thing in the register.

lerno: Depending on what we want to do with 0 of course. Negative numbers should probably be contract failures.

ManuLinares: x=0 should return 1 in my opinion. Zero is not a "power of 2"

lerno:
Thank you!

PR #2846, opened 2026-01-26
Added Xorshiro128++ random number generator.

laura240406:
As you wished for something more modern than MT19937, I implemented Xorshiro128++.

This is also one of the RNGs (the other one being Java's LCG) that [is used by Minecraft](https://minecraftathome.com/minecrafthome/forum_thread.php?id=204) to generate the world in newer versions and should be statistically good while being very fast.

### DIFF
--- a/lib/std/io/io.c3
+++ b/lib/std/io/io.c3
@@ -50,31 +50,17 @@ faultdef
  or to the end of the stream, whatever comes first.
  "\r" will be filtered from the String.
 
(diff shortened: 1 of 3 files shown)


lerno:
Could you add tests as well please?

lerno:
Adding tests for this, I found two differences compared to the standard C implementation here: https://prng.di.unimi.it/xoroshiro128plusplus.c  which I corrected.

PR #2858, opened 2026-01-27
[stdlib] Add PEM Encoding/Decoding Module

NotsoanoNimus:
Adds a PEM encoding/decoding module to the standard library. This will enable C3-based interactions with DER-encoded certificates, signatures, key files, and other items.

The unit tests provide full coverage over single- or multi-PEM operations, along with two leak checks and tests with invalid PEM data.

Also adds `tjoin` to `string.c3`, because there wasn't a temp-allocator version available, nor was there a doc comment over `join`.

### DIFF
--- a/lib/std/encoding/pem.c3
+++ b/lib/std/encoding/pem.c3
@@ -24,6 +24,8 @@ const String PRE_EB_PREFIX @local = EB_DELIMITER +++ "BEGIN ";
 <* All PEM blobs will terminate with this Encapsulation Boundary prefix. *>
 const String POST_EB_PREFIX @local = EB_DELIMITER +++ "END ";
 
+alias PemHeader = String[2];
+
 <* Specify a set of possible PEM en/decoding faults. *>
 faultdef
 	BODY_REQUIRED,   // encoding: no body given (or too few of them)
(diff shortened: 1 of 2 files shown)


lerno:
This one is using a lot of macros but it seems that this is unnecessary. Rather than using macros with `...`, use `String... args` and `String[]... args`

NotsoanoNimus:
> This one is using a lot of macros but it seems that this is unnecessary.

The macros from this library are intended to be used in calling projects only a single time, typically. Regardless, I'll try to get a change worked out today, as I have no issue with changing it.

But it seems that this "wrap-the-body-in-a-function" is a ceremony in C3 which affects macros overall, that might be better solved by using some new macro attribute -- e.g., its own `@noinline` variant -- which makes this simpler when the body of a macro is excessive. I know this is likely a tall order, but the solution is warranted IMO.

NotsoanoNimus:
I should also note here that a runtime splat is much less flexible than what's currently defined - as I'm rewriting this and dealing with non-const splats, it feels like I'm actually diminishing the usability of the library. I think I'll need to just use `String[]` for most calls if these macros really are unacceptable, unfortunately.

lerno:
I've done some modifications to this now. I honestly found it easy to extract functions from the macros. In addition to this, the main use of the macro vaargs seems to be able to use the hashmap init functions. However, using those would lead to use-after-free bugs, since the values in a hash map need to be explicitly copied (and freed).

lerno:
It's worth considering whether key-values should be grouped, so 
```c3
	String result = pem::tencode(SINGLE_DATA, "CERTIFICATE", 
		{ "My-Header", "Example" }, 
		{ "This-is-a-longer-header", "with some extra content that will bleed over to the next line." }, 
		{ "Last", "Header is single-line." })!!;
```

lerno:
I'll merge this for now though.

lerno:
Hmm I updated the API anyway.

Issue #554, opened 2022-09-05
Inline asm support all x64 functions


lerno:
List here: https://www.felixcloutier.com/x86/

data-man:
And here:
https://github.com/netwide-assembler/nasm/blob/master/x86/insns.dat :)
https://github.com/netwide-assembler/nasm/blob/master/x86/regs.dat

And very big (105M): https://uops.info/instructions.xml

lerno:
It's pretty easy to add more instructions so I appreciate contributions there.

lerno:
x64/x86 opcodes are added here: 
https://github.com/c3lang/c3c/blob/master/src/compiler/asm_target.c in `init_asm_x86`
While the code needs some work to support vector and floats + addresses, all the general register opcodes can be added (and the rest could be added by commented out with pseudo code). This only needs work looking through the intel manuals.

lerno:
Also: https://docs.oracle.com/cd/E53394_01/pdf/E54851.pdf

lerno:
https://fastapi.metacpan.org/source/BOGDRO/Asm-X86-0.62/lib/Asm/X86.pm

lerno:
https://docs.google.com/document/d/1hwABu-SL6u2cdhVkjOCUU6T0UyALujep1NfwYfQzWSo/edit#

data-man:
I don't know Python but this can be useful (with `instructions.xml` from uops.info): https://github.com/andreas-abel/XED-to-XML/blob/master/xmlToAssembler.py

PR #2875, opened 2026-01-29
#554 add more x86 instructions

gantsevdenis:
refers to #554  (labeled as "Good first issue")

I was thinking: the fact that C3 uses Oracle/Solaris syntax (the "{bwlq}" suffixes at the end of *some* instructions) makes it hard to mechanically add support for new ones, because the suffixes may or may not be valid for a given instruction... I am wondering, why not simply using "Intel/AMD mnemonics" as they are referred to in [ASM Reference](https://docs.oracle.com/cd/E53394_01/pdf/E54851.pdf ) ? Then we could simply process the _instructions.xml_ linked in the issue?

### DIFF
--- a/src/compiler/asm_target.c
+++ b/src/compiler/asm_target.c
@@ -626,13 +626,45 @@ static void init_asm_x86(PlatformTarget* target)
 	reg_instr(target, "vaddps", "w:v128/v256/v512, v128/v256/v512, v128/v256/v512/mem");
 	reg_instr(target, "vaddsd", "w:v128, v128, v128/mem");
 	reg_instr(target, "vaddss", "w:v128, v128, v128/mem");
-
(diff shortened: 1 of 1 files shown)


lerno:
Mostly because of the case of the integers where the desired size of the op would need to be inferred from the register, which would need additional verification code.

lerno:
It would be great if it could be generated, so maybe there's a way around it?

lerno:
Should I accept this for now and we look forward for some other way to add the rest?

gantsevdenis:
> Should I accept this for now and we look forward for some other way to add the rest?

No; I am still analyzing the .xml to generate as much as possible automatically; I think this can be closed

lerno:
I'll add these for now.

PR #2878, opened 2026-01-30
Fix formatting of path assignment in msvc_build_libraries.py

SmiteWindows:
There was an error with the semicolon, which has been corrected to the right one.


lerno:
Thank you

PR #2879, opened 2026-01-30
Optional byte limit for `io::readline_to_stream` and add `io::readline_limit`

NotsoanoNimus:
Allows a byte limit to be provided to readline in both cases. For `readline_limit` a Pair is used to bubble up whether the limit was reached while still keeping the in-flight read value intact.

### DIFF
--- a/releasenotes.md
+++ b/releasenotes.md
@@ -9,12 +9,12 @@
 - Integrated download of the MSVC SDK when compiling for Windows.
 
 ### Stdlib changes
-- Add optional line-length limitations to `io::readline` and `io::readline_to_stream`. #2879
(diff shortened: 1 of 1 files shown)


lerno:
Like I discussed on discord, I think that a better API would be an out parameter actually, not Pair

NotsoanoNimus:
> Like I discussed on discord, I think that a better API would be an out parameter actually, not Pair

I ended up scrapping the separate `readline_limit`. The caller can either use the return `usz?` or `String?.len` value to determine whether the limit was hit, instead of defining and providing an out-param for a one-line check.

This PR now just adds the optional `limit` param to the calls.

PR #2880, opened 2026-01-30
Improve linker error message

lyranie:
Add error message if directory with output file name already exists

### DIFF
--- a/src/compiler/linker.c
+++ b/src/compiler/linker.c
@@ -8,8 +8,6 @@
 #include <glob.h>
 #endif
 
-#include <sys/stat.h>
-
 const char *quote_arg = "\"";
 const char *concat_arg = ":";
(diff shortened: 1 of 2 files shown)


src/compiler/linker.c
```diff
@@ -1261,3 +1274,22 @@ bool linker(const char *output_file, const char **files, unsigned file_count)
 		  .Case("elf64_sparc", {ELF64BEKind, EM_SPARCV9})
 		  .Case("msp430elf", {ELF32LEKind, EM_MSP430})
  */
+
+bool output_file_exists(const char *output_file)
```
lerno: These functions already exit in `file_utils.c`, use them instead.

lyranie: ah didn't look to see if there already is one

lerno:
Thank you!

PR #2884, opened 2026-01-31
fix: add non-API Android NDK lib search path

lualvsil:
Android NDK places some libraries (such as libc++_shared.so)
directly under sysroot/usr/lib/<triple> instead of the
API-specific directory.

Previously, only the API-specific directory was added to the
linker search paths, which caused libc++_shared to not be found.

This change adds the non-API sysroot library directory to
match the actual NDK layout.


lerno:
Could this be harmonized with the #2664 PR?

lualvsil:
> Could this be harmonized with the #2664 PR?

Hi, thanks for the review.  

I’m using #2664 as a personal fork for Termux, but my patch specifically adds the non-API sysroot lib path to the Android linker search paths.  

It fixes the issue with libc++_shared.so not being found and should be compatible with #2664.  

I can adjust it if needed to harmonize with that PR.

lerno:
@Book-reader could you test this?

Book-reader:
I'm unable to compile a test program using this PR in its current form because nixpkgs LLVM is built without zstd support, and `libc.a` -- which is also stored directly under `sysroot/usr/lib/` -- has several compressed sections that causes lld to error out because it now tries to link with the static `libc.a` instead of the api-version-specific `libc.so`:
image

however when I put the changes from this PR after [src/compiler/linker.c#L551-L560](https://github.com/c3lang/c3c/blob/120f56ac5fe3bf9a2c18336431c963564ad80759/src/compiler/linker.c#L551-L560) instead of before it seems to prioritise the dynamic `libc.so` library and I'm able to successfully link with a symbol exported by `libc++_shared.so` when I add `-l c++`:
image
without this PR I'm unable to link with any symbols exported by `libc++_shared.so`:
image

lualvsil:
Updated the PR to apply this after the API-specific libc path handling. This should preserve dynamic libc.so priority.

lerno:
Can you verify this @Book-reader ?

Book-reader:
Yes, it works fine for me now

lerno:
Thank you!

PR #2891, opened 2026-02-02
codepage: add single-byte code page support

konimarti:
Add std::encoding::codepage with a shared engine for converting between single-byte code pages and UTF-8 using table-driven mappings.

Introduce generated tables and wrappers for several code pages[1] each exposing encode/decode helpers built on a common CodePageTable structure.

The mapping data is generated by cpgen[2] from the Unicode Consortium’s published code page mapping files and follows the Unicode standard’s interpretation of control characters (abstract characters) rather than historical VGA glyph shapes.

[1] Code page overview/groups:

    DOS/OEM code pages (legacy PC):
    cp437 cp737 cp775 cp850 cp852 cp855 cp857 cp860 cp861 cp862 cp863
    cp864 cp865 cp866 cp869 cp874

    Windows code pages (ANSI/Windows):
    cp1250 cp1251 cp1252 cp1253 cp1254 cp1255 cp1256 cp1257 cp1258

    ISO/IEC 8859 series (Latin/Regional):
    iso_8859_1 iso_8859_2 iso_8859_3 iso_8859_4 iso_8859_5 iso_8859_6
    iso_8859_7 iso_8859_8 iso_8859_9 iso_8859_10 iso_8859_11 iso_8859_13
    iso_8859_14 iso_8859_15 iso_8859_16

[2] github.com/konimarti/cpgen

### DIFF
--- a/lib/std/encoding/codepage.c3
+++ b/lib/std/encoding/codepage.c3
@@ -66,63 +66,74 @@ struct CodePageTable
     char[1024] from_codepoint;
 }
 
-<* Returns a CodePageTable for the given charset name. *>
-fn CodePageTable*? charset(String charset_name) => @pool()
+enum CodePage : (String name, CodePageTable* table)
+{
+	CP1250      = { "cp1250",      &codepage::CP1250 },
+	CP1251      = { "cp1251",      &codepage::CP1251 },
+	CP1252      = { "cp1252",      &codepage::CP1252 },
+	CP1253      = { "cp1253",      &codepage::CP1253 },
+	CP1254      = { "cp1254",      &codepage::CP1254 },
+	CP1255      = { "cp1255",      &codepage::CP1255 },
+	CP1256      = { "cp1256",      &codepage::CP1256 },
+	CP1257      = { "cp1257",      &codepage::CP1257 },
+	CP1258      = { "cp1258",      &codepage::CP1258 },
+	CP437       = { "cp437",       &codepage::CP437 },
+	CP737       = { "cp737",       &codepage::CP737 },
+	CP775       = { "cp775",       &codepage::CP775 },
+	CP850       = { "cp850",       &codepage::CP850 },
+	CP852       = { "cp852",       &codepage::CP852 },
+	CP855       = { "cp855",       &codepage::CP855 },
+	CP857       = { "cp857",       &codepage::CP857 },
+	CP860       = { "cp860",       &codepage::CP860 },
+	CP861       = { "cp861",       &codepage::CP861 },
+	CP862       = { "cp862",       &codepage::CP862 },
+	CP863       = { "cp863",       &codepage::CP863 },
+	CP864       = { "cp864",       &codepage::CP864 },
+	CP865       = { "cp865",       &codepage::CP865 },
+	CP866       = { "cp866",       &codepage::CP866 },
+	CP869       = { "cp869",       &codepage::CP869 },
+	CP874       = { "cp874",       &codepage::CP874 },
+	ISO_8859_1  = { "iso-8859-1",  &codepage::ISO_8859_1 },
+	ISO_8859_10 = { "iso-8859-10", &codepage::ISO_8859_10 },
+	ISO_8859_11 = { "iso-8859-11", &codepage::ISO_8859_11 },
+	ISO_8859_13 = { "iso-8859-13", &codepage::ISO_8859_13 },
+	ISO_8859_14 = { "iso-8859-14", &codepage::ISO_8859_14 },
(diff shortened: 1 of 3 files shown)


lerno:
Isn't it possible to encode each of these to big base64 blobs?

lerno:
Like, imagine packing those into base64 using `b64"...."` and indexing into that.

lerno:
Since it's a  `char[4] bytes; char len;` we can think of each conversion entry as a `char[5]` we can then pack all of those into memory, after which we convert it into a binary blob which is 5 * 256 long.

konimarti:
Using a b64-encoded blob should work well for the byte->utf8 mapping. I'll work on that. I'm not sure about the uint[] array. That would probably require some runtime decoding/overhead but I need to look into this a bit more.

konimarti:
@lerno The base64-encoded blob would look like this:
```c3
        .to_codepoint = b64`AQAAAAEBAAABAgAAAQMAAAEEAAABBQAAAQYAAAEHAAABCAAAAQkA
        AAEKAAABCwAAAQwAAAENAAABDgAAAQ8AAAEQAAABEQAAARIAAAETAAABFAAAARUAAAEWAAAB
        FwAAARgAAAEZAAABGgAAARsAAAEcAAABHQAAAR4AAAEfAAABIAAAASEAAAEiAAABIwAAASQA
        AAElAAABJgAAAScAAAEoAAABKQAAASoAAAErAAABLAAAAS0AAAEuAAABLwAAATAAAAExAAAB
        MgAAATMAAAE0AAABNQAAATYAAAE3AAABOAAAATkAAAE6AAABOwAAATwAAAE9AAABPgAAAT8A
        AAFAAAABQQAAAUIAAAFDAAABRAAAAUUAAAFGAAABRwAAAUgAAAFJAAABSgAAAUsAAAFMAAAB
        TQAAAU4AAAFPAAABUAAAAVEAAAFSAAABUwAAAVQAAAFVAAABVgAAAVcAAAFYAAABWQAAAVoA
        AAFbAAABXAAAAV0AAAFeAAABXwAAAWAAAAFhAAABYgAAAWMAAAFkAAABZQAAAWYAAAFnAAAB
        aAAAAWkAAAFqAAABawAAAWwAAAFtAAABbgAAAW8AAAFwAAABcQAAAXIAAAFzAAABdAAAAXUA
        AAF2AAABdwAAAXgAAAF5AAABegAAAXsAAAF8AAABfQAAAX4AAAF/AAACw4cAAsO8AALDqQAC
        w6IAAsOkAALDoAACw6UAAsOnAALDqgACw6sAAsOoAALDrwACw64AAsOsAALDhAACw4UAAsOJ
        AALDpgACw4YAAsO0AALDtgACw7IAAsO7AALDuQACw78AAsOWAALDnAACwqIAAsKjAALCpQAD
        4oKnAsaSAALDoQACw60AAsOzAALDugACw7EAAsORAALCqgACwroAAsK/AAPijJACwqwAAsK9
        AALCvAACwqEAAsKrAALCuwAD4paRA+KWkgPilpMD4pSCA+KUpAPilaED4pWiA+KVlgPilZUD
        4pWjA+KVkQPilZcD4pWdA+KVnAPilZsD4pSQA+KUlAPilLQD4pSsA+KUnAPilIAD4pS8A+KV
        ngPilZ8D4pWaA+KVlAPilakD4pWmA+KVoAPilZAD4pWsA+KVpwPilagD4pWkA+KVpQPilZkD
        4pWYA+KVkgPilZMD4pWrA+KVqgPilJgD4pSMA+KWiAPiloQD4paMA+KWkAPiloACzrEAAsOf
        AALOkwACz4AAAs6jAALPgwACwrUAAs+EAALOpgACzpgAAs6pAALOtAAD4oieAs+GAALOtQAD
        4oipA+KJoQLCsQAD4omlA+KJpAPijKAD4oyhAsO3AAPiiYgCwrAAA+KImQLCtwAD4oiaA+KB
        vwLCsgAD4pagAsKgAA==`,
```
We can reduce the LOC at the expense of the data readbility. But I think that's a fair tradeoff.

I haven't figured out a better representation for the uint[256] array, though. As soon as we start encoding the uints we would have to deal with endianness / an additional decoding step. My proposal would be to keep it as-is. This would also be in-line with e.g. Go's stdlib (they also use an uint[256] array). What do you think?

lerno:
You could just load it and convert if it's big endian. I don't think it's much of a problem.

lerno:
I'd like to have the data in a codepage_private.c3 and then the API itself in codepage.c3, so that it's easy to read.

lerno:
Thank you!

PR #2892, opened 2026-02-02
[stdlib] Add array `even`, `odd`, and `unzip` macros

NotsoanoNimus:
Adds some extra zip-related functionality (with associated tests) to the `array` core module for getting piecewise copies of an input list or array.

The input `[0, 1, 2, 3, 4]` becomes...
```
@unzip --> [0, 2, 4] and [1, 3]
even   --> [0, 2, 4]
odd    --> [1, 3]
```

This even/odd relationship applies to _indices_, not to the values themselves. Certainly open to considering different names here.

### DIFF
--- a/lib/std/core/array.c3
+++ b/lib/std/core/array.c3
@@ -330,6 +330,72 @@ macro bool @all(array, #predicate)
 
 
 <*
+ Extract a copy of all even-index elements from an input array.
+
+ @param [&inout] allocator : "The allocator used to create the return array."
+ @param [in] array : "The array from which to extract all even elements."
+
+ @require @is_valid_list(array) : "Expected a valid list"
+ @require $defined(array[:lengthof(array)]) : "Expected a sliceable list"
+*>
+macro even(Allocator allocator, array)
+{
+	return unlace_impl{$typeof(array[0])}(allocator, array[:lengthof(array)]);
+}
+
+
+<*
+ Extract a copy of all odd-index elements from an input array.
+
+ @param [&inout] allocator : "The allocator used to create the return array."
+ @param [in] array : "The array from which to extract all odd elements."
+
+ @require @is_valid_list(array) : "Expected a valid list"
+ @require $defined(array[:lengthof(array)]) : "Expected a sliceable list"
+*>
+macro odd(Allocator allocator, array)
+{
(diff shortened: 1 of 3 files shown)


NotsoanoNimus:
Regarding naming, `@unzip` does _not_ perform the inverse of `@zip`, so it should not be named as such. I was thinking `deinterlace`, `unlace`, or something similar instead. Unfortunately, C3 slices could never support a python-style `step` option like `my_slice[0:6:2]`, which would render these functions entirely useless.

e: It seems `unlace` would be best IMO. I chose to use `out` params instead of a `Pair` to avoid making tuple returns any kind of convention in this module, but would gladly use them instead. Additionally, it's worth considering a `range_from` function too, which can do the aforementioned extraction of stepped elements automatically; e.g., `array::range_from(mylist, start: 1, step: 2)`.

lerno:
We probably don't need the t-variants here. Let us remove those unless something is used *very* frequently.

NotsoanoNimus:
> We probably don't need the t-variants here. Let us remove those unless something is used _very_ frequently.

Done :+1:

lib/std/core/array.c3
```diff
@@ -329,6 +329,65 @@ macro bool @all(array, #predicate)
...
+<*
+ Extract a copy of all even-index elements from an input array.
+
+ @param [&inout] allocator : "The allocator used to create the return array."
+ @param [in] array : "The array from which to extract all even elements."
+
+ @require @is_valid_list(array) : "Expected a valid list"
+*>
+macro even(Allocator allocator, array)
+{
```
lerno, line 342: Change this into a macro that does the following:
1. If this is an array, turn it into a slice, otherwise if it's a slice keep the old value.
2. Take the resulting slice, and call a generic even function that expects a slice.

This means we will only get 1 copy of this function per element type. Otherwise we get one inlined copy every time this is called, which would have been useful only if this took an array and returned an array with a compile time size on the stack. Now it just takes an array or slice and returns a slice, and so doing it generic per type (e.g. int[3] would have a different implementation than int[2]) would be a waste.

Odd can be rewritten the same way.

NotsoanoNimus: I've completed the small rewrite, but per my brief Discord blurb, I've removed linked-list compatibility with these to make it all work without needing a ton of macro magic to create contiguous arrays.

lerno:
Thank you!

PR #2896, opened 2026-02-05
update error message when assigning value to enum with no associated values

### LINKED ISSUE #2895: error message when assigning value to enum still reccomends inline associated values instead of a const enum
```c3
enum Foo
{
	BAR = 1, //  Error: There are no associated values defined for this enum. Did you perhaps want C style gap enums? In that case, try enums with inline associated values.
}
```

Book-reader:
fixes #2895

### DIFF
--- a/releasenotes.md
+++ b/releasenotes.md
@@ -4,6 +4,7 @@
 
 ### Changes / improvements
 - Method resolution and `$define` now works together well unless definitions are out of order for real.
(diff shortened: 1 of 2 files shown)


Book-reader:
Fixed in https://github.com/c3lang/c3c/commit/054fcd2deb7828184a168b718562a9fc30b29453

PR #2899, opened 2026-02-05
Simplify CI, do more with less.

ManuLinares:
- Rename Docker job to **"LLVM ${{ matrix.llvm_version }} Compatibility"** for better visibility.
- Limit `build_type` to **Debug** for MSYS2, OpenBSD, NetBSD, and Nix jobs to reduce CI time.
- Refactor **Alpine Linux** job: updated to 3.23, simplified dependency installation, and removed redundant LLVM matrix.
- Modernize **Docker** builds:
    - Added automatic architecture detection (x86_64/ARM) for CMake downloads.
    - Generalized LLVM repository logic to support multiple Ubuntu codenames (Focal/Jammy/Noble).
    - Fixed git permissions issues in containerized builds using `safe.directory`.
- Standardized Android build packaging and added conditional testing for Debug builds.

image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -77,7 +77,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        build_type: [Release, Debug]
+        build_type: [Debug]
     defaults:
       run:
         shell: msys2 {0}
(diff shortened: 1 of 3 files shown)


lerno:
I haven't looked deeply, but it seems fine. I like that it's more compact now!

PR #2901, opened 2026-02-06
fix unix builds without curl

gantsevdenis:
Hm I am not sure this is what is the plan, but this fixes master build
offending commit: https://github.com/c3lang/c3c/commit/0be291e0d72b32db21e26bc48f638fbbf8ca5083
Otherwise the build fails:
```
/home/den/sources/c3c/src/utils/fetch_msvc.c:191:23: error: implicit declaration of function ‘download_file’ [-Wimplicit-function-declaration]
  191 |     const char *err = download_file(url, "", dst);
```

### DIFF
--- a/lib/std/hash/murmur.c3
+++ b/lib/std/hash/murmur.c3
@@ -1,0 +1,243 @@
+module std::hash::murmur3;
+
+<*
+ @param [in] data : "The data to hash"
+ @param seed : "The seed to use for hashing"
+ @require (data.len / 4) <= int.max : "Too much data"
+*>
+fn uint hash32(char[] data, uint seed)
+{
+	int nblocks = (int)data.len / 4;
+	uint h1 = seed;
+
+	const uint C1 = 0xcc9e2d51;
+	const uint C2 = 0x1b873593;
+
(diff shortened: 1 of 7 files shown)


lerno:
@ManuLinares can you have a look at this?

ManuLinares:
I've done another PR to solve this https://github.com/c3lang/c3c/pull/2904
Thanks for bringing attention to this.

A simple `#if FETCH_AVAILABLE` and a message should do the trick cleanly.
```
#else
void fetch_msvc(BuildOptions *options)
{
	error_exit("Failed to find Windows SDK.\n"
			   "Windows applications cannot be cross-compiled without it.\n"
			   "This build of c3c lacks cURL support and cannot download the SDK automatically.\n"
			   "Please provide the SDK path manually using --winsdk.");
}
#endif
```

lerno:
Thank you for taking to finding this error and making this PR @gantsevdenis. I'll go with Manu's smaller PR for this, but it's great that you found the issue.

gantsevdenis:
no pb, the other PR makes more sense

PR #2910, opened 2026-02-08
Enable default NDK detection for Android x86_64

lualvsil:
The compiler auto-detects the Android NDK for android-aarch64.
This change extends the same logic to android-x86_64.


lualvsil:
Closing for now — it looks like this code path also runs for non-Android builds and affects entrypoint/IR generation in CI. I’ll rework this to ensure Android-specific defaults are only applied when Android is the active target.

lerno:
Thank you

PR #2911, opened 2026-02-08
fix/error-printing: pass `decl->name` instead of `decl->span`

alsakandari:
The format specifies that the argument passed to it should be a string, decl->span is a struct


alsakandari:
CI is finding more bugs it seems, things unreachable got reached, not the concern of my PR though

lerno:
I think this was fixed separately, can you verify?

alsakandari:
Seems like you fixed it in master, I can close this now

PR #2915, opened 2026-02-09
Implement lazy-loaded libcurl

ManuLinares:
- Uses `dlopen` to load libcurl at runtime for better portability.
- Replaces the FETCH_AVAILABLE macro with download_available() check.
- Provides descriptive error messages when libcurl is missing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -129,9 +129,6 @@ if(C3_USE_MIMALLOC)
     )
     FetchContent_MakeAvailable(mimalloc)
 endif()
-if (NOT WIN32)
-    find_package(CURL)
-endif()
 
 find_package(Git QUIET)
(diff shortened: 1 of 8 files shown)


lerno:
Is there something in the CI scripts that tests it with vendor fetch or something? I know vendor fetch is tested for Windows, but what about the others?

ManuLinares:
> Is there something in the CI scripts that tests it with vendor fetch or something? I know vendor fetch is tested for Windows, but what about the others?

Yes, they all tested fine.
https://github.com/c3lang/c3c/blob/master/scripts/tools/ci_tests.sh#L117

The only one that is purposedly skipped is [nix](https://github.com/c3lang/c3c/blob/master/nix/default.nix#L90), because nix ci build doesn't have network.

lerno:
Awesome!

Issue #2914, opened 2026-02-09
[Bug] Can't produce dynamic libraries in fedora 43.


Borwe:
I have c3 installed from release builds version 0.7.9, and also tried compiling master to get c3, but all seem to not be able to produce dynamic libraries for linux on fedora 43, a code sample:

```c3
//test.c3
module builder;

fn void build(){}
```

running `c3c dynamic-lib ./test.c3` prints out the following error

<img width="905" height="225" alt="Image" src="https://github.com/user-attachments/assets/10924390-407a-40a8-92d5-27fdf0379689" />

When compiled with -vvv flag the following log is produced

[log.txt](https://github.com/user-attachments/files/25188988/log.txt)

There was a [discussion](https://discord.com/channels/650345951868747808/927010276128202762/1470457077032947967) on discord that it could be something to do with search paths for `/usr/lib/crt1.o` and other files but they were located at /usr/lib64/*` on fedora 43 by default, if this info might help coming up with a solution.

PR #2917, opened 2026-02-10
fix fedora dyn lib 

gantsevdenis:
fixes https://github.com/c3lang/c3c/issues/2914


ManuLinares:
Could you update the error `No crt in /usr/lib/*/` to reflect the change?

Maybe:
```
No crt in /usr/{lib,lib64}/*/
```

lerno:
Great! Thank you.

PR #2918, opened 2026-02-10
Minor attribute combination checking

ThatOneDamien:
Fixes an extremely minor typo in the attribute analysis function. The line that was meant to check for a bad combination of `@nodiscard` and `@noreturn` was not functioning properly. I fixed that typo, and added a few more checks for bad combinations of attributes like `@used` and `@unused`, or `@noinline` and `@inline`. 

This is an extremely minor change that I made simply because I found the typo when going through the code base.


src/compiler/sema_decls.c
```diff
@@ -3182,7 +3182,6 @@ static bool sema_analyse_attribute(SemaContext *context, ResolvedAttrData *attr_
 		RETURN_SEMA_ERROR(attr, "'%s' is not a valid %s attribute.", attr->name, attribute_domain_to_string(domain));
 	}
 
-	// No attribute has more than one argument right now.
```
ManuLinares: why remove this comment?

ManuLinares:
I don't think this patch works:

```c3
import std::io;

int global_counter @used @unused;

fn int foo() @inline @noinline
{
    return 1;
}

/*fn int fail() @nodiscard @noreturn
{
    abort();
}*/ // this one works:  Error: @noreturn cannot be combined with @nodiscard.

fn void main()
{
	int a = foo();
	global_counter++;
}
```

I get no errors when `@used @unused;` or when `@inline @noinline`

lerno:
You can add some tests in /test/test-suite. For example look at the tests here: https://github.com/c3lang/c3c/blob/master/test/test_suite/attributes/ and add a test file for these situations.

ThatOneDamien:
@ManuLinares @lerno Sorry for the delayed response, had some things going on. For me the patch is working, when running `c3c compile test.c3` with test.c3 containing the snippet you provided, I get the following output:
```
1: import std::io;
2:
3: int global_counter @used @unused;
                            ^^^^^^^
(/home/Damien/Dev/c3cme/build/test.c3:3:26) Error: @unused cannot be combined with @used.

 2:
 3: int global_counter @used @unused;
 4:
 5: fn int foo() @inline @noinline
                         ^^^^^^^^^
(/home/Damien/Dev/c3cme/build/test.c3:5:22) Error: @noinline cannot be combined with @inline.
```
I will try recloning my repository in a bit to see if somehow the force-pushes I did caused an issue (I only did them because I had no intellisense/LSP and I made a typo multiple times).

As for the tests, I will add those shortly.

lerno:
Thank you!

PR #2925, opened 2026-02-11
Fix array initializer analysis

### LINKED ISSUE #2850: Non-descriptive array element error message
So I had this code,
```c
const char[128] SCANCODE_ASCII_MAP =
{
	0, 27, '1','2','3','4','5','6','7','8','9','0','-','=', 8,
	'\t','q','w','e','r','t','y','u','i','o','p','[',']','\n',
	0, 'a','s','d','f','g','h','j','k','l',';','\'','`',
	0, '\\','z','x','c','v','b','n','m',',','.','/',
	0, '*', 0, ' ',
};
```

And I expected c2c to fill out the rest with zeros hopefully, but nah. So I had gpt fill in the rest of the zeros, but it didn't get them all. So I had to play higher/lower with the compiler until I actually got the amount of 0's I needed.
```c
const char[128] SCANCODE_ASCII_MAP =
{
	0, 27, '1','2','3','4','5','6','7','8','9','0','-','=', 8,
	'\t','q','w','e','r','t','y','u','i','o','p','[',']','\n',
	0, 'a','s','d','f','g','h','j','k','l',';','\'','`',
	0, '\\','z','x','c','v','b','n','m',',','.','/',
	0, '*', 0, ' ',

	// pad to 128
	0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
	0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
	0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
	0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
	0,0,0,0,0,0,
};
```

Current error messages:
```
c3c build kernel-debug   --obj-out build/c3 --output-dir build
 35:    0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
 36:    0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
 37:    0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
 38:    0,0,0,0,0,0,0,
                    ^
(/home/me/src/osdev3/kernel/ps2.c3:38:14) Error: Too many elements in initializer, expected only 128.
```
```
c3c build kernel-debug   --obj-out build/c3 --output-dir build
 35:    0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
 36:    0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
 37:    0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
 38:    0,0,0,0,
              ^
(/home/me/src/osdev3/kernel/ps2.c3:38:8) Error: Too few elements in initializer, 128 elements are needed.
```

I think it would be cool if the error message could give the amount of current elements as well, to, you know, help out a bit.

sam-paiva:
This PR fixes the handling of array initializers in the semantic analyzer. Previously, certain cases with arrays having fewer or more elements than expected, and would produce “Too few/too many elements in initializer” errors.

This should fix #2850

### DIFF
--- a/src/compiler/sema_initializers.c
+++ b/src/compiler/sema_initializers.c
@@ -350,7 +350,7 @@ static inline bool sema_expr_analyse_array_plain_initializer(SemaContext *contex
 	}
 	if (expected_members > 0 && count > 0 && count != expected_members)
 	{
-		RETURN_SEMA_ERROR(elements[0], "Too %s elements in initializer, expected %u. Current count: %d.", count > expected_members ? "many" : "few", expected_members, count);
+		RETURN_SEMA_ERROR(elements[0], "Too %s (%u) elements in initializer, expected %u.", count > expected_members ? "many" : "few", count, expected_members);
 	}
 
 	bool optional = false;
--- a/test/test_suite/arrays/array_initializer_size_err.c3
+++ b/test/test_suite/arrays/array_initializer_size_err.c3
@@ -1,4 +1,4 @@
 fn void main()
 {
-    const int[3] ARR = {1, 2}; // #error: Too few elements in initializer, expected 3. Current count: 2.
+    const int[3] ARR = {1, 2}; // #error: Too few (2) elements in initializer, expected 3
 }
--- a/test/test_suite/arrays/array_initializer_size_err2.c3
+++ b/test/test_suite/arrays/array_initializer_size_err2.c3
@@ -1,4 +1,4 @@
 fn void main()
 {
-    const int[3] ARR = {1, 2, 3, 4}; // #error: Too many elements in initializer, expected 3. Current count: 4.
+    const int[3] ARR = {1, 2, 3, 4}; // #error:  Too many (4) elements in initializer, expected 3
 }
--- a/test/test_suite/arrays/array_of_inferred_array.c3
+++ b/test/test_suite/arrays/array_of_inferred_array.c3
@@ -4,5 +4,5 @@ macro void test(int[2][6] a)
 }
 fn void main()
 {
-	test({ { 1, 2 } }); // #error: Too few elements in initializer
+	test({ { 1, 2 } }); // #error:  Too few (1) elements in initializer, expected 6
 }


lerno:
Can you also add some test cases in this folder: https://github.com/c3lang/c3c/tree/master/test/test_suite/initializer_lists so that it's clear it works?

sam-paiva:
I included the tests

lerno:
I think you just changed C3 semantics to match C semantic here?

sam-paiva:
That was the intention, and from what I understood from the open issue. If I misunderstood, I would appreciate feedback on that.

lerno:
It should just give how many there were / what's needed to fill out. C semantics where the rest is implicitly zeroed out is... problematic from a correctness perspective.

Something that could be done – but is more work, is to allow designated initializer after regular initializers, e.g.
`int[10] a = { 1, 2, 3, [3..9] = 0 }`. This is quite a bit more involved though, with traps for the unwary. So for now just the first part you did, adding the count, is enough.

lerno:
But if you do want to work on "designated initializers after regular initializers" let me know. The desired semantics is that
`int[10] a = { 1, 2, 3, [3..9] = 0 }` essentially is transformed into `int[10] a = { [0] = 1, [1] = 2, [2] = 3, [3..9] = 0 }` so that it can then use normal analysis for designated initializers. Note that this also changes handling of struct init to some degree.

This would be valid even though it overwrites, since overwriting is valid for designated initializers: `int[3] a = { 1, 2, 3, [0..2] = 0 }`

sam-paiva:
That makes sense. I’ll revert my changes regarding the auto-initializer, and ensure the count is included in the error messages.

Regarding the designated initializers, I’m not sure I’ll have time to implement that in the near term, so it may be better to address it in a separate pull request, so we don't leave this PR open for too long.

My suggestion is that we keep this PR focused on adding the count to the error messages so we can close out the current issue. Does that sound reasonable?

lerno:
Sounds good

sam-paiva:
I think I covered the scenarios we discussed.

lerno:
Thank you

PR #2930, opened 2026-02-12
stdlib: `std::compression::zip` and `std::compression::deflate`

ManuLinares:
- C3 implementation of DEFLATE (RFC 1951) and ZIP archive handling.
- Support for reading and writing archives using STORE and DEFLATE methods.
- Supports both fixed and dynamic Huffman blocks
- Compression using greedy LZ77 matching.
- Stream-based entry reading and writing.
- ZIP64 support for large files/archives.
- Unit tests and benchmarks.

```
$ gcc -O3 -c dependencies/miniz/miniz.c -o build/miniz.o
$ build/c3c -O3 compile-run test/compression/deflate_benchmark.c3 build/miniz.o


 DEFLATE BENCHMARK  Comparing C3 std::compression::deflate with miniz (in-process)

Test Case                  | C3 Rat. | Miz Rat. | C3 MB/s | Miz MB/s | Winner
---------------------------+---------+---------+---------+---------+-----------
Redundant (100MB 'A')      |   0.10% |   0.10% |  1342.2 |   911.9 | C3 (1.5x)
Compiler Source (Bulk)     |  22.06% |  22.15% |   121.3 |   132.4 | Miniz (1.1x)
Stdlib Source (Bulk)       |  24.87% |  24.61% |   118.1 |   122.5 | Miniz (1.0x)
Log Files (Simulated)      |  11.10% |  11.13% |   343.5 |   301.2 | C3 (1.1x)
Web Content (Simulated)    |   1.51% |   1.51% |   751.7 |   779.9 | Miniz (1.0x)
CSV Data (Simulated)       |  15.05% |  14.70% |   244.2 |   254.3 | Miniz (1.0x)
Binary Data (Structured)   |  28.77% |  28.85% |   102.2 |    96.5 | C3 (1.1x)
Random Noise (Scaled)      |  65.82% |  64.57% |    79.8 |    65.8 | C3 (1.2x)
Tiny File (asd.c3)         | 140.00% | 116.67% |     3.2 |    12.9 | Miniz (4.1x)
Natural Text (Scaled)      |   0.29% |   0.29% |  1358.9 |   879.4 | C3 (1.5x)

OVERALL SUMMARY
  Average Throughput C3:       446.5 MB/s
  Average Throughput Miniz:    355.7 MB/s
  C3 is 1.3x faster on average!
```

### DIFF
--- a/test/unit/stdlib/compression/deflate.c3
+++ b/test/unit/stdlib/compression/deflate.c3
@@ -60,10 +60,10 @@ fn void test_deflate_large_repetitive() @if($feature(SLOW_TESTS))
 
 	mem::set(original.ptr, (char)'A', len);
 
-	char[] compressed = deflate::compress(original)!!;
+	char[] compressed = deflate::compress(mem, original)!!;
 	defer free(compressed.ptr);
 
-	char[] decompressed = deflate::decompress(compressed)!!;
+	char[] decompressed = deflate::decompress(mem, compressed)!!;
 	defer free(decompressed.ptr);
 
 	assert(decompressed.len == len, "Length mismatch");


lib/std/compression/zip.c3
```diff
@@ -0,0 +1,613 @@
...
+		defer (catch err) (void)f.close();
+
+		f.seek(0, Seek.END)!;
+		usz file_size = (usz)f.seek(0, Seek.CURSOR)!;
+
+		if (file_size < ZipEOCD.sizeof) return CORRUPTED_DATA~;
+
+		f.seek((isz)(file_size - ZipEOCD.sizeof), Seek.SET)!;
+		ZipEOCD eocd;
+		if (f.read(((char*)&eocd)[:ZipEOCD.sizeof])! != ZipEOCD.sizeof || eocd.signature != ZIP_EOCD_SIG)
```
konimarti: At a cursory glance, does this work when there's a variable-sized comment after the EOCD header?

lib/std/compression/zip.c3
```diff
@@ -0,0 +1,613 @@
...
+		archive.file.seek((isz)eocd.cd_offset, Seek.SET)!;
+
+		for (ushort i = 0; i < eocd.num_entries; i++)
+		{
+			ZipCDH cdh;
+			if (archive.file.read(((char*)&cdh)[:ZipCDH.sizeof])! != ZipCDH.sizeof) break;
+			if (cdh.signature != ZIP_CDH_SIG) break;
+
+			char[] name = allocator::malloc(allocator, cdh.filename_len)[:cdh.filename_len];
+			if (archive.file.read(name)! != (usz)cdh.filename_len)
```
konimarti: According to the [zip specs](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT) Appendix D, I believe the filename is encoded in code page 437 (for historical reasons) unless the language encoding flag is set or there's a unicode extra header. Luckily, the stdlib can handle cp437 since Monday :)

lib/std/compression/zip.c3
ManuLinares: no it does not

ManuLinares: now it does

lib/std/compression/deflate.c3
```diff
@@ -0,0 +1,1079 @@
...
+	char[1] b;
+	if (try n = self.read(b[..]))
+	{
+		if (n == 0) return io::EOF~;
+		return b[0];
+	}
+	return io::EOF~;
+}
+
+fn char[]? decompress(char[] input, Allocator allocator = mem)
```
Book-reader: The style in the stdlib is usually to have the allocator as the first parameter instead of using a default parameter. The refcount `new` function does use `Allocator allocator = mem`, but the `std::compression::qoi` functions and all the encoding functions have the allocator as the first parameter without a default value so it's probably better to go with that for consistency.

lib/std/compression/deflate.c3
```diff
@@ -0,0 +1,1079 @@
...
+{
+	if (input.len == 0) return allocator::new_array(allocator, char, 0);
+
+	io::ByteReader mem_stream;
```
Book-reader:
```suggestion
	ByteReader mem_stream;
```

Book-reader (commented):
It's great to see zip support being added to the stdlib! I don't have much knowledge of the zip spec so I can't comment on the implementation, but I do have this one small style nitpick

lib/std/compression/deflate.c3
ManuLinares: You're absolutely right, thanks! I am aware and will correct the style appropriately.

lib/std/compression/zip.c3
```diff
@@ -0,0 +1,1118 @@
...
+	writer.close()!;
+}
+
+<*
+ Writes an entire file to the archive.
+ @param filename : `The name of the file to create.`
+ @param data : `The data to write.`
+ @param method : `Compression method.`
+*>
+fn void? ZipArchive.write_file(&self, String filename, char[] data, ushort method = ZIP_METHOD_DEFLATE)
```
Book-reader: the `method` parameter here isn't used, is it intended to be used later or left over from an older version of the code?

ManuLinares: Nice catch, this is helpful; leftover from when deflate wasn't fully implemented but it should be passed to
```c3
	ZipEntryWriter? writer_opt = self.open_writer(filename, method);
```

lib/std/compression/zip.c3
```diff
@@ -0,0 +1,1118 @@
...
+				};
+
+				// Header(4) + Uncomp(8) + Comp(8) + Offset(8)
+				ushort extra_size = 28;
+				extra_data = allocator::malloc(self.allocator, extra_size)[:extra_size];
+				extra_data[0] = ZIP64_EXTRA_ID;
+				extra_data[2] = (char)(extra_size - 4);
+				extra_data[4] = (char)entry.uncompressed_size;
+				extra_data[12] = (char)entry.compressed_size;
+				extra_data[20] = (char)entry.offset;
```
Book-reader:
```suggestion
				bitorder::write(ZIP64_EXTRA_ID, extra_data[:2], UShortLE);
				bitorder::write((ushort)(extra_size - 4), extra_data[2:2], UShortLE);
				bitorder::write(entry.uncompressed_size, extra_data[4:8], ULongLE);
				bitorder::write(entry.compressed_size, extra_data[12:8], ULongLE);
				bitorder::write(entry.offset, extra_data[20:8], ULongLE);
```

Casting the values to char and assigning them to individual indicies of the array will truncate them to one byte, resulting in the compressed/uncompressed size and offset becoming corrupted for any files larger than `0xFFFFFFFE`.

Even with this change the resulting zip file will be corrupted, so something is still wrong:
```c3
import std;

fn void main()
{
	ZipArchive data = zip::open("./test.zip", "w")!!;
	ZipEntryWriter writer = data.open_writer("large_file", zip::ZIP_METHOD_DEFLATE)!!;
	char[] empty = mem::new_array(char, 0xFFFF);
	defer free(empty);
	$assert 0xFFFFFFFF % 0xFFFF == 0;
    // write exactly 0xFFFFFFFF bytes to large_file in 0xFFFF chunks
	for (usz i = 0; i < 0xFFFFFFFF; i += 0xFFFF)
	{
		io::write_all(&writer, empty)!!;
	}
	writer.close()!!;
	data.close()!!;
}
```
```console
$ c3c compile-run test.c3 --run-once -O2
$ unzip test.zip
Archive:  test.zip
error [test.zip]:  start of central directory not found;
  zipfile corrupt.
  (please check that you have transferred or created the zipfile in the
  appropriate BINARY mode and that you have compiled UnZip properly)
```

lib/std/compression/zip.c3
```diff
@@ -0,0 +1,1154 @@
...
+}
+
+<*
+ Opens a writer for a new entry.
+ @param filename : `The name of the file to create in the archive.`
+ @param method : `Compression method.`
+ @return `A writer for the new entry.`
+ @require method == ZIP_METHOD_STORE || method == ZIP_METHOD_DEFLATE
+*>
+fn ZipEntryWriter? ZipArchive.open_writer(&self, String filename, ushort method = ZIP_METHOD_STORE)
```
Book-reader: I'm not sure if there should be a default value for the method, or if there is I think it should be `ZIP_METHOD_DEFLATE`. I was testing this and wondering why my files weren't getting compressed because I assumed that writing to a zip file would compress it and hadn't noticed that the default value was `ZIP_METHOD_STORE`

ManuLinares: yes default will be `method = ZIP_METHOD_DEFLATE`

lib/std/compression/zip.c3
Book-reader: while this suggested change doesn't fix writing large files, it does correct the value of `extra_data`.

`extra_data` from the above example without the change (filled with random garbage because malloc doesn't zero and only a few indicies got assigned to):
```
[1, 128, 24, 255, 255, 128, 255, 255, 253, 128, 255, 255, 14, 128, 255, 255, 255, 128, 255, 255, 0, 129, 255, 255, 1, 129, 255, 255]
```
`extra_data` from the above example with the change:
```
// id: 0x0001, header size: 24, uncompressed size: 0xFFFFFFFF, compressed size: 4161806, offset: 0
[1, 0, 24, 0, 255, 255, 255, 255, 0, 0, 0, 0, 14, 129, 63, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
```

lib/std/compression/zip.c3
```diff
@@ -0,0 +1,1118 @@
...
+			bool is_zip64 = entry.uncompressed_size >= 0xFFFFFFFF || entry.compressed_size >= 0xFFFFFFFF || entry.offset >= 0xFFFFFFFF;
+
+			char[] extra_data;
+			if (is_zip64)
+			{
+				cdh = {
+					.compressed_size = 0xFFFFFFFF,
+					.uncompressed_size = 0xFFFFFFFF,
+					.relative_offset = 0xFFFFFFFF
+				};
```
Book-reader:
```suggestion
				cdh.compressed_size = 0xFFFFFFFF;
				cdh.uncompressed_size = 0xFFFFFFFF;
				cdh.relative_offset = 0xFFFFFFFF;
```

doing `val = { .field = x}` will zero all the other fields

Book-reader (changes requested):
figured out the other problem, with this change and https://github.com/c3lang/c3c/pull/2930#discussion_r2802023719 writing files `0xFFFFFFFF` and larger works

lib/std/compression/zip.c3
ManuLinares: oof, i refactored that without thinking. it was originally correct

ManuLinares: Thanks, I found the code before refactoring some stuff. I was doing for example
```
*(ulong*)&extra_data[4] = entry.uncompressed_size;
```

and forgot about the zeroing of other fields in `cdh`. Again, this is really helpful.

lib/std/compression/deflate.c3
ManuLinares: On second thought, I'll keep the default at the end for better ergonomics. I am still open to discuss this of course.

edit: probably will end up at front like the rest of the stdlib but I'm done for today.

ManuLinares:
> Should we add a test case for writing a file with size above `0xFFFFFFFE`?

I added a specific test case for this scenario. If you would like to extend the testing further, that would be appreciated.

konimarti:
A few more observations:
* when you write the zip headers to a file with `self.file.write(((char*)&cdh)[:ZipCDH.sizeof])!;` is that safe on big-endian systems? The zip format expects every header in little-endian. You already take care of the endianness when writing the extra bytes.
* the `ArchiveStreamAdapter` seems to act like the io::LimitReader from the stdlib; maybe this one could be used instead?
* code attribution: this is probably personal style but I'd suggest to give credit where appropriate. the package-merge algorithm is most likely from https://create.stephan-brumme.com/length-limited-prefix-codes/ since it seems to be related to my c3 port. That was actually the trickiest part in the dynamic Huffman deflate implementation because it's not covered in the RFC.

ManuLinares:
> A few more observations:
> 
>     * when you write the zip headers to a file with `self.file.write(((char*)&cdh)[:ZipCDH.sizeof])!;` is that safe on big-endian systems? The zip format expects every header in little-endian. You already take care of the endianness when writing the extra bytes.
> 
>     * the `ArchiveStreamAdapter` seems to act like the io::LimitReader from the stdlib; maybe this one could be used instead?
> 
>     * code attribution: this is probably personal style but I'd suggest to give credit where appropriate. the package-merge algorithm is most likely from https://create.stephan-brumme.com/length-limited-prefix-codes/ since it seems to be related to my c3 port. That was actually the trickiest part in the dynamic Huffman deflate implementation because it's not covered in the RFC.

Hi @konimarti, excellent feedback!

- Endianness: I've updated the ZIP header structs (ZipCDH, ZipLFH, etc.) to use the little-endian types from `std::core::bitorder`.
- The adapter here is a bit more specialized than the stdlib version. It has to seek to the correct offset before every read because multiple readers share the same file handle.
- I've added the proper attribution to the pkg_merge function, giving credit to your C3 implementation and Stephan Brumme's original work, thanks.

*I also fixed a small bug in the STORE read method.*

ManuLinares:
@lerno I've pushed the *final* optimizations and fixes. All tests and benchmarks are passing. Last change would be:

quoting @Book-reader "The style in the stdlib is usually to have the allocator as the first parameter instead of using a default parameter", yes?

lib/std/compression/zip.c3
```diff
@@ -0,0 +1,1226 @@
...
+			defer allocator::free(allocator, comment_data.ptr);
+			if (archive.file.read(comment_data)! == (usz)eocd.comment_len.val)
+			{
+				archive.comment = codepage::decode(allocator, comment_data, CodePage.CP437)!!;
+			}
+		}
+
+		defer (catch err) (void)archive.close();
+
+		if (eocd.cd_offset.val > (uint)isz.max) return io::OVERFLOW~;
```
Book-reader: Is this correct? `isz.max` is significantly higher than the maximum value of `uint` so it'll get truncated to `uint.max`, and `cd_offset.val` can never be greater than `uint.max`

Book-reader: actually it could be higher on 32 bit platforms, is that what it's for?

Book-reader:
Would it be possible to compress the data written to a `ZipEntryWriter` as it's streamed in? Currently compressing large files consumes the size of the uncompressed data as it's being written, which can be a lot.
This might require making an `InStream` that does the same as `deflate::compress` though, so that could be a future thing after this PR unless it's very easy to implement. I don't this PR to be held back just for that.

lib/std/compression/zip.c3
ManuLinares: > actually it could be higher on 32 bit platforms, is that what it's for?

indeed, I havent't fully tested but this was part of future wasm32 work

ManuLinares:
> Would it be possible to compress the data written to a `ZipEntryWriter` as it's streamed in? Currently compressing large files consumes the size of the uncompressed data as it's being written, which can be a lot. This might require making an `InStream` that does the same as `deflate::compress` though, so that could be a future thing after this PR unless it's very easy to implement. I don't this PR to be held back just for that.

Yes, it's definitely possible, but it would require a streaming deflate compressor which the `std::compression::deflate` module currently lacks.

```c3
fn void? compress_stream(Allocator allocator = mem, InStream input, OutStream output)
```

This is indeed a more complex implementation than the current one and would be a good future improvement. I had some issues with implementing this last week but will definitively give it another go when I have the time.

lib/std/compression/deflate.c3
```diff
@@ -0,0 +1,1123 @@
...
+	inflater.init(&mem_stream, &tmp_bit_buf);
+
+	usz out_cap = input.len * 2;
+	if (out_cap < 1024) out_cap = 1024;
+
+	usz out_len = 0;
+	char* out = allocator::malloc(allocator, out_cap);
+	defer allocator::free(allocator, out);
+
+	// TODO: could this be a List {char} instead?
```
Book-reader: This was a note to myself that I forgot to remove before commiting in https://github.com/ManuLinares/c3c/pull/18, you can remove it

lib/std/io/os/file_libc.c3
```diff
@@ -123,3 +123,22 @@ macro fault file_seek_errno() @local
 	}
 }
 
+
+struct Utimbuf
+{
+	libc::Time_t actime;
+	libc::Time_t modtime;
```
Book-reader: the `libc::` prefix for `Time_t` isn't necessary, same for in `native_set_modified_time` in `file_libc.c3` and `file_nolibc.c3`, `SetModifiedTimeFn`, and `set_modified_time`

lib/std/compression/deflate.c3
```diff
@@ -0,0 +1,1123 @@
...
+				self.state = READ_STORED_LEN;
+			} else if (type == 1) {
+				self.lit_ptr = &lit_fixed_global;
+				self.dist_ptr = &dist_fixed_global;
+				self.state = DECODE_SYMBOL;
+			} else if (type == 2) {
+				self.state = READ_DYN_COUNTS;
+			} else {
+				return CORRUPTED_DATA~;
+			}
```
Book-reader: this could be rewritten as a `switch` using ranges which might make it slightly more readable, same at https://github.com/ManuLinares/c3c/blob/847c91f02b86cf459ba94e64e8d668c98366c911/lib/std/compression/deflate.c3#L1030-L1047 and https://github.com/ManuLinares/c3c/blob/847c91f02b86cf459ba94e64e8d668c98366c911/lib/std/compression/deflate.c3#L1064-L1074. You don't have to change it, I just now noticed that it could be done (tbh I feel like a bit of an asshole pointing out such minor things, so feel free to ignore this comment if you want).

lib/std/compression/deflate.c3
```diff
@@ -0,0 +1,1123 @@
...
+				self.write_byte((char)symbol);
+			} else if (symbol == 256) {
+				self.state = self.final ? DONE : START_BLOCK;
+			} else if (symbol <= 285) {
+				uint len_idx = symbol - 257;
+				self.match_len = LENGTH_BASE[len_idx] + self.reader.read_bits(LENGTH_EXTRA[len_idx])!;
+				self.state = READ_DIST_SYM;
+			} else {
+				return CORRUPTED_DATA~;
+			}
```
Book-reader: same here

Book-reader (approved):
Great! Most of the feedback I have now are these incredibly minor code style things I missed before.
I do think the allocator should be changed to be the first parameter in `deflate::compress` etc, the reason the stdlib style changed in 0.7 to be like that was so it was clear which functions would cause allocations that would need to be freed because they would need to be explicitly given an allocator.

lib/std/compression/deflate.c3
ManuLinares: I thought I did remove it 👌

lib/std/compression/deflate.c3
ManuLinares: More readable is good. On the contrary, I asked for this kind of feedback.

Thanks

konimarti:
There's an issue in the `deflate::decompress_stream` function: it will read beyond the end-of-block marker of the final deflate block. The inflate decompressor should only read up until the end of the deflate stream and not further. This issue did not pop up in the current tests because the zip format provide the lengths for both the compressed and uncompressed data. However, this is not the case in many other formats or applications of deflate.

The following tests highlights the issue and fails with an io::EOF error in the last read_byte call:
```c3
fn void test_deflate_embedded_stream() => @pool()
{
	String base = "This is a streaming test for DEFLATE. ";

	char[] compressed = deflate::compress(mem, base[..])!!;
	defer free(compressed.ptr);

	usz append_len = compressed.len + 1;
	char[] append = mem::malloc(append_len)[:append_len];
	defer free(append.ptr);

	append[:compressed.len] = compressed[..];
	append[^1] = 'c';

	ByteReader reader;
	reader.init(append);

	ByteWriter writer;
	writer.tinit();

	deflate::decompress_stream(&reader, &writer)!!;
	assert(writer.str_view() == base);

	assert(reader.read_byte()!! == 'c');
}
```

lerno:
Maybe something I introduced?

konimarti:
> Maybe something I introduced?

I dropped the last two commits and re-tested and it still fails. It seems not related to your changes. I think it's related to the refill logic of the bitreader.

lerno:
Yeah, I looked through it and this happens due to the internal buffering.

Issue #2014, opened 2025-03-03
riscv64 target llvm error


wizard-lgtm:
project.json
```jsonc
{
  // Language version of C3.
  "langrev": "1",
  // Warnings used for all targets.
  "warnings": [ "no-unused" ],
  // Directories where C3 library files may be found.
  "dependency-search-paths": [ "/root/c3/lib" ],
  // Libraries to use for all targets.
  "dependencies": [ ],
  // Authors, optionally with email.
  "authors": [ "wizard-lgtm" ],
  // Version using semantic versioning.
  "version": "0.1.0",
  // Sources compiled for all targets.
  "sources": [ "src/**" ],

  "output": "build",
  // Architecture and OS target.
  // You can use 'c3c --list-targets' to list all valid targets.
  // "target": "windows-x64",
  "target": "elf-riscv64", // riscv64
  // Targets.
  "targets": {
    "barebones_c3": {
      // Executable or library.
      "type": "executable",
      // Additional libraries, sources
      // and overrides of global settings here.
    },
  },
  // Global settings.
  // CPU name, used for optimizations in the LLVM backend.
  "cpu": "generic",
  // Optimization: "O0", "O1", "O2", "O3", "O4", "O5", "Os", "Oz".
  "opt": "O0"
  // See resources/examples/project_all_settings.json and 'c3c --list-project-properties' to see more properties.
}
```
```
root@DESKTOP-P9MNADN:~/barebones_c3# c3c build --no-entry
```
```c
WARNING! This architecture is unsupported.
] Assuming non-Posix environment
ld.lld: error: undefined symbol: __divdi3
>>> referenced by math.c3:1021 (/root/c3/lib/std/math/math.c3:1021)
>>>               build/obj/elf-riscv64/std.core.mem.o:(std.core.mem.aligned_offset)
>>> referenced by formatter_private.c3:35 (/root/c3/lib/std/io/formatter_private.c3:35)
>>>               build/obj/elf-riscv64/std.io.o:(std.io.Formatter.floatformat)
>>> referenced by formatter_private.c3:35 (/root/c3/lib/std/io/formatter_private.c3:35)
>>>               build/obj/elf-riscv64/std.io.o:(std.io.Formatter.floatformat)
>>> referenced 3 more times
>>> did you mean: __divti3
>>> defined in: build/obj/elf-riscv64/std.math.o

ld.lld: error: undefined symbol: __muldi3
[...]

Issue #1567, opened 2024-10-19
RISC-V64: command line argument --riscvfloat does not work


Release-Candidate:
The files generated by `./c3c --target linux-riscv64 --riscvfloat=double compile` use the `soft-float` ABI instead of the `double-float` ABI on RISC-V64. This leads to the following errors:

```shell
./c3c --target linux-riscv64 --riscvfloat=double compile ../resources/examples/hash.c3
...
ld.lld: error: std.hash.o: cannot link object files with different floating-point ABI from /usr/lib/riscv64-linux-gnu/Scrt1.o
ld.lld: error: std.math.o: cannot link object files with different floating-point ABI from /usr/lib/riscv64-linux-gnu/Scrt1.o
ld.lld: error: std.ascii.o: cannot link object files with different floating-point ABI from /usr/lib/riscv64-linux-gnu/Scrt1.o
...
```

An object file generated by the above compiler invocation:

```text
$ file std.core.mem.o
std.core.mem.o: ELF 64-bit LSB relocatable, UCB RISC-V, soft-float ABI, version 1 (SYSV), with debug_info, not stripped
```

but should be:

```text
$ file /usr/lib/riscv64-linux-gnu/Scrt1.o
/usr/lib/riscv64-linux-gnu/Scrt1.o: ELF 64-bit LSB relocatable, UCB RISC-V, RVC, double-float ABI, version 1 (SYSV), for GNU/Linux 4.15.0, not stripped
```

lerno:
From what I read, setting it to double should indeed set compiler.platform.riscv.flen = 8, perhaps there is some additional flag that has to be packed into the LLVM IR for this. Could you perhaps dump the LLVM IR for a proper C compilation of a particular function that fails with C3?

Release-Candidate:
This is a flag in the ELF header: https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#e-flags-layout

LLVM IR: 
It's in the `"target-features"`, I guess. LLVM sets `+64bit,+a,+c,+d,+f,+m,+relax,+zicsr,+zmmul` if i understand that correctly, "d" I guess means RISC-V double support.

```c
int main(void) { return 0; }
```

```text
; ModuleID = 'main.c'
source_filename = "main.c"
target datalayout = "e-m:e-p:64:64-i64:64-i128:128-n32:64-S128"
target triple = "riscv64-unknown-linux-gnu"

[...]

PR #2937, opened 2026-02-14
fixes for RISC-V ABI Implementation #1567

ManuLinares:
Fixed RISC-V floating-point ABI by correcting the target triple to `riscv64-unknown-linux-gnu`, adding the `target-abi` module flag, and ensuring ABI-required CPU features are enabled.

I tested this with:
```bash
build/c3c compile-only --target linux-riscv64 rv_hello.c3
readelf -h obj/linux-riscv64/rv_hello.o | grep Flags
# Output: Flags: 0x5, RVC, double-float ABI
```

```bash
# and qemu because I don't have a riscv machine :/
qemu-riscv64-static -L /usr/riscv64-linux-gnu ./rv_hello
```

---

@lerno I purposedly left these two failing tests to clearly see the difference, so these will make CI fail. Please check this when you have the time.

```
test/test_suite/abi/riscv64-lp64-lp64f-abi-1.c3t
test/test_suite/abi/riscv64-lp64-abi.c3t
```

https://github.com/c3lang/c3c/issues/1567
https://github.com/c3lang/c3c/issues/2014 (this one refers to the stdlib, not the abi)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,6 +73,9 @@ out/
 /cmake-build-debug/
 /cmake-build-release/
 
+CMakeFiles/cmake.check_cache
+CMakeCache.txt
+
 # etags(Emacs), ctags, gtags
 TAGS
 GPATH
@@ -94,8 +97,8 @@ result
 
 # tests
 /test/tmp/*
-/test/testrun
-/test/test_suite_runner
+testrun
+test_suite_runner
 
 # patches, originals and rejects
 *.patch
--- a/benchmarks/stdlib/hash/blake3.c3
+++ b/benchmarks/stdlib/hash/blake3.c3
@@ -1,0 +1,57 @@
+module blake3_bench;
+
+fn void initialize_bench() @init
+{
+	set_benchmark_warmup_iterations(3);
+	set_benchmark_max_iterations(128);
+
+    input = mem::alloc_array(char, BUFSZ);
+    input[:BUFSZ] = (char[]){ [0..BUFSZ-1] = 0xA5 }[..];
+    input_slice = input[:BUFSZ];
+}
+
+fn void teardown_bench() @finalizer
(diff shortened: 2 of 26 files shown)


lerno:
There is something failing here?

ManuLinares:
> There is something failing here?

I don't think so. There are two tests that needs the expected output corrected but I left those without change so you could take a look, as you are more familiar with LLVM IR

test/test_suite/abi/riscv64-lp64-lp64f-abi-1.c3t
test/test_suite/abi/riscv64-lp64-abi.c3t

lerno:
We seem to have something else failing?

src/compiler/linker.c
```diff
@@ -382,29 +386,57 @@ static const char *get_linux_crt_begin_arch_glob(void)
...
+		if (file_exists(arch_linux_64_crt1_path))
+		{
+			const char* arch_linux_path = "/usr/lib64";
+			INFO_LOG("Found crt at %s", arch_linux_path);
+			return arch_linux_path;
+		}
 	}
-	INFO_LOG("Found crt at %s", path);
-	return path;
+
```
lerno, line 440: This seems to be missing some debug info?

lerno:
Alpine Linux to be specific

ManuLinares:
> We seem to have something else failing?
> Alpine Linux to be specific

Fixed now, my bad, forgot to commit a macro.

I moved `target_host_arch()` from `hostinfo.c` to `target.c` and set `linker.c` to use that instead.

lerno:
Looks like the android builds are not passing now?

ManuLinares:
> Looks like the android builds are not passing now?

nah, that's just termux servers intermittently not working

lerno:
Thank you!

PR #2938, opened 2026-02-16
optimize `test_ct_intlog2` test and `whirlpool` hash

ManuLinares:
reduce code bloat for both and makes whirlpool faster


lerno:
Actually you could put all the compile time data into an array, then do a runtime check on that result.

PR #2948, opened 2026-02-17
Optimized adler32 hashing algorithm.

soerlemans:
Optimized adler32 hashing algorithm.
I replaced the current adler32 hashing algorithm with an implementation from https://github.com/weidai11/cryptopp/blob/60f81a77e0c9a0e7ffc1ca1bc438ddfa2e43b78e/adler32.cpp#L8.
Which is public domain code according to the license https://github.com/weidai11/cryptopp/blob/master/License.txt.

The previous implementation would use modulo on every iteration of the `foreach` loop which is extremely slow and not cache optimizable.

Current implementation minimizes the amount of `%` needed and is way way faster for big files.
Personal testing with this algorithm in a separate project of mine shows a roughly 2x speedup.

Other libraries like zlib use similar operations https://github.com/madler/zlib/blob/develop/adler32.c.
This implementation is completely safe, cross platform (works on any system providing a uint which is atleast 32 bit).

There are other optimizations that can be done here for 64 bit systems (increasing bitwidth of a and b and unrolling till 16 bytes) (or more extreme hashing using SIMD).
But this is the best cross section between performance, simplicity and cross platform support.

@weidai11 thank you for your amazing work and contribution to open source.

### DIFF
--- a/lib/std/hash/adler32.c3
+++ b/lib/std/hash/adler32.c3
@@ -41,8 +41,10 @@ fn void Adler32.update(&self, char[] data)
 	usz len = data.len;
 
 	// Align pointer traversing buffer pointer to the unroll alignment size.
-	if (len % UNROLL_SIZE != 0) {
-		do {
+	if (len % UNROLL_SIZE != 0)
+	{
+		do
+		{
 			a += *buf;
 			b += a;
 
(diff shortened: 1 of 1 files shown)


soerlemans:
Also added some rudimentary unit tests.

soerlemans:
Zig has a similar implementation to zlib: **https://github.com/ziglang/zig/blob/738d2be9d6b6ef3ff3559130c05159ef53336224/lib/std/hash/Adler32.zig#L10**

lerno:
Can you update it so that it uses tabs for consistency with the rest of the stdlib?

soerlemans:
Yes updated the whitespace

lerno:
I updated to Allman braces and made the unrolling a macro.

lerno:
Thank you!

PR #2953, opened 2026-02-18
Allow Async w/ Subprocess for POSIX Systems

NotsoanoNimus:
This PR attempts to fix async I/O with controlled subprocesses on POSIX-compliant platforms.

1. ~~Use the `O_NONBLOCK` flag with `pipe2` to create asynchronous pipes when `read_async` is true.~~ _Update_: not using `pipe2`, as it's not universal. Decided to use `pipe` and a `fcntl` method which already exists.
2. Close the original pipes on the forked side _after_ they are duplicated. This wasn't being done before and was leaving the old pipes there.
3. Add `native_fflush` to flush file contents/buffers on demand. This is `@maydiscard` because many implementations can ignore the flush and at least _try_ to proceed, but it can be changed if desired.
4. Actually flush the `fwrite` operation to the target file. This was not happening before and was causing any call to `native_fwrite` to delay writing to the file.

~~Note: also more than happy to find a better way to use `O_NONBLOCK` if sourcing it from `std::net::os` is uncomfortable.~~

### DIFF
--- a/releasenotes.md
+++ b/releasenotes.md
@@ -36,6 +36,7 @@
 - Stream functions now use long/ulong rather than isz/usz for seek/available. 
 - `instream.seek` is replaced by `set_cursor` and `cursor`.
 - `instream.available`, `cursor` etc are long/ulong rather than isz/usz to be correct on 32-bit.
+- Enable asynchronous, non-blocking reads of subprocess STDOUT/STDERR pipes on POSIX systems.
 
 ### Fixes
 - Add error message if directory with output file name already exists


ManuLinares:
Hi, at a quick glance just a question: where are you using `O_NONBLOCK` in here? I don't see it.

NotsoanoNimus:
> Hi, at a quick glance just a question: where are you using `O_NONBLOCK` in here? I don't see it.

Ah, yeah. After I submitted a follow-up change, I took out the `pipe2` and `O_NONBLOCK` and replaced it with a call that already exists to do the `fcntl` for me after creating each pipe. I just didn't have time to update the PR comment.

lerno:
@ManuLinares can you review this?

ManuLinares:
> @ManuLinares can you review this?

Well after testing a little bit, at least my tests don't hang up anymore so I think this is working.

I know they are disabled but these are my tests for `test/unit/stdlib/os/subprocess.c3`

```c3
// Currently disabled as they somewhat mess up the test output.
module std::os::process @test @if(true);
import std::os, std::io, std::thread, std::time;

// ... etc

fn void test_async_read_while_running() @test
{
	$if !env::POSIX:
		return;
	$endif

	// Small shell script that prints, sleeps, then prints again
	String[] command = { "sh", "-c", "echo 'First'; sleep 1; echo 'Second'" };

	SubProcess process = process::create(command,
		{ .search_user_path = true, .read_async = true })!!;
	defer process.destroy();

	char[100] buffer;

	// Read first output
	usz bytes_read = 0;
	while (bytes_read == 0)
	{
		bytes_read = process.read_stdout(&buffer[0], buffer.len)!!;
		if (bytes_read == 0) thread::sleep(time::ms(10));
	}

	assert(((String)buffer[:bytes_read]).contains("First"), "Expected 'First' in output, got %s", (String)buffer[:bytes_read]);

	// Now try to read again immediately, should return 0 (EAGAIN) because of sleep 1
	bytes_read = process.read_stdout(&buffer[0], buffer.len)!!;
	assert(bytes_read == 0, "Expected 0 bytes read while process is sleeping, got %d", bytes_read);

	// Wait for the second output
	DString out_acc = dstring::temp_with_capacity(100);
	while (true)
	{
		bytes_read = process.read_stdout(&buffer[0], buffer.len)!!;
		if (bytes_read > 0)
		{
			out_acc.append(buffer[:bytes_read]);
			if (out_acc.str_view().contains("Second")) break;
		}
		else
		{
			if (!process.is_running()!!) break;
			thread::sleep(time::ms(10));
		}
	}

	assert(out_acc.str_view().contains("Second"), "Expected 'Second' in output, got %s", out_acc.str_view());

	process.join()!!;
}

fn void test_write_to_stdin_fflush() @test
{
	$if !env::POSIX:
		return;
	$endif

	// Use cat to echo back stdin
	String[] command = { "cat" };

	SubProcess process = process::create(command,
		{ .search_user_path = true, .read_async = true })!!;
	defer process.destroy();

	String msg = "Hello Async World\n";
	usz written = process.write_to_stdin((char[])msg)!!;
	assert(written == msg.len, "Expected to write %d bytes, wrote %d", msg.len, written);

	// Wait a bit for cat to process input
	thread::sleep(time::ms(50));

	char[100] buffer;
	usz bytes_read = process.read_stdout(&buffer[0], buffer.len)!!;
	assert(bytes_read == msg.len, "Expected to read back %d bytes, got %d", msg.len, bytes_read);
	assert((String)buffer[:bytes_read] == msg, "Content mismatch");

	process.join()!!;
}
```

lerno:
Can we have some notes in releasenotes, then I can merge it.

NotsoanoNimus:
Done. Updated from mobile so please double-check the placement in the release notes.

lerno:
Looks good

ManuLinares:
Sup, I found that if not setting these fcntl, processes running in threads don't set up their respective pipes 😔 and wait for their parent ones, more 😔

```diff
diff --git a/lib/std/os/process.c3 b/lib/std/os/process.c3
index 70354310e..fae3fc9ce 100644
--- a/lib/std/os/process.c3
+++ b/lib/std/os/process.c3
@@ -235,6 +235,17 @@ fn Process? spawn(String[] command_line, SpawnOptions options = { }, String[] en
 		if (posix::pipe(&stdoutfd)) return FAILED_TO_OPEN_STDOUT~;
 		if (!(options & STDERR_TO_STDOUT) && posix::pipe(&stderrfd)) return FAILED_TO_OPEN_STDERR~;
 
+		// Set FD_CLOEXEC to prevent leaking pipe handles between concurrent spawns
+		os::fcntl((NativeSocket)stdinfd[0], 2, 1);
+		os::fcntl((NativeSocket)stdinfd[1], 2, 1);
+		os::fcntl((NativeSocket)stdoutfd[0], 2, 1);
+		os::fcntl((NativeSocket)stdoutfd[1], 2, 1);
+		if (!(options & STDERR_TO_STDOUT))
+		{
+			os::fcntl((NativeSocket)stderrfd[0], 2, 1);
+			os::fcntl((NativeSocket)stderrfd[1], 2, 1);
+		}
+
 		if (options & ASYNC_OUTPUT)
 		{
 			((NativeSocket)stdoutfd[0]).set_non_blocking(true)!;
```

since you have done some testing, maybe you can test this. I'll try to make a few minimal examples and maybe an issue later.

PR #2963, opened 2026-02-20
deflate/zip: impl `compress_stream()` and optimizations

ManuLinares:
Done

```
implement `compress_stream`

dyn huffman + compress fastpass

various optimizations

11-bit lookup table for Huffman decoding

128bit simd vector moves for matches to maximise throughput

rebuilt the bitstream reader around a 64bit buffer and refactored the
decoder into a tight hot loop to eliminate statr machine overhead and
align the sliding window to 64bit boundaries

add `test_deflate_embedded_stream()` test

add more zip tests
```

### DIFF
--- a/lib/std/io/os/file_libc.c3
+++ b/lib/std/io/os/file_libc.c3
@@ -129,21 +129,34 @@ macro fault file_seek_errno() @local
 }
 
 
-struct Utimbuf
+struct Utimbuf @if(!env::WIN32)
 {
 	Time_t actime;
 	Time_t modtime;
 }
-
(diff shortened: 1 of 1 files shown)


data-man:
I think that `StreamBitReader` and `BitWriter` can be useful as public in `std::io`.
What do you think?

lerno:
Please look at deflate, because I had to update it.

ManuLinares:
> Please look at deflate, because I had to update it.

Done, I still couldn't debug why it breaks in windows and alpine.

lerno:
First thing I'd check is if it works if the vector usage is removed.

PR #2964, opened 2026-02-20
More x86 instructions

srkkov:
- Added all BMI1 instructions.
- Added LZCNT.
- Added all BMI2 instructions.
- Fixed ADCX entry and added ADOX, that completes ADX set.
- Added PCLMULQDQ and VPCLMULQDQ.
- Added all SSE4.2 instructions, and their VEX versions (except EVEX PCMPGTQ).
- Merged POPCNT entries, amended associated test.
- Added VZEROALL and VZEROUPPER.
- Added cryptographic instruction sets including AES, VAES, AESKLE, AES_WIDE, KEY_LOCKER, SHA, SHA512, SM3, SM4.
- Added RDRAND and RDSEED.
- Added XGETBV instruction.


srkkov:
I merged all of the `popcnt` entries into one, llvm in infers it correctly since it has a simple inference, and I think in the future we should also omit suffixes on simple inference cases. All `crc32` entries are also merged, but it will let in wrong encodings since `crc32` has wonky encodings. llvm (AT&T) actually assigns suffixes based off of the source operand size here, so we can't just split it into `crc32l` and `crc32q` unfortunately. I decided that letting in wrong encodings is fine since the current way we register instructions already does that (ex. `vaddps $xmm0, $ymm1, $zmm2` is allowed).
```
CRC32 r32, r/m8
CRC32 r32, r/m16
CRC32 r32, r/m32
CRC32 r64, r/m8
CRC32 r64, r/m64
```
I know we want to automate it, so I just focused on instructions I though were important to have early, and that would not be subject to suffixes.

lerno:
Looks like it ran into some bug?

srkkov:
It's because I changed `adcxl,adcxq -> adcx` AND because originally, it tested that `adcxl $eax, $ecx` will clobber `$ecx`, which is wrong, but the instruction entry was also wrong (first operand should be rw), so it passed. (Same deal with adcxq)
```c
reg_instr_clob(target, "adcxl", cc_flag_mask, "r32, rw:r32/mem"); // original
```
So it should be fixed now. Sorry I didn't run the test_suite_runner beforehand.

lerno:
Thank you!

PR #2966, opened 2026-02-22
zip: read AES-encrypted zip files (AE-1 and AE-2)

konimarti:
Read AES-encrypyted Zip files using the AE-1 or AE-2 format [1];

[1]: https://www.winzip.com/en/support/aes-encryption/

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -613,7 +613,7 @@ jobs:
 
   release:
     runs-on: ubuntu-22.04
-    needs: [build-msvc, build-linux, build-mac, build-linux-alpine, build-openbsd, build-netbsd, build-android]
+    needs: [build-msvc, build-linux, build-mac, build-linux-alpine, build-openbsd, build-netbsd, build-android, build-linux-musl-static]
     if: github.ref == 'refs/heads/master'
     env:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -629,6 +629,7 @@ jobs:
 
           mv c3-linux-Release/c3-linux-Release.tar.gz c3-linux.tar.gz
           mv c3-linux-musl-Release/c3-linux-musl-Release.tar.gz c3-linux-musl.tar.gz
+          mv c3-linux-static-Release/c3-linux-static-Release.tar.gz c3-linux-static.tar.gz
           mv c3-openbsd-Release/c3-openbsd-Release.tar.gz c3-openbsd.tar.gz || true
           mv c3-netbsd-Release/c3-netbsd-Release.tar.gz c3-netbsd.tar.gz || true
           mv c3-android-aarch64-Release/c3-android-aarch64-Release.tar.gz c3-android-aarch64.tar.gz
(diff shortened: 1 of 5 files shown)


lerno:
Can the Zip password ever *be* an empty string?

konimarti:
The empty string for the password doesn't make sense but I just added it as the default function value when it's not encrypted. I need to reconsider that. Also, there's an issue with an encrypted deflate stream that I initially didn't see and haven't resolved yet. This needs more work.

konimarti:
I discovered that WinZip’s AES-CTR uses a little-endian counter, so I had to increment the IV starting at the lowest-order byte instead of the highest (which is the common way). After adding a little-endian counter to AES-CTR, the decrypted DEFLATE streams should be correct now.

konimarti:
The failed nix test is because of a stochastic element in std::math::distributions. I thought I removed most of them but apprently not. Will provide a fix for this soon, too.

konimarti:
> Can the Zip password ever _be_ an empty string?

I changed the password handling to a callback which provides for a cleaner api.

konimarti:
For testing purposes, AES-encrypted zip files can be created with 7z:

```7z a -tzip -mem=AES256 -pYourPassword secure.zip file_to_encrypt.txt```

**Comment by lerno** [comment]
Are you imagining that the callback would pop some dialog or something? I think more reasonable would be to fail on PASSWORD_NEEDED if the password isn't provided and it's passworded, and a PASSWORD_MISMATCH if the password doesn't work.

**Comment by konimarti** [comment]
It was meant for flexibility: you can have different passwords for different zip entry files in the same archive and an empty string as a password is technically valid. I'm fine with changing it back to have the password in the function signature, though.

**Comment by lerno** [comment]
I'm sorry but I think this needs to be merged now.

File: lib/std/compression/zip.c3 [hunk 1]
```diff
@@ -623,33 +628,21 @@ fn ZipEntry? ZipArchive.stat(&self, String filename)
...
-	{
-		if (reader.adapter.start_offset > (ulong)isz.max) return io::OVERFLOW~;
-		self.file.set_cursor(reader.adapter.start_offset)!;
-		if (entry.compressed_size > (ulong)usz.max) return io::OVERFLOW~;
-		char[] compressed = allocator::alloc_array(allocator, char, (usz)entry.compressed_size);
-		defer allocator::free(allocator, compressed);
-		self.file.read(compressed)!;
-		data = deflate::decompress(allocator, compressed)!;
-	}
+	data = io::read_fully(allocator, &reader)!;
```
lerno, line 644: I am concerned about this change here. It looks like it's completely ignoring the old behaviour?

konimarti: `ZipArchive.read_file_all` duplicates the STORE/DEFLATE dispatch logic that `ZipEntryReader.read` already handle. The duplication can be removed by just using the `ZipEntryReader`. Unit tests pass. @ManuLinares can you confirm?

ManuLinares: Yes.

Could you resolve the conflicts so I can review this?

ManuLinares:
LGTM 🚀

lerno:
Thank you!

PR #2975, opened 2026-02-24
distributions: fix tests

konimarti:
Remove stochastic elements in the distributions tests and ensure they are deterministic.


lerno:
👍

PR #2976, opened 2026-02-24
vendor-fetch lists available libraries

ManuLinares:
from `github.com/c3lang/vendor` when no library is specified.

### DIFF
--- a/releasenotes.md
+++ b/releasenotes.md
@@ -21,6 +21,7 @@
 - Properly support `@deprecated` as contract.
 - Support deprecating enum values.
 - Improve error when trying to use an extern const as a compile time constant. #2969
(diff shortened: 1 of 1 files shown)


lerno:
Add to releasenotes?

ManuLinares:
> Add to releasenotes?

No problem

lerno:
Awesome! We really need to make a package handler... we're so close to make it great.

ManuLinares:
> Awesome! We really need to make a package handler... we're so close to make it great.

we should define the scope like, managing versions and what else?

ˋˋˋ
c3c add/install
c3c remove/rm/delete
c3c search/find
c3c list
ˋˋˋ

??

PR #2978, opened 2026-02-25
Add Linux's syscalls

data-man:
I'll improve it later for more targets if this PoC is OK.


lerno:
If I understand correctly it becomes like `Syscall.READ.call(a, b, c)`, why shouldn't we simply have: `linux::syscall_read(a, b, c)`? `linux::syscall_write(a, b c)`

data-man:
> `linux::syscall_read(a, b, c)`? `linux::syscall_write(a, b c)`

Add hundreds of functions?

data-man:
Or just add constdefs?

lerno:
I don't see how using constants/enums would be better than macros?

data-man:
Some std modules uses this pattern:
```c3
constdef Foo : inline uint
{
    FOO1        = 1,
    FOO2        = 2
}

const uint BAR1 = Foo.FOO1;
const uint BAR2 = Foo.FOO2;
```

So, what about this:
```c3
constdef SysCall : inline uint
{
    <* (unsigned int fd, char *buf, size_t count) *>
    READ         = 0,
    <* (unsigned int fd, const char *buf, size_t count) *>
    WRITE        = 1

    etc.
}

<* (unsigned int fd, char *buf, size_t count) *>
const uint SYS_READ = SysCall.READ;

<* (unsigned int fd, const char *buf, size_t count) *>
const uint SYS_WRITE = SysCall.WRITE;

etc.
```

lerno:
That doesn't answer why it couldn't be macros? Wouldn't that be nicer?

data-man:
> That doesn't answer why it couldn't be macros? Wouldn't that be nicer?

How useful are hundreds of macros without ported Linux types to C3 types?

lerno:
You are suggesting:

```c3
SYS_READ.call(fd, buf, count); 
```
With everything untyped.

I don't see why one shouldn't do
```c3
macro sys_read(uint, char* buf, usz count) { ... }
```

It seems like a gross misuse of macros and enums?

lerno:
This PR has been dead now for over a month, what should we do with it?

lerno:
Closed due to inactivity

Issue #2981, opened 2026-03-01
[Feature Request] Implement hash method for ZString


fesoliveira014:
Strings and char[] implement a hash method, allowing them to be used in hashmaps, hashsets, etc, but ZStrings do not. This seems like an oversight, so we should implement it.

lerno:
This is added now so I close it.

PR #2982, opened 2026-03-01
Add hash method for ZString

fesoliveira014:
Fixes issue #2981 by adding a hash method for ZString in std::core::builtin and adds a test for it.


lerno:
Thank you!

PR #2988, opened 2026-03-04
Refactor LLVM definitions handling in CMakeLists

MehdiChinoune:
That's the better way to use them, see https://github.com/llvm/llvm-project/blob/llvmorg-22.1.0/llvm/docs/CMake.rst#developing-llvm-passes-out-of-source

### DIFF
--- a/install/install.ps1
+++ b/install/install.ps1
@@ -5,7 +5,8 @@
     This script installs C3 on Windows from the command line.
 .PARAMETER C3Version
     Specifies the version of C3 to install.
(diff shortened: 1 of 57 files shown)


lerno:
What does it do, and how does it improve things?

MehdiChinoune:
> What does it do, and how does it improve things?

It treats definitions as separate not as a single string, for example "-DA=1 -DB", if they are not separated the compiler could interpret it (or cmake pass it as) -DA="1 -DB". That happened with GCC on MinGW-w64.

lerno:
Ah ok

PR #2993, opened 2026-03-05
fix #2906: auto-fallback to built-in linker

### LINKED ISSUE #2906: Detect if a C compiler is missing in the environment
If a linker is missing in a Linux environment, there is no immediate fallback even though we have a built-in linker. Detect this situation. Motivated by: https://github.com/c3lang/c3-web/pull/162

ManuLinares:
- Search linker in PATH
- Use builtin linker if CC missing

https://github.com/c3lang/c3c/issues/2906


lerno:
Thank you!

PR #2995, opened 2026-03-05
feat(std::core::string): add try_to_* numeric parsing and generic *_or fallbacks

fernandolguevara:


Book-reader:
c3 has an error handling feature that does the same thing as `to_integer_or`:
```c3
// default to 7 if it fails
int x = "12345".to_int() ?? 7;
```
and there are a few ways to recreate `try_to_int` with the error handling syntax (see https://c3-lang.org/language-common/optionals-essential/#-unwrapping-an-optional), as well as with the `@try` macro depending on your usecase:
https://github.com/c3lang/c3c/blob/e2140b8e776cd716fab74e6f55bc7a29e1690f79/lib/std/core/builtin.c3#L536-L549

```c3
int y;

if (catch @try(y, "12345".to_int()))
{
  // it failed...
}
// if it succeeded y is now set to the result


int? z = "12345".to_int();

if (catch z)
{
  // it failed...
  // returning or calling any @noreturn function like abort() will unwrap z and allow it to be used like a normal variable after this if statement
  // z can also be used like a normal value in the else branch even if the if branch doesn't return
  return;
}
// z can be used as a normal variable since it's guaranteed to have a value


int a = "12345".to_int()!!; // panic if it fails

int? b = "12345".to_int();
if (try b)
{
 // b can be used like a normal variable
}


if (try c =  "12345".to_int())
{

}
```

Issue #2806, opened 2026-01-23
AES implementation is not constant time


laura240406:
I know that this is probably just a minor problem right now as nobody is using this in a context where this would be an issue but we should document this somewhere or fix it.

The issue is that the SBOX lookup is done by looking in a 256 byte table, which can be a major issue as [explained by djb](https://cr.yp.to/antiforgery/cachetiming-20050414.pdf), which can even allow full key recovery over the network in a realistic setting.

There are 3 ways I can think of that could be used to mitigate this:
1. Bitslicing the SBOX as [demonstrated here](https://gist.github.com/oreparaz/939f83695aab0e3ca0a0). This is constant time since it basically evaluates a circuit representing the SBOX transformation. This is medium slow.
2. Reading each entry of the SBOX in sequence, ANDing it with a mask that is generated from the index in constant time (0x00 for all entries except for the one we want which gets 0xff) and ORing it into a final result that is returned. This is really slow as it doesn't do anything 99.6% of the time in the SBOX part
3. Use AES hardware instructions like AES-NI or the ARM ones. This is limited to platforms that support it but is really fast.

I think option 1 is the best (or option 2 if you don't want that much complexity) and option 3 could be optionally added to improve speed.
I might try to implement this proposal soon but I have exams coming up.

What are your thoughts on this?

laura240406:
The most simple implementation would be option 2 with:

```diff
diff --git a/lib/std/crypto/aes.c3 b/lib/std/crypto/aes.c3
index b42a8a13..72026527 100644
--- a/lib/std/crypto/aes.c3
+++ b/lib/std/crypto/aes.c3
@@ -373,8 +373,27 @@ fn void ctr_xcrypt_buffer(Aes *aes, char[] in, char[] out)
 	 }
 }
 
-macro char get_sbox_value(num) => SBOX[num];
-macro char get_sbox_invert(num) => RSBOX[num];
+fn char get_sbox_value(char num)
+{
+    char res = 0;
+
+    for (int i = 0; i < 256; i++) {
+        res |= SBOX[i] & (char) ((((i ^ num) - 1) >> 31) & 0xff);
+    }
+
+    return res;
+}
+
+fn char get_sbox_invert(char num)
+{
+    char res = 0;
+
+    for (int i = 0; i < 256; i++) {
+        res |= RSBOX[i] & (char) ((((i ^ num) - 1) >> 31) & 0xff);
+    }
+
+    return res;
+}
 
 const char[256] SBOX =
[...]

PR #2996, opened 2026-03-05
This PR addresses the side-channel issues in the current AES implementation (issue #2806).

ManuLinares:
https://github.com/c3lang/c3c/issues/2806

I have refactored the core transformations (SBOX, ShiftRows, MixColumns) to use a bitsliced approach instead of the old table lookups. To keep the performance from tanking, I used C3 generics to implement a parallel path, it handles 4 blocks at once using ulong bit-planes.

I've tested it against the existing unit tests and everything is passing. Benchmarking CTR mode shows it's actually about 2.8x faster than the single-block bitsliced version I started with.

```
 build/c3c compile-benchmark -O1 --stdlib lib/std/ benchmarks/stdlib/crypto/aes_bench.c3

------------------------ BENCHMARKS -------------------------
Benchmarking std::crypto::aes_bench::bench_ctr_xcrypt .. [COMPLETE] 3.23 microseconds, 13573.43 CPU clocks, 10000 iterations (runtime 32.32 milliseconds)
```

Let me know if the code style or the generic implementation needs any tweaks, I would like to implement a path for using AES hardware instructions but hit a snag in `sema_asm.c`, doesn't support passing vectors as asm arguments which is totally fixable.

### DIFF
--- a/releasenotes.md
+++ b/releasenotes.md
@@ -18,6 +18,7 @@
 - Improved stringify. 
 - PollSubscribe was incorrectly an int instead of ushort. #2997
 - SubProcessOptions.search_user_path did nothing on non-windows systems despite comment saying it should #2845
+- AES implementation fixed to be constant time #2806
 
 ## 0.7.10 Change list
 


lerno:
Thank you!

Issue #2997, opened 2026-03-05
PollSubscribe should be of size ushort.


dazoe:
I was attempting to use the poll api and discovered it wasn't functioning correctly. I was getting 0 back in revents and my events was being changed. Looks like events was defaulting to a larger size then was expected.
PollSubscribe should be `constdef PollSubscribe : ushort`

https://github.com/c3lang/c3c/blob/e2140b8e776cd716fab74e6f55bc7a29e1690f79/lib/std/net/socket.c3#L28C9-L28C23

ManuLinares:
this can be closed: https://github.com/c3lang/c3c/pull/2998

PR #2998, opened 2026-03-05
set size of PollSubscribe to ushort

dazoe:
Fix for #2997


lerno:
Thank you! Next time, consider also updating the release notes though.

PR #2999, opened 2026-03-05
fix(ps-install): ensure c3-windows-release folder exists before renaming

fernandolguevara:
```
PS C:\Users\vivaperon> iwr -useb https://raw.githubusercontent.com/c3lang/c3c/refs/heads/master/install/install.ps1 | iex

This script will automatically download and install C3 (latest) for you.
Getting it from this url: https://github.com/c3lang/c3c/releases/latest/download/c3-windows.zip
The binary will be installed into 'C:\Users\vivaperon\.c3'
Rename-Item : Cannot rename because item at 'C:\Users\vivaperon/c3-windows-Release' does not exist.
At line:163 char:5
+     Rename-Item -Path "$Env:USERPROFILE/c3-windows-Release" -NewName  ...
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [Rename-Item], PSInvalidOperationException
    + FullyQualifiedErrorId : InvalidOperation,Microsoft.PowerShell.Commands.RenameItemCommand
```

PR #3001, opened 2026-03-05
install.ps1: fix and improve Windows install script

ManuLinares:
- Fix Rename-Item crash when zip extracts to an unexpected folder name. Now discovers the extracted folder dynamically
- Extract to a dedicated temp dir instead of $USERPROFILE to avoid leftover clutter
- Add `-UseBasicParsing` to Invoke-WebRequest to fix failures on Server Core / CI environments
- Suppress progress bar for faster downloads on PS5
- Print the real exception in catch instead of a hardcoded generic message
- Fix PATH deduplication
- Add prerelease as a recognized version keyword mapping to latest-prerelease-tag

### DIFF
--- a/lib/std/collections/object.c3
+++ b/lib/std/collections/object.c3
@@ -78,7 +78,7 @@ fn usz? Object.to_format(&self, Formatter* formatter) @dynamic
 
 fn Object* new_obj(Allocator allocator)
 {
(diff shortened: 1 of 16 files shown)


ManuLinares:
apparently this works fine: https://github.com/c3lang/c3c/pull/2999#issuecomment-4008297534

lerno:
Thank you

PR #3002, opened 2026-03-06
add `fault.short_name` method to get just the name without the module path

LowByteFox:
Adds a method `short_name` to all faults, useful when program wants to display name of the fault to the user without the full module path:
```c3
module main;
import std::io;

faultdef FOO;

fn int main(String[] args)
{
	io::printn(FOO); /* main::FOO */
	io::printn(FOO.short_name()); /* FOO */
	return 0;
}
```


lib/std/core/builtin.c3
```diff
@@ -1071,3 +1071,14 @@ fn void install_signal_handlers() @init(101) @local @if(env::BACKTRACE)
...
 }
+
+fn String fault.short_name(self)
+{
+	String name = self.nameof;
+	return name[name.index_of("::") + 2..]!!;
+}
+
+<*
+ @require $defined(var $v = $self)
```
lerno: This one should be unnecessary.

lerno:
No test cases? 👀

lib/std/core/builtin.c3
LowByteFox: Just tried it without it, works just fine

I kept it there because of the code I sent in discord.

LowByteFox:
right... I forgot, let me add some quickly 😅

LowByteFox:
added tests and removed the macro since `.nameof` for constant fault literals gives the name itself without parent module path, if this is a bug the macro can be recovered.

the documentation has this mentioned here https://c3-lang.org/language-overview/types/#fault-nameof however that happens only with runtime unknown faults

lerno:
Thank you

PR #3003, opened 2026-03-06
Add last_modified function

### LINKED ISSUE #2769: Std::io::file missing functions
```c3
import std::io::file;

File? f = file::open(String filename, String mode);
File? f = file::open_path(Path path, String mode);
bool b = file::exists(String file);
File f = file::from_handle(CFile file);
bool b = file::is_file(String path);
bool b = file::is_dir(String path);
usz? n = file::get_size(String path);
void? file::delete(String filename);

File f;
void? f.reopen(String filename, String mode);
usz? n = f.seek(isz offset, Seek seek_mode = Seek.SET|CURSOR|END) @dynamic;
```
This is all at std::io::file.

Seems to me, two most frequent ones are missing.
```c3
ulong? mtime = file::last_modified(String filename);
File f;
void? f.flush() @dynamic; // conformed to OutStream
```

Flanderzz:
Closes #2769 

- Adds a `file::modified_time()` function

### DIFF
--- a/test/unit/stdlib/io/fileinfo.c3
+++ b/test/unit/stdlib/io/fileinfo.c3
@@ -1,5 +1,5 @@
 module std::io::fileinfo @test;
-import std::io::os, std::io::file;
+import std::io, std::math;
 
 fn void test_native_is_file() @if(env::LINUX || env::DARWIN || env::ANDROID)
 {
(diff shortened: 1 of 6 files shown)


Flanderzz:
to me, Time_t? makes more sense for a return type than the proposed ulong?, but i can change if needed.

lerno:
I expected a `Time` from time.c3, not the libc `Time_t`. Could you make the change?

test/unit/stdlib/io/fileinfo.c3
```diff
@@ -31,3 +31,9 @@ fn void test_native_is_dir() @if(env::BSD_FAMILY)
 	assert(!os::native_is_dir("/dev/sd0a"));
 	assert(!os::native_is_dir("/dev/null"));
 }
+
+fn void test_native_get_modified_time() @if(env::LINUX || env::DARWIN || env::ANDROID || env::BSD_FAMILY)
+{
+	assert(@ok(os::native_get_modified_time("/etc/passwd")));
```
lerno: Uh... better create a temp file and check its modified time.

Flanderzz: Fixed

Flanderzz:
> I expected a `Time` from time.c3, not the libc `Time_t`. Could you make the change?

fixed to return a Time not a Time_t

lerno:
Thank you!

PR #3004, opened 2026-03-06
Benchmark Runner QoL Improvements

ManuLinares:
Fixed merge conflicts of https://github.com/c3lang/c3c/pull/2672

---

From original PR by @NotsoanoNimus :

 I work with `compile-benchmark` quite often - only fair I help with its
upkeep. Slapped this together in a couple hours, because I really need
more consistent results and a way to visualize them, and the benchmark
runner is a bit of a rat's nest.

Changes, in no particular order:
- Add a MEDIAN metric to the results
  - The _mean_ gives us throughput information, but it's too
heavily skewed by a set's outliers.
  - The _median_ gives us an idea of performance expectations with
a 50% probability. Put another way, the tested function has performed at
or better than the median in 50% of samples taken.
  - Caveat: a sorted set is required. For high-volume iterations,
this can get expensive.
- Improve the output units and refactor that into `NanoDuration`,
so it can be used elsewhere as desired
- Provide some pretty colors, oooooo
- Added CSV reporting option to get a resultant data-set
  - Not the most useful thing, but makes it easy to plug benchmark
outputs into other software. Could probably be improved upon. and yeah
I'm a boomer who likes CSV, sue me
- Restructure the benchmark runtime to be more like the tester
runtime
- Rudimentary command-line options to adjust benchmark options


lerno:
I would drop the median in favour of standard deviation

ManuLinares:
> I would drop the median in favour of standard deviation

I added standard deviation by default (no need for a flag)
And a few tweaks here and there on colors and stuff.
Progress bar doesn't affect bench results.

ManuLinares:
I tweaked all stdlib benchmarks to target ~300ms execution time

image

lerno:
I just fixed this. You should revert your changes.

lerno:
(And merge with the latest)

lerno:
Out of scope for this one, but the benchmarking progress indicator is an old thing and works poorly, and I think it should be removed anyway:

```
Benchmarking crypto_hash_benchmarks::streebog_512_1mib .............. [####################] 980 / 1024 (96%)
Benchmarking crypto_hash_benchmarks::streebog_512_1mib .............. [####################] 990 / 1024 (97%)
Benchmarking crypto_hash_benchmarks::streebog_512_1mib .............. [####################] 1000 / 1024 (98%)
Benchmarking crypto_hash_benchmarks::streebog_512_1mib .............. [####################] 1010 / 1024 (99%)
Benchmarking crypto_hash_benchmarks::streebog_512_1mib .............. [####################] 1020 / 1024 (100%)
Benchmarking crypto_hash_benchmarks::streebog_512_1mib .............. [COMPLETE] 5.44 milliseconds, 5437982.00 CPU clocks, 1024 iterations (runtime 5.57 seconds)
```

lerno:
Another point of improvement would be to add a symbol for standard deviation instead.

PR #3006, opened 2026-03-07
android: add -nostartfiles to avoid duplicate CRT objects

### LINKED ISSUE #2984: Duplicate symbol conflicts when compiling executables on Termux with --target matching host
When attempting to compile on Termux, explicitly passing the --target flag while the target architecture matches the host system, a duplicate symbol conflict occurs. This happens because the compiler forces the inclusion of crtbegin_dynamic.o from both Termux’s internal libraries and the NDK at the same time. This conflict only arises when generating executables; dynamic libraries work normally.

lualvsil:
When building on Termux with an explicit --target matching the host architecture, both the host toolchain and the Android NDK inject startup files (e.g. crtbegin_dynamic.o), causing duplicate symbol errors during linking.

Adding -nostartfiles prevents the host toolchain from injecting its own startup files while allowing C3 to pass the required Android CRT objects explicitly.

Tested on Termux. Additional confirmation for cross-compilation to Android from other hosts would be appreciated.

Fixes #2984

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -613,7 +613,7 @@ jobs:
 
   release:
     runs-on: ubuntu-22.04
-    needs: [build-msvc, build-linux, build-mac, build-linux-alpine, build-openbsd, build-netbsd, build-android]
+    needs: [build-msvc, build-linux, build-mac, build-linux-alpine, build-openbsd, build-netbsd, build-android, build-linux-musl-static]
     if: github.ref == 'refs/heads/master'
     env:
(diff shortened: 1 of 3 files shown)


ManuLinares:
I've tested this PR on linux with `build/c3c compile --target android-aarch64 --android-ndk android-ndk-r27d/ asd.c3` and it works

and on Android/termux indeed removes the `Failed to link executable '' using command ...`

lerno:
Can you add something to the release notes as well?

lerno:
Thank you!

PR #3007, opened 2026-03-07
gzip: add gzip compress/decompress implementation

konimarti:
Implements RFC 1952 gzip file format with support for all header flags (FHCRC, FEXTRA, FNAME, FCOMMENT), Latin-1 filename/comment encoding, and CRC32 integrity verification.


lerno:
I updated a little with bitstruct and such. Thank you!

Issue #1745, opened 2024-12-30
feature request: allow building runtime switch statements at compile time with `$for/$foreach/$if`


Book-reader:
```zig
struct Foo
{
  Bar bar;
  BarType bar_type;
}

union Bar
{
  char as_CHAR;
  uint as_UINT;
}

enum BarType
{
  CHAR,
  UINT,
}

fn void main()
{
  Foo f = get_foo();

  switch (f.bar_type)
  {
    $foreach ($type : BarType)
      case $type:
        io::printfn("f was of type %s with value %s", f.bar_type, f.bar.$eval("as_" +++ $type.nameof));
    $endforeach
  }
}
```
the current equivalent is this, which is not as elegant and the compiler may not be able to optimize it as much (unsure about this):
```zig
if SWITCH: (true)
{
  $foreach ($type : BarType)
    if (f.bar_type == $type)
    {
      io::printfn("f was of type %s with value %s", f.bar_type, f.bar.$eval("as_" +++ $type.nameof));
      break SWITCH;
    }
  $endforeach
}
```

this is low priority for me

HMart81:
I also think is useful, so has my vote.

One example where I wanted to use this

```c
switch(input_text) 
    {
$if $defined(TextInputBox):
        case "exit":
            exitConsoleInput();
$else
        case "exit":
$endif
        case "quit":
            exitConsoleInput();
            quit();
        case "clear":
            clearConsoleInputBuffer();
        default:
            io::printfn("Unknown command: %s", input_text);
    }```

lerno:
@HMart81 that particular case could be handled by $if inside of the case and an nextcase, right?

ManuLinares:
Building `switch` statements at compile-time can be done with `$expand`

I don't know if this will be "expanded in the future" in the macro system itself, but wanted to write this here; the long switch statement in [builtin.c3](https://github.com/c3lang/c3c/blob/master/lib/std/core/builtin.c3#L792) `get_frameaddress()` and `get_returnaddress` can be:

```c3
const MAX_FRAMEADDRESS = 128;
<*
 @require n >= 0
*>
macro void* get_frameaddress(int n)
{
	if (n > MAX_FRAMEADDRESS) return null;
	var $code = "switch (n) {\n";
	$for var $i = 0; $i <= MAX_FRAMEADDRESS; $i++:
		$code = $code +++ @sprintf("case %s: return $$frameaddress(%s);\n", $i, $i);
	$endfor
[...]

PR #3008, opened 2026-03-07
Improve Linux backtrace readability and symbolization

ManuLinares:
- Clean up backtrace output by stripping internal panic and runtime startup frames.
- Enable search_user_path for addr2line and realpath to fix Linux symbolization.
- Repetition compression for deep recursive stacks.
- Simplify and compress frame address macros
   - Before: 128 manual cases `O(1) jump table`
   - Now: Compact binary search O(log n) path

Thanks @BWindey


lib/std/core/builtin.c3
```diff
@@ -185,29 +203,52 @@ fn bool print_backtrace(String message, int backtraces_to_ignore, void *added_ba
...
+		usz repetition_count = 0;
 		foreach (i, &trace : backtrace)
 		{
 			if (i < backtraces_to_ignore) continue;
-			String inline_suffix = trace.is_inline ? " [inline]" : "";
-			if (trace.is_unknown())
-			{
-				io::eprintfn("  in ???%s", inline_suffix);
-				continue;
+			if (trace.has_file() && trace.file.ends_with("/lib/std/core/private/main_stub.c3")) {
```
data-man, line 219: What about `--stdlib <dir>"` switch?

ManuLinares: It'll print backtrace with a few lines added. I think this is good.

lerno:
I'm not sure I understand the point of changing the code for getting the frame address

ManuLinares:
> I'm not sure I understand the point of changing the code for getting the frame address

Your choice 👍, I remember a conversation where you said this could be simplified.

It's less generated code in the binary for a path that doesn't need to be fast, and the new code it's only 7 instructions.

Also, less lines of code make reading the stdlib builtin better in my opinion.

lerno:
> > I'm not sure I understand the point of changing the code for getting the frame address
> 
> Your choice 👍, I remember a conversation where you said this could be simplified.
> 
> It's less generated code in the binary for a path that doesn't need to be fast, and the new code it's only 7 instructions.
> 
> Also, less lines of code make reading the stdlib builtin better in my opinion.

Leave it as it was. The new version requires reading macro code, the old one didn't. And we might get https://github.com/c3lang/c3c/issues/1745 at which point it would be easy to write a readable version of it.

lerno:
Don't forget the release notes! :D

PR #3018, opened 2026-03-10
feat(thread): add OneShotChannel

fernandolguevara:
```c3
import std::io;
import std::thread::channel;
fn void main(String[] args)
{
	OneShotChannel {int} channel;
	channel.init(mem)!!;
	defer channel.destroy();

	Thread thread;
	thread.create(fn int (void* arg)
	{
		OneShotChannel {int} channel = (OneShotChannel {int})arg;
		channel.push(123)!!;
		return 0;
	}, (void*)channel)!!;
	defer thread.join();

	int value = channel.pop()!!;
	io::printfn("pop: %d", value);
}
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -68,3 +68,23 @@ pristine with the highest possible quality.
 ## 4. How to contribute to various tools
 
 In general, file a pull request. Depending on who maintains it, rules may differ.
+
+## 5. AI usage in contributions
+
+AI tools do not change contributor responsibilities.
+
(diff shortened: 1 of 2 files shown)


fernandolguevara:
@lerno, Any thoughts on this PR?

lerno:
I'm on the fence

lerno:
I've been thinking whether it's important enough to include or not.

lerno:
Thank you for this. Upon review I also found a bug in the UnbufferedChannel by accident!

PR #3019, opened 2026-03-11
Unified `fetch-sdk` command and Android NDK support

ManuLinares:
Update: Unified SDK fetching and standardize download progress

- Consolidated Fetching, unified all SDK fetching under `c3c fetch-sdk <target>` (supports `windows`, `android` and `macos` unimplemented yet).
- Android NDK Support: Implemented auto-download of Android NDK (r29) for cross-compilation. (I'll add more versions later)
- Unified the download UI with an interactive block progress bar into http.c for all download paths (`fetch-sdk`, `vendor-fetch`, and project dependencies). Falls back to showing the % every 10% for non-ansi or non-tty.

- `fetch-msvc` gives a deprecation warning

---

NOTE: my macos sdk *auto-downloader* does many MANY things to work, I don't think I'll keep currently working on that for now but I gained much knowledge in the process. (It has almost zero dependencies, all C code of course). We should just consider adding a simple downloader from https://github.com/joseluisq/macosx-sdks (or a fork) with a license agreement but I'm not a lawyer. In my non-lawyer opinion and apple experience, they don't care and we gain some UX and consistency for all major platforms.

---

You can test the android-ndk auto-fetching with
```bash
c3c compile --target android-aarch64 hello_world.c3
```

### DIFF
--- a/releasenotes.md
+++ b/releasenotes.md
@@ -11,6 +11,7 @@
 - `@weaklink` for just affecting linkage.
 - Add a fully static build of `c3c` for Linux. #2949
 - `@weak` now allows direct overriding of `@weak` definitions with a real definition.
+- Unified SDK fetching under `c3c fetch-sdk <target>` (windows, android) and added support for automatic Android NDK (r29) download. Better progress bar. #3019
 
 ### Stdlib changes
 - Add contract on `any_to_enum_ordinal` and `any_to_int` to improve error when passed an empty any. #2977


lerno:
This looks like it might not be compatible with 0.7.10? Keep the old command line / project options and add deprecation warnings on their use.

lerno:
I think fetch-msvc etc had the advantage of being more easily discoverable?

ManuLinares:
> I think fetch-msvc etc had the advantage of being more easily discoverable?

I'll leave the old command line with a deprecation

updated PR description.

lerno:
Thank you!

PR #3022, opened 2026-03-11
Add OS-based Secure Random Bytes (Entropy)

NotsoanoNimus:
Based on an earlier PR that was closed without a merge. Adds a way to generate (or fill) a byte array of arbitrary length with cryptographically-secure random bytes. Leverages the existing runtime platform's method for doing so.

Also adds a builtin `TIMEOUT` fault type, since these are rather ubiquitous. If the `TIMEOUT` code here seems a bit arbitrary or useless, I don't mind removing it. I wanted to provide a way for the call to these `entropy` functions to not block forever, but still have a chance to complete (_since signals can interrupt the system call_).

FYI: The Darwin Security framework includes a random-bytes function. Since I'm not a mac user, I have no idea how to incorporate linkage w/ the framework appropriately with an `extern`, or if that would be feasible to begin with.

-----
Also, if this would be **better placed** under the `lib/std/os` or `lib/std/core/os` hierarchies, I can move it there.

-----

Links: #2812 and #2808

### DIFF
--- a/lib/std/crypto/entropy.c3
+++ b/lib/std/crypto/entropy.c3
@@ -99,13 +99,13 @@ fn void? fill(char[] buffer, bool timeout = false)
 		$case env::NETBSD:
 			$if (@in(env::ARCH_TYPE, ARM, ARMB, AARCH64, X86, X86_64)):
 				isz result = $$syscall(entropy_syscall_id(), (uptr)buffer.ptr, buffer.len, 0);
-				if (result < 0 && result != errno::EINTR) return SECURE_RANDOM_UNAVAILABLE~;   /* it's possible for signal handlers to interrupt this call */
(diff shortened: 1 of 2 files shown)


ManuLinares:
I would default to `/dev/urandom` instead of `/dev/random` for the fallback to avoid blocking.

lerno:
You need to rebase this one.

NotsoanoNimus:
> You need to rebase this one.

On vacation but think it should be g2g now.

NotsoanoNimus:
I’ll fix these warnings on Monday.

lerno:
The branch is protected so I can't make any changes to this PR. For that reason I've opened #3022 containing this PR + some changes.

lerno:
So, we'll continue the discussion there.

ManuLinares:
https://github.com/c3lang/c3c/pull/3063

> The branch is protected so I can't make any changes to this PR. For that reason I've opened #3022 containing this PR + some changes.

lerno:
Ooops, yes

PR #3023, opened 2026-03-11
Bump Docker build to use LLVM 22

NotsoanoNimus:
I often use this convenient script with fresh pulls of the repository on a development VPS. The build process requires LLVM 19-23, so I figured bumping to the second-to-most-recent version would be best.


lerno:
Thank you!

PR #3027, opened 2026-03-12
Enhance Available Xo[ro]shiro PRNGs

NotsoanoNimus:
Revises the current `Xoshiro128++` implementation to also include all other published versions from the original creators. As [their main site](https://prng.di.unimi.it/) describes, many of these have different, ideal use-cases. Additionally, the `SplitMix64` pseudo-_pseudo_-RNG was added to help better seed these Xorshift PRNGs.

The set of supported Xorshift PRNGs in this family now includes the list following this comment. This long list is justified by two things:

1. The list looks extensive, but the difference is (annoyingly) only one or two statements. The implementation is less than 200 actual LOC.
2. If C3's stdlib is going to have this Xorshift family as an available PRNG type, there's no point in not having all of the published variants, lest we have many different implementations for the family as new ones are requested. <sub>(In fact, I was going to add a single one at first, then decided to complete them all instead.)</sub>

Finally, this PR makes use of private generic modules with macros that should not bubble up to any public APIs. If the amount of implementations of each module **causes any compiler slowdown**, I would revise or reconsider the design (but would obviously prefer not to).

```
Xoshiro128+
Xoshiro128++
Xoshiro128**
Xoshiro256+
Xoshiro256++
Xoshiro256**
Xoshiro512+
Xoshiro512++
Xoshiro512**
Xoroshiro64*
Xoroshiro64**
Xoroshiro128+
Xoroshiro128++
Xoroshiro128**
Xoroshiro1024*
Xoroshiro1024**
Xoroshiro1024++
```


lerno:
Thank you

Issue #3031, opened 2026-03-13
Segmentation fault in `sema_compare_weak_decl` when replacing a function declaration from a `.c3i` file


ManuLinares:
### Description
The compiler segfaults during the semantic analysis phase (`ANALYSIS_DECLS`) when a function declaration in a [.c3i](cci:7://file:///home/mb/scripts/c3c/resources/testproject/src/foo.c3i:0:0-0:0) file is replaced by a definition in a [.c3](cci:7://file:///home/mb/scripts/c3c/resources/msvc_stack.c3:0:0-0:0) file within the same module.

The crash occurs in [sema_compare_weak_decl](cci:1://file:///home/mb/scripts/c3c/src/compiler/sema_decls.c:5700:0-5730:1) because it attempts to access `function.prototype->raw_type` on a function type that has not yet had its prototype initialized.

### Minimal Reproduction
Create a project with the following structure:

**project.json**
```json
{
  "sources": ["src"],
  "targets": {
    "repro": { "type": "executable" }
  }
}
```

**src/repro.c3i**
```c3
module repro;
// This declaration is implicitly weak
fn void test();
```

**src/repro.c3**
```c3
module repro;

fn void test() {
    // Implementation
}

fn void main() {
    test();
}
```

### Steps to Reproduce
Run the compiler on the project: `c3c build`

---

### Stack Trace (GDB)
```
Program received signal SIGSEGV, Segmentation fault.
0x0000555555631c5b in sema_compare_weak_decl (context=0x7fffffff32c0, replaced=0x7ffb6aa004c8, replacement=0x7ffb6aa013a8)
    at src/compiler/sema_decls.c:5709
5709    if (replaced_type->function.prototype->raw_type != replacement_type->function.prototype->raw_type)

#0  0x0000555555631c5b in sema_compare_weak_decl (context=..., replaced=..., replacement=...) at src/compiler/sema_decls.c:5709
#1  0x00005555556a0576 in sema_analyse_weak_decls (context=..., decls=...) at src/compiler/sema_passes.c:733
#2  sema_analysis_pass_decls (module=...) at src/compiler/sema_passes.c:791
#3  0x0000555555699981 in sema_analyze_stage (module=..., stage=ANALYSIS_DECLS) at src/compiler/semantic_analyser.c:214
#4  0x0000555555699f89 in sema_analyze_to_stage (stage=ANALYSIS_DECLS) at src/compiler/semantic_analyser.c:328
[...]

PR #3033, opened 2026-03-13
Fix crash during weak declaration comparison #3031

ManuLinares:
https://github.com/c3lang/c3c/issues/3031


lerno:
I fixed the underlying problem (it should have run in a later pass), so I'll close this PR.

PR #3037, opened 2026-03-14
fetch-sdk windows: improve library casing

ManuLinares:
Expanded the MSVC SDK fetcher's library aliasing to resolve case-sensitivity issues during Linux cross-compilation. RGFW didn't link properly without these.


lerno:
👍

PR #3039, opened 2026-03-15
Fix bitset doc comments

Skunkynator:
Fixes doc comments for the bitwise operations of Bitset.

There is also: https://github.com/c3lang/c3c/blob/master/lib/std/collections/bitset.c3#L189
which has the line `defaults to the heap allocator` while there is no default value for that variable.
Should that be changed as well?


lerno:
Thank you!

PR #3042, opened 2026-03-16
fix(riscv): allow immediate constants in CSR inline assembly

### LINKED ISSUE #2628: RISC-V CSR inline asm problem
I am trying to port some C to C3, which needs to call a specific RISC-V CSR.
The C code looks like this:

	int x = 42;
	asm volatile( "csrrw x0, 0x13c, %0\n" : : "r" (x));

In C3, I have:

    int x = 42;
    asm {
        csrrw $zero, 0x13c, x;
    }

But, I am getting:

	Error: 'csrrw' does not support a direct integer constant here.

Am I using the `asm` statement wrong?

R-Camacho:
CSR instructions now accept 12-bit immediates for the CSR address. 
This fixes the issue for both linux-riscv32 and linux-riscv64 targets.

Closes #2628

### DIFF
--- a/test/test_suite/asm/asm_imm_err_rv.c3
+++ b/test/test_suite/asm/asm_imm_err_rv.c3
@@ -56,3 +56,11 @@ fn void test7()
 		slli $a0, 1, 31u; // #error: 'slli' does not support a direct integer constant here.
 	}
 }
+
+fn void test8()
+{
+	asm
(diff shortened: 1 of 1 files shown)


ManuLinares:
LGTM

lerno:
Thank you!

PR #3052, opened 2026-03-21
WIP: Add `--target emscripten`

### LINKED ISSUE #2950: Create Emscripten target
Meehai wrote on the Discord:

> My hope was to have the same code for both desktop and wasm without any changes in the C3 code. And emscripten seems to be the only thing designed for this purpose that works. In your case you seem to implement a bunch of stuff in JS whereas I hope to make a small game that will work both on desktop as well as web in pure C3 (I was hoping to port [this minipaint I did](https://gitlab.com/meehai/minipaint-raylib-multilang) to C3 as a PoC).
>
> TBH I tried a bunch of other langs and besides C (and c++) only zig which explicitly has a target=wasm32-emscripten implemented in the compiler worked fine, so I assume there needs to be some extra work somewhere at the compiler level to support this. The [repo I posted above](https://gitlab.com/meehai/c3-raylib-wasm-helloworld) has a Makefile for all my attempts, including the c3 one and as I said, it compiles but crashes at runtime.
>
> I'll try today to give it one more try to maybe understand where/why it actually crashes currently.

Emscripten should ideally made to work out of the box. Can we make that happen?

ManuLinares:
Very very much a work in progress alpha.

- Added default linker flags to `test` and `benchmark` only.

All unit tests pass:
```bash
./build/c3c compile-test -O1 --target emscripten -o test.js test/unit/ && node test.js
```

Fixes: https://github.com/c3lang/c3c/issues/2950

edit:
- Disabled unsupported POSIX process spawning/tests on Emscripten to fix `posix_spawnp` linker errors.
- Disabled set priority for threads on Emscripten 
- Added a job to the CI: "test-emscripten-32bit", it executes this file `scripts/tools/emscripten_tests.sh` which runs the unit tests and some other tests from the "resources" folder. Good for 32bit stdlib support.


ManuLinares:
I've merged the latest isz/sz change. I'm hoping to get more eyes on this.

This is still WIP until somebody reviews it or I get more time to properly test. 32bit systems like this one aren't fully stdlib tested.

PR #3053, opened 2026-03-21
- Enums now no longer directly support `+` and `-` – use ordinals ins…

lerno:
…tead.

- For enums, using `++` and `--` will step through enums with implicit wrap-around.


lerno:
Will be merged together with 0.8.0 update

PR #3056, opened 2026-03-22
Add Fedora COPR installation instructions

Sisyphus1813:
Hey! I've created a COPR repository for the compiler, so this PR adds documentation to reference that in the installation section.

For transparency, the RPM files can be found [here](https://github.com/Sisyphus1813/c3-rpm).
And the COPR repository in question is located [here](https://copr.fedorainfracloud.org/coprs/sisyphus1813/c3)

Please let me know if there is any issue or if you would like something changed. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,6 +298,15 @@ cd c3c-git
 makepkg -si
 ```
 
+#### Installing on Fedora
+
+C3 is available as a community maintained COPR package for Fedora.  
+To install via dnf:
+```sh
+sudo dnf copr enable sisyphus1813/c3
+sudo dnf install c3
+```
+
 #### Installing via Nix
 
 You can access `c3c` via [flake.nix](./flake.nix), which will contain the latest commit of the compiler. To add `c3c` to your `flake.nix`, do the following:


ManuLinares:
Hi, I don't use fedora. What's the deal with the comment on your repo:
```
Compatibility

C3 currently only builds on F43. F44 is broken due to toolchain issues (LLVM/Clang + RPM flags). Looking into this further but we are most likely waiting on upstream Fedora fixes.
```

NOTE: I have tested the c3c static build in all linux distros: https://github.com/c3lang/c3c/releases/download/latest-prerelease-tag/c3-linux-static.tar.gz

Sisyphus1813:
So essentially Fedora 44 is still in pre-release. As such, the way that fedora packages LLVM & Clang versions is completey different between Fedora 43 & Fedora 44 due to compatability concerns. That is to say, the default Clang version for Fedora 43 has always been V21. In Fedora 44, its V22, which C3 will not compile with, at least insofar as I can tell. Which is fine, because Fedora 44 still has V21 available, but through a different package name...and its stored in a completely different directory...and even once you fix all of that, because of hardening flags that Fedora passes to Clang by default, the build still fails. I could give you some of the exact error messages if you want -- it'd just take me a while to revert everything and do more mocks.

To be clear though -- if I wanted to make it work on Fedora 44 right now, it's technically doable. But given that Fedora 44 is still pre-release, and attempting to get it to build at that version has been a nightmare, I figured it's not worth my time at this moment.

ManuLinares:
> So essentially Fedora 44 is still in pre-release. ...

I've tested Fedora 44 and c3c builds fine.

`Containerfile.fedora44`
```dockerfile
FROM docker.io/library/fedora:44

# Install build dependencies
RUN dnf install -y \
    gcc gcc-c++ \
    cmake ninja-build \
    llvm-devel clang-devel lld-devel \
    zlib-devel libcurl-devel libffi-devel libzstd-devel \
    python3 bash git tar && \
    dnf clean all

WORKDIR /app
COPY . .

ENV BUILD_TYPE=Debug
ENV CMAKE_GENERATOR=Ninja

RUN cmake -B build \
    -D C3_LINK_DYNAMIC=ON \
    -D C3_FETCH_LLVM=OFF \
    -D CMAKE_BUILD_TYPE="$BUILD_TYPE" \
    -D C3_ENABLE_CLANGD_LSP=ON \
    -S . && \
    ninja -C build

CMD ["./build/c3c", "--version"]
```

```
podman build -t c3c-fedora44 -f Containerfile.fedora44 . && podman run --rm -it localhost/c3c-fedora44:latest
```

Sisyphus1813:
I'm looking more into this, I'll get back to you shortly.

Sisyphus1813:
Thanks - I dug further into this.
After some additional debugging I'm able to get it to compile on F44.
However, the RPMbuild itself still fails (during check-rpaths) with the following error message:

```
*******************************************************************************
*
* WARNING: 'check-rpaths' detected a broken RPATH OR RUNPATH and will cause
*          'rpmbuild' to fail. To ignore these errors, you can set the
*          '$QA_RPATHS' environment variable which is a bitmask allowing the
*          values below. The current value of QA_RPATHS is 0x0000.
*
*    0x0001 ... standard RPATHs (e.g. /usr/lib); such RPATHs are a minor
*               issue but are introducing redundant searchpaths without
*               providing a benefit. They can also cause errors in multilib
*               environments.
*    0x0002 ... invalid RPATHs; these are RPATHs which are neither absolute
*               nor relative filenames and can therefore be a SECURITY risk
*    0x0004 ... insecure RPATHs; these are relative RPATHs which are a
*               SECURITY risk
*    0x0008 ... the special '$ORIGIN' RPATHs are appearing after other
*               RPATHs; this is just a minor issue but usually unwanted
*    0x0010 ... the RPATH is empty; there is no reason for such RPATHs
*               and they cause unneeded work while loading libraries
*    0x0020 ... an RPATH references '..' of an absolute path; this will break
*               the functionality when the path before '..' is a symlink
*
*
* Examples:
* - to ignore standard and empty RPATHs, execute 'rpmbuild' like
*   $ QA_RPATHS=$(( 0x0001|0x0010 )) rpmbuild my-package.src.rpm
* - to check existing files, set $RPM_BUILD_ROOT and execute check-rpaths like
*   $ RPM_BUILD_ROOT=<top-dir> /usr/lib/rpm/check-rpaths
*
*******************************************************************************
ERROR   0010: file '/usr/bin/c3c' contains an empty runpath in [/usr/lib64/llvm21/lib64:]
```

Again I have not had such any issues while building on F43, so this does seem to be F44 specific.

ManuLinares:
> Again I have not had such any issues while building on F43, so this does seem to be F44 specific.

Could you try patching the CMakeLists.txt?

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index f1eb82f52..a43c05297 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,7 @@ endif()
 
 # Set the project and version
 project(c3c VERSION ${CMAKE_MATCH_1} LANGUAGES C CXX)
+set(CMAKE_SKIP_RPATH ON)
 message("Configuring C3C ${CMAKE_PROJECT_VERSION} for ${CMAKE_SYSTEM_NAME}")
 
 # Helper functions
@@ -280,7 +281,7 @@ if(C3_WITH_LLVM)
     list(GET VERSION_LIST 0 LLVM_MAJOR_VERSION)
 
     include_directories(${LLVM_INCLUDE_DIRS})
-    link_directories(${LLVM_LIBRARY_DIRS})
+    # link_directories(${LLVM_LIBRARY_DIRS})
     separate_arguments(LLVM_DEFINITIONS_LIST NATIVE_COMMAND ${LLVM_DEFINITIONS})
     add_definitions(${LLVM_DEFINITIONS_LIST})
```

ManuLinares:
also try this diff separately please:
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index f1eb82f52..7d7b2d5a8 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -260,6 +260,7 @@ if(C3_WITH_LLVM)
     endif()
 
     list(REMOVE_DUPLICATES LLVM_LIBRARY_DIRS)
+    list(FILTER LLVM_LIBRARY_DIRS EXCLUDE REGEX "^$")
 
     message(STATUS "Found LLVM ${LLVM_PACKAGE_VERSION}")
     message(STATUS "Using LLVMConfig.cmake in: ${LLVM_DIR}")
@@ -332,6 +333,7 @@ if(C3_WITH_LLVM)
         if(NOT ${C3_LLD_DIR} EQUAL "" AND EXISTS ${C3_LLD_DIR})
             list(APPEND LLVM_LIBRARY_DIRS ${C3_LLD_DIR})
             list(REMOVE_DUPLICATES LLVM_LIBRARY_DIRS)
+            list(FILTER LLVM_LIBRARY_DIRS EXCLUDE REGEX "^$")
         endif()
 
         message(STATUS "Looking for static lld libraries in ${LLVM_LIBRARY_DIRS}")
```

Sisyphus1813:
So the first diff failed to run at all (because I'm building against the tagged release instead of master branch). And the second diff did run but the rpmbuild failed again with the same error message. However, I tweaked the first diff slightly:

```
diff --git a/CMakeLists.txt b/CMakeLists.txt
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,7 @@ endif()

 # Set the project and version
 project(c3c VERSION ${CMAKE_MATCH_1} LANGUAGES C CXX)
+set(CMAKE_SKIP_RPATH ON)
 message("Configuring C3C ${CMAKE_PROJECT_VERSION} for ${CMAKE_SYSTEM_NAME}")

 # Helper functions
@@ -280,7 +281,7 @@
     list(GET VERSION_LIST 0 LLVM_MAJOR_VERSION)

     include_directories(${LLVM_INCLUDE_DIRS})
-    link_directories(${LLVM_LIBRARY_DIRS})
+    # link_directories(${LLVM_LIBRARY_DIRS})
     add_definitions(${LLVM_DEFINITIONS})

     if(NOT C3_LINK_DYNAMIC)
```

And the build now succeeds with no issues on both F43 & F44. I'll go ahead and get the COPR updated. Thanks for your help!

ManuLinares:
> And the build now succeeds with no issues on both F43 & F44. I'll go ahead and get the COPR updated. Thanks for your help!

For fedora might be ok but I don't like disabling RPATH. It's very useful to have those so the second diff I provided should be better (sanitizing RPATH to exclude empty ones).

Although I'm testing c3c master, not 0.7.10

Sisyphus1813:
Out of curiosity, I tried building from the latest commit instead of the tagged release. The issue still occurs on F44 if I exclude the patch. I also tried applying your second diff against the latest commit, but that failed with the same rpmbuild error.

For now, I'll go ahead and set the patch to only apply to F44 builds. Of course that's still a temporary work around but.

ManuLinares:
this one **should** work on the latest c3c master

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index f1eb82f52..7d7b2d5a8 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -260,6 +260,7 @@ if(C3_WITH_LLVM)
     endif()
 
     list(REMOVE_DUPLICATES LLVM_LIBRARY_DIRS)
+    list(FILTER LLVM_LIBRARY_DIRS EXCLUDE REGEX "^$")
 
     message(STATUS "Found LLVM ${LLVM_PACKAGE_VERSION}")
     message(STATUS "Using LLVMConfig.cmake in: ${LLVM_DIR}")
@@ -332,6 +333,7 @@ if(C3_WITH_LLVM)
         if(NOT ${C3_LLD_DIR} EQUAL "" AND EXISTS ${C3_LLD_DIR})
             list(APPEND LLVM_LIBRARY_DIRS ${C3_LLD_DIR})
             list(REMOVE_DUPLICATES LLVM_LIBRARY_DIRS)
+            list(FILTER LLVM_LIBRARY_DIRS EXCLUDE REGEX "^$")
         endif()
 
         message(STATUS "Looking for static lld libraries in ${LLVM_LIBRARY_DIRS}")
```

Sisyphus1813:
Yup the diff itself does work just fine --  I just still get the same rpmbuild error.

ManuLinares:
> Yup the diff itself does work just fine -- I just still get the same rpmbuild error.

Mmm, from my testing it builds the RPM fine. I should ask you, if you have the time, to find a way to not `set(CMAKE_SKIP_RPATH ON)` for the rpm to build on fedora 44 but honestly the next C3 version will ship a linux-static which will run on every linux platform, so this build mess is something I am not particularly worried.

lerno:
Anything else to add or are we good with approving this @ManuLinares

ManuLinares:
> Anything else to add or are we good with approving this @ManuLinares

We're good.

Sisyphus1813:
Thanks guys!

PR #3057, opened 2026-03-24
FreeBSD fixes and sockets support

siladi:
Compiles and passes all tests


siladi:
> We don't have FreeBSD up and running in CI, so I can't check this fully, but I'll include it and we can hopefully get some CI later.

Working on geting CI for FreeBSD setup and working.  I have reached out on discord for feedback/guidance so I think have a good idea how to make it happen soon.

Issue #1368, opened 2024-08-18
Segmentation fault on library project build when exporting headers


ndelta0:
I tried building a template project (`c3c --template static-lib init <name>`) with only one export, doing that results in a segmentation fault (along with a dump).

From what information I gathered it only occurs when building a project (specifically a library which has enabled header exporting) and not doing `c3c static-lib <file>` and it is related to an export with no explicit name (i.e. `fn ... () @export` and not `fn ... () @export("export_name")`), which the documentation clearly states as a valid thing to do.

I've had success with adding a null check (`if (decl->extname && ...)`) in this if condition:
https://github.com/c3lang/c3c/blob/20b0bf43ada793d4d60e78493f9dd069393bfcb1/src/compiler/headers.c#L278

After that the export is correctly named and the project compiles.

I attach a link to the core dump. It was tested on the commit to which the aforementioned code line points to (which is the head of the dev branch at the time of writing). Be warned that the core dump extracts to a 4.4GB file.
[Google Drive](https://drive.google.com/file/d/1upYHX2R7VG-55LkXJ3s7lyqRaGJ9Vuku/view?usp=sharing)

```shell
$ c3c -V
C3 Compiler Version:       0.6.2 (Pre-release, Aug 18 2024 16:57:38)
Installed directory:       /usr/local/bin/
LLVM version:              18.1.8
LLVM default target:       x86_64-pc-linux-gnu

$ uname -a
Linux papaya 6.8.8-arch1-1 #1 SMP PREEMPT_DYNAMIC Sun, 28 Apr 2024 15:59:47 +0000 x86_64 GNU/Linux

$ lsb_release -a
LSB Version:	n/a
Distributor ID:	Arch
Description:	Arch Linux
Release:	rolling
Codename:	n/a
```

lerno:
Use --no-headers for now. The headers are still not really 100% sure to be correct I think.

PR #3058, opened 2026-03-24
Fix segmentation fault when exporting headers with no explicit name

nmurrell07:
This PR fixes issue #1368 - Segmentation fault on library project build when exporting headers.

The problem occurs when building a template project with an export that has no explicit name (i.e.,  without providing a name). The code was accessing  and  without first checking if  is NULL, causing a segmentation fault.

The fix adds a null check before accessing the array elements:

This ensures that if  returns NULL for any reason, we safely return early instead of dereferencing a null pointer.


lerno:
The headers need a complete overhaul, but thank you for the fix, I'll include it for now.

PR #3067, opened 2026-03-26
Add OrderedMap stdlib module (skip list implementation)

ChristianReifberger:
Adds a sorted map collection backed by a probabilistic skip list, filling the gap between `HashMap` (fast, unordered) and the sort module (one-shot sorting). An `OrderedMap` maintains keys in sorted order at all times with O(log n) insert, lookup, and removal.

## Types
```c3
OrderedMap{Key, Value}      // ascending order (default)
OrderedMapDesc{Key, Value}  // descending order
```

Custom comparator variant (an option to replace built-in ordering/compare function):
```c3
PrivateOrderedMap{Key, Value, false}.tinit_with_cmp(&my_cmp_fn)
```

## API
| Method | Description |
|--------|-------------|
| `set(key, value)` | Insert or update |
| `get(key)` | Lookup → `Value?` |
| `get_ref(key)` | Lookup → `Value*` for in-place mutation |
| `get_or_set(key, default)` | Return existing or insert default |
| `has_key(key)` | Existence check |
| `remove(key)` | Remove by key |
| `first_key()` / `last_key()` | Min/max key in iteration order |
| `floor_key(key)` / `ceiling_key(key)` | Nearest key ≤ / ≥ target |
| `lower_key(key)` / `higher_key(key)` | Nearest key strictly < / > target |
| `pop_first()` / `pop_last()` | Remove and return first/last entry |
| `clone(allocator)` / `tclone()` | Deep copy |
| `merge(other)` | Insert all entries from another map |
| `iter()` | Full iteration in sorted order |
| `iter_from(key)` | Iterate from ceiling of key to end |
| `iter_range(from, to)` | Iterate `[from, to)` in cmp order |

## Implementation notes

- Two modules in one file: `orderedmap_node<Key, Value>` holds the self-referential `SkipNode` struct (required for the forward pointer array to compile); `orderedmap` holds the rest.
- `OrderedMap` / `OrderedMapDesc` are `inline typedef` aliases over `PrivateOrderedMap{Key, Value, DESCENDING}`, following the `PriorityQueue` pattern.
- Uses `std::math::random::SimpleRandom` seeded via `seed_entropy` for level randomisation.
- Works with any key type supported by the `less()` builtin (integers, floats, chars); custom comparators lift this restriction.

## Tests

`test/test_suite/stdlib/orderedmap.c3t` covers all methods with multiple key/value type combinations (`int`, `float`, `long`, `char` keys; `String`, `int`, `bool`, `double` values) and both ascending and descending ordering.

### DIFF
--- a/lib/std/collections/orderedmap.c3
+++ b/lib/std/collections/orderedmap.c3
@@ -49,12 +49,12 @@ struct Entry
 <*
  @require types::@comparable_value((Key){}) : "The Key must support ordered comparison"
 *>
-struct OrderedMap (Printable)
+struct OrderedMap (Printable) @mustinit
 {
 	Allocator allocator;
 	SkipNode* head;
 	int current_level;
-	usz count;
+	sz count;
 	SimpleRandom rng;
 	CmpFn custom_cmp; // null = use built-in less() ordering
 }
@@ -65,7 +65,7 @@ struct OrderedMapIterator
 {
 	OrderedMap* map;
 	SkipNode* current;
-	usz index;
+	sz index;
 }
 
 struct OrderedMapRangeIterator
@@ -73,8 +73,8 @@ struct OrderedMapRangeIterator
 	OrderedMap* map;
 	SkipNode* start_node;
 	SkipNode* current;
-	usz current_idx;
-	usz count;
+	sz current_idx;
+	sz count;
 }
 
 fn int orderedmap_random_level(SimpleRandom* rng) @private
@@ -155,7 +155,7 @@ fn bool OrderedMap.set(&self, Key key, Value value) @operator([]=)
 
 	for (int i = self.current_level - 1; i >= 0; i--)
 	{
-		while (current.forward[i] != null && orderedmap_cmp(self.custom_cmp, current.forward[i].key, key))
+		while (current.forward[i] && orderedmap_cmp(self.custom_cmp, current.forward[i].key, key))
 		{
 			current = current.forward[i];
 		}
@@ -165,7 +165,7 @@ fn bool OrderedMap.set(&self, Key key, Value value) @operator([]=)
 	current = current.forward[0];
 
 	// Key already exists -- update value
-	if (current != null && !orderedmap_cmp(self.custom_cmp, current.key, key) && !orderedmap_cmp(self.custom_cmp, key, current.key))
+	if (current && !orderedmap_cmp(self.custom_cmp, current.key, key) && !orderedmap_cmp(self.custom_cmp, key, current.key))
 	{
 		current.value = value;
 		return true;
@@ -200,13 +200,13 @@ fn Value? OrderedMap.get(&self, Key key) @operator([])
 	SkipNode* current = self.head;
 	for (int i = self.current_level - 1; i >= 0; i--)
 	{
-		while (current.forward[i] != null && orderedmap_cmp(self.custom_cmp, current.forward[i].key, key))
+		while (current.forward[i] && orderedmap_cmp(self.custom_cmp, current.forward[i].key, key))
 		{
 			current = current.forward[i];
 		}
 	}
 	current = current.forward[0];
-	if (current != null && !orderedmap_cmp(self.custom_cmp, current.key, key) && !orderedmap_cmp(self.custom_cmp, key, current.key))
+	if (current && !orderedmap_cmp(self.custom_cmp, current.key, key) && !orderedmap_cmp(self.custom_cmp, key, current.key))
 	{
 		return current.value;
 	}
@@ -221,13 +221,13 @@ fn Value*? OrderedMap.get_ref(&self, Key key) @operator(&[])
 	SkipNode* current = self.head;
 	for (int i = self.current_level - 1; i >= 0; i--)
 	{
-		while (current.forward[i] != null && orderedmap_cmp(self.custom_cmp, current.forward[i].key, key))
+		while (current.forward[i] && orderedmap_cmp(self.custom_cmp, current.forward[i].key, key))
 		{
 			current = current.forward[i];
 		}
 	}
 	current = current.forward[0];
-	if (current != null && !orderedmap_cmp(self.custom_cmp, current.key, key) && !orderedmap_cmp(self.custom_cmp, key, current.key))
+	if (current && !orderedmap_cmp(self.custom_cmp, current.key, key) && !orderedmap_cmp(self.custom_cmp, key, current.key))
 	{
 		return &current.value;
 	}
@@ -263,7 +263,7 @@ fn bool OrderedMap.remove(&self, Key key)
 
 	for (int i = self.current_level - 1; i >= 0; i--)
 	{
-		while (current.forward[i] != null && orderedmap_cmp(self.custom_cmp, current.forward[i].key, key))
+		while (current.forward[i] && orderedmap_cmp(self.custom_cmp, current.forward[i].key, key))
 		{
 			current = current.forward[i];
 		}
@@ -272,7 +272,7 @@ fn bool OrderedMap.remove(&self, Key key)
 
 	current = current.forward[0];
 
-	if (current == null || orderedmap_cmp(self.custom_cmp, current.key, key) || orderedmap_cmp(self.custom_cmp, key, current.key)) return false;
+	if (!current || orderedmap_cmp(self.custom_cmp, current.key, key) || orderedmap_cmp(self.custom_cmp, key, current.key)) return false;
 
 	for (int i = 0; i < self.current_level; i++)
 	{
@@ -282,7 +282,7 @@ fn bool OrderedMap.remove(&self, Key key)
 
 	alloc::free(self.allocator, current);
 
-	while (self.current_level > 0 && self.head.forward[self.current_level - 1] == null)
+	while (self.current_level > 0 && !self.head.forward[self.current_level - 1])
 	{
 		self.current_level--;
 	}
@@ -317,6 +317,7 @@ fn Key? OrderedMap.last_key(&self)
 <*
  Remove and return the first entry in iteration order.
  @return "the first Entry, or NOT_FOUND if empty"
+ @return? NOT_FOUND
 *>
 fn Entry? OrderedMap.pop_first(&self)
 {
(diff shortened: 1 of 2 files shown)


lib/std/collections/orderedmap.c3
```diff
@@ -0,0 +1,560 @@
...
+// SkipNode is in its own module-level generic module so that the
+// self-referential forward pointer array compiles correctly.
+module std::collections::orderedmap_node <Key, Value>;
+
+struct SkipNode
+{
+	Key key;
+	Value value;
+	int level;
+	SkipNode*[32] forward; // 32 = orderedmap::MAX_LEVEL
```
Book-reader: This should use the `MAX_LEVEL` constant directly if they're meant to be equal

lib/std/collections/orderedmap.c3
```diff
@@ -0,0 +1,560 @@
...
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+//  SOFTWARE.
+
+// SkipNode is in its own module-level generic module so that the
+// self-referential forward pointer array compiles correctly.
```
Book-reader: this doesn't seem right, moving `struct SkipNode {...}` into `std::collections::orderedmap` as `struct SkipNode <Key, Value> {...}`  works fine for me (on c3c 0.7.11 pre-release commit `c1985c08459dc595e98326addb8754115247e9df` at least)

ChristianReifberger:
@Book-reader Thanks for the feedback, I implemented all of them; seems like i got something wrong in the github workflow because the commit from lerno is also included? (my first time using a PR)

ChristianReifberger:
fixed the commits so lernos commit doesnt show up in the pr

lib/std/collections/orderedmap.c3
```diff
@@ -0,0 +1,554 @@
...
+	{
+		return current.value;
+	}
+	return NOT_FOUND~;
+}
+
+<*
+ @return "a pointer to the value for key, or null if absent"
+*>
+fn Value* PrivateOrderedMap.get_ref(&self, Key key) @operator(&[])
```
Book-reader:
```suggestion
fn Value*? PrivateOrderedMap.get_ref(&self, Key key) @operator(&[])
```
this should return the fault `NOT_FOUND~` instead of `null` when a key isn't found to enforce checking it in the type system.

lib/std/collections/orderedmap.c3
```diff
@@ -0,0 +1,554 @@
...
+	{
+		return &current.value;
+	}
+	return null;
+}
+
+<*
+ Returns the value for key; inserts and returns default_value if key is absent.
+*>
+fn Value PrivateOrderedMap.get_or_set(&self, Key key, Value default_value)
```
Book-reader: It could be useful to make this a macro with a lazily evaluated expression parameter for the default value like the `HashMap` has ([lib/std/collections/hashmap.c3#L221](https://github.com/c3lang/c3c/blob/f44b5f4b41fe0cab7d814bdabe804ff8bb55e57c/lib/std/collections/hashmap.c3#L221)) so the default value only gets evaluated if the key doesn't exist:
```c3
// `some_expensive_function()` will only be called if `key` doesn't exist
map.@get_or_set(key, some_expensive_function());
```
It would look something like this (the `@` in the name is required for macros with `#expr` parameters):
```c3
macro Value PrivateOrderedMap.@get_or_set(&self, Key key, Value #default_value)
{
	if (try v = self.get(key)) return v;
    Value val = #default_value;
	self.set(key, val);
	return val;
}
```

lib/std/collections/orderedmap.c3
```diff
@@ -0,0 +1,554 @@
...
+
+fn int orderedmap_random_level(SimpleRandom* rng) @private
+{
+	int level = 1;
+	while (level < MAX_LEVEL && (rng.next_int() & 1) == 0) level++;
+	return level;
+}
+
+// Returns true if 'a' should come before 'b' in the list.
+macro bool PrivateOrderedMap.cmp(&self, Key a, Key b) @private
```
Book-reader: `@private` no longer affects methods, This is a warning for this but it seems like it doesn't work on macros (https://github.com/c3lang/c3c/issues/3071). Making this a function reveals the warning:

Image

If it should be private it needs to be made a standalone function instead of being a method on `PrivateOrderedMap`.

lib/std/collections/orderedmap.c3
```diff
@@ -0,0 +1,554 @@
...
+	$else
+		return less(a, b);
+	$endif
+}
+
+fn PrivateOrderedMap* PrivateOrderedMap.init(&self, Allocator allocator)
+{
+	*self = { .allocator = allocator };
+	self.head = alloc::alloc(allocator, SkipNode{Key, Value});
+	*self.head = {};
```
Book-reader:
```suggestion
	self.head = alloc::new(allocator, SkipNode{Key, Value});
```
`alloc::new` will do the same as `alloc::alloc` and manually zeroing

lib/std/collections/orderedmap.c3
```diff
@@ -0,0 +1,554 @@
...
+	Value value;
+	int level;
+	SkipNode*[MAX_LEVEL] forward;
+}
+
+// CmpFn is a comparator: returns true if a should come before b in the list.
+alias CmpFn <Key> = fn bool(Key, Key);
+
+typedef OrderedMap     <Key, Value> = inline PrivateOrderedMap{Key, Value, false};
+typedef OrderedMapDesc <Key, Value> = inline PrivateOrderedMap{Key, Value, true};
```
Book-reader: these should just be aliases, there isn't any reason to make them distinct types and they lose the operator overloads.

test/test_suite/stdlib/orderedmap.c3t
Book-reader, line 1: the number of tests are good, but they aren't in the right place. The tests should be placed in `test/unit/stdlib/collections/`, have the `@test` attribute, and assert that the values are as expected instead of printing them.
[test/unit/stdlib/collections/list.c3](https://github.com/c3lang/c3c/blob/master/test/unit/stdlib/collections/list.c3) and [test/unit/stdlib/collections/linked_map.c3](https://github.com/c3lang/c3c/blob/master/test/unit/stdlib/collections/linked_map.c3) are good examples of this (in these the `@test` attribute is added to the module declaration so it applies to all functions in that section)

lib/std/collections/orderedmap.c3
```diff
@@ -0,0 +1,554 @@
...
+
+fn PrivateOrderedMap* PrivateOrderedMap.tinit_with_cmp(&self, CmpFn{Key} cmp_fn)
+{
+	return self.init_with_cmp(tmem, cmp_fn) @inline;
+}
+
+<*
+ Insert or update the value associated with key.
+*>
+fn bool PrivateOrderedMap.set(&self, Key key, Value value) @operator([]=)
```
Book-reader: the stdlib convention is to automatically initialise containers with the temporary allocator when used without calling their `init` method:
https://github.com/c3lang/c3c/blob/def0d20052457dfa1ba0ac3ac14ca61bf85a765e/lib/std/collections/hashmap.c3#L250-L261

the `case &dummy` is to support using `HashMap {int, String} x = map::ONHEAP` to make a hashmap that will automatically initialise to use the heap allocator when used without calling the `init` method.

lib/std/collections/orderedmap.c3
ChristianReifberger: Fixed

test/test_suite/stdlib/orderedmap.c3t
ChristianReifberger: I'll move it over, should i remove the .c3t test file or keep it?

ChristianReifberger:
I think I addressed all your comments; one note: i had to remove the Desc, I couldn't get it to work with module-level generics, which I think are necessary for the automatic initialization like in Hashmap (&dummy). That shouldnt be a problem though, as there might only be a few use cases where that actually maters, people can always just go from reverse or provide a custom cmp fn that inverts less(). Lastly I also added the Printable interface.

test/test_suite/stdlib/orderedmap.c3t
Book-reader: It should probably be removed. I'm honestly not sure why some stdlib tests are in the `test_suite` folder, afaict the `test_suite` folder is meant to test the functionality of the compiler and not the stdlib.

ChristianReifberger: its removed

lib/std/collections/orderedmap.c3
```diff
@@ -0,0 +1,597 @@
...
+	Value v = self.get(k)!!;
+	self.remove(k);
+	return { k, v };
+}
+
+// --- Floor / ceiling / lower / higher ---
+//
+// All four operate in the map's iteration (cmp) order.
+// For ascending maps this matches natural key ordering.
+// For descending maps, "before" means "greater" in natural ordering.
```
Book-reader: since descending maps aren't a separate type now this comment should probably be updated

lib/std/collections/orderedmap.c3
```diff
@@ -0,0 +1,597 @@
...
+	SkipNode* n = start;
+	while (n != null) { count++; n = n.forward[0]; }
+
+	return { .map = self, .start_node = start, .current = start, .current_idx = 0, .count = count };
+}
+
+<*
+ Returns an iterator over keys in [from_key, to_key) in cmp order.
+ For ascending maps: from_key <= key < to_key.
+ For descending maps: pass from and to in descending order (from > to naturally).
```
Book-reader: same here

ChristianReifberger: done

lib/std/collections/orderedmap.c3
ChristianReifberger: done

lerno:
Thank you!

PR #3069, opened 2026-03-26
Replace matrix_old with the actual original matrix types

TechnicalFowl:
In coming back to an old project I found that even after replacing my matrix types and functions with the corresponding matrix_old types and functions things did not work right. This is because somehow in the creation of matrix_old several functions were replaced with transposed variants (adjoint, apply, inverse, and perspective I believe). This change is a direct copy of the matrix.c3 version from the last tag, reordered to match the (really quite strange) ordering of matrix_old.c3. I know that matrix_old is a temporary bridge but this should help translating to the newer stuff if the old still works.

### DIFF
--- a/lib/std/math/matrix_old.c3
+++ b/lib/std/math/matrix_old.c3
@@ -1,6 +1,19 @@
+module std::math;
+
+// Predefined matrix types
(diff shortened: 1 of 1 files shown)


lerno:
Thank you!

PR #3072, opened 2026-03-27
Add std::encoding::ini module

ChristianReifberger:
Adds a new INI parsing and encoding module to the stdlib with:
- parse()/tparse() and encode()/tencode()
- Mutations: set, remove_key, remove_section, clear, free
- Iteration: iter(), @each_key, @each (section, key, value)
- IniOptions: delimiter, padding, inline comments, duplicate-key policy
- GLOBAL section for pre-header keys
- tests covering parse, encode, round-trip, edge cases

**Types**
- `IniMap` — the parsed/constructed map; owns all allocated strings
- `IniOptions` — parser and encoder configuration; pass by value starting from `ini::DEFAULT_OPTIONS`
- `DuplicateKeyBehavior` — enum: `DUPLICATE_OVERWRITE` (default) or `DUPLICATE_ERROR`
- `const String GLOBAL = ""` — sentinel for keys before the first section header
- `faultdef INVALID_SECTION, INVALID_KEY, DUPLICATE_KEY`

## API
Method | Description
-- | --
ini::parse(allocator, src, options) | Parse INI text, heap-allocated; returns IniMap?
ini::tparse(src, options) | Parse INI text, temp-allocated; returns IniMap?
init(allocator) | Create empty map for programmatic construction
tinit() | init using temp allocator
set(section, key, value) | Insert or overwrite a key; creates section if absent
remove_key(section, key) → bool | Remove a single key; returns true if it existed
remove_section(section) → bool | Remove a section and all its keys; GLOBAL cannot be removed
get(section, key) → String? | Look up a value, or NOT_FOUND
get_section(section) → HashMap*? | Return the inner map for a section
has_section(section) → bool | True if section exists
has_key(section, key) → bool | True if key exists in section
len() → usz | Number of sections; empty GLOBAL excluded
section_len(section) → usz | Number of keys in a section, or 0 if absent
iter() → String[] | Temp-allocated slice of section names; empty GLOBAL excluded
@each_key(section; key, value) | Iterate key/value pairs within one section
@each(section, key, value) | Iterate every entry; section may be ini::GLOBAL
encode(allocator, options) → String | Serialize to INI text; global keys emitted first
tencode(options) → String | encode using temp allocator
clear() | Remove all entries; GLOBAL section kept empty
free() | Free all owned memory; safe to call after a parse error

## Implementation notes
- Single file lib/std/encoding/ini.c3. 
- Section names, keys, and values are all heap-copied into the provided allocator; ini::GLOBAL ("") is the one exception — it is a compile-time constant and never heap-allocated. 
- The outer map is a HashMap{String, HashMap{String, String}}; each section is its own inner HashMap. 
- _ensure_section copies the section name on first insertion only, so re-entering the same section is allocation-free. 
- encode() writes global keys first, then named sections, using a temp-backed ByteWriter and copying the result into the target allocator at the end.
-  @each in clear()/free() yields by-value copies that share backing pointers — safe because HashMap.clear()/free() never dereferences stored value internals.

## Tests
`test/unit/stdlib/ini.c3t` covers all cases: basic parse, global keys, comments (; and #), whitespace trimming, empty values, colon delimiter, duplicate key policies (DUPLICATE_OVERWRITE / DUPLICATE_ERROR), error cases (INVALID_SECTION, INVALID_KEY), programmatic construction via init/set, mutation (remove_key, remove_section, clear), iteration (iter(), \@each_key/each including GLOBAL visibility), round-trip parse → encode → parse, inline comments, CRLF line endings, and serialization options (pad_delimiter, custom delimiter).

### DIFF
--- a/lib/std/core/dstring.c3
+++ b/lib/std/core/dstring.c3
@@ -325,16 +325,9 @@ fn Char32[] DString.copy_utf32(&self, Allocator allocator)
 	return self.str_view().to_utf32(allocator) @inline!!;
 }
 
-<*
- @require $defined(String s = str) ||| $typeof(str) == DString : "Expected string or DString"
-*>
-macro void DString.append_string(&self, str)
+fn void DString.append_string(&self, String str)
 {
-	$if $typeof(str) == DString:
-		self.append_string_deprecated(str);
-	$else
-		self.append_bytes((String)str);
-	$endif
+	self.append_bytes((String)str);
 }
 
 fn void DString.append_dstring(&self, DString str)
--- a/lib/std/encoding/ini.c3
+++ b/lib/std/encoding/ini.c3
@@ -9,40 +9,18 @@ faultdef INVALID_SECTION, INVALID_KEY, DUPLICATE_KEY;
 
 enum DuplicateKeyBehavior
 {
-    DUPLICATE_OVERWRITE,
-    DUPLICATE_ERROR,
+	OVERWRITE,
+	ERROR,
 }
 
-struct IniOptions
-{
-    bool allow_global_keys;              // Keys before the first section (default: true)
-    bool allow_hash_comments;            // Treat # as a comment char (default: true)
-    bool allow_inline_comments;          // Strip trailing inline comments from values (default: false).
-                                         // A comment is recognised as " ;" (or " #" when allow_hash_comments
-                                         // is true) — the leading space prevents false positives on values
-                                         // like "a=v;data" while still handling "a=v ; comment" correctly.
-    char delimiter;                      // Key/value separator: '=' or ':' (default: '=')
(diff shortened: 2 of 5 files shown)


lerno:
The IniMap is duplicating what Object already does. Reuse `Object` instead. This would remove everything except the parse and encode functionality.

ChristianReifbergerVUS:
@lerno thanks for the feedback, I've changed it to use Object. I did not look into it but you're right, much of the code could be removed

lerno:
Also, can you make sure that this submission conforms to the rules for contributing here: https://github.com/c3lang/c3c/blob/master/CONTRIBUTING.md

lerno:
I ended up rewriting this one significantly because the encodings should really properly be stream first. Unfortunately this is not the case for some of the older code, but new should do so.

PR #3075, opened 2026-03-28
fix(build): fix directory checking in project-based commands

### LINKED ISSUE #2980: `-L` flag checks directories relative to current working directly but passed to the linker starting at project root
Create empty project and in the root run `c3c build -L build` and then change into `build` and run the same command, which will fail. Now in `build` create empty dir `build` and run the command with `-vv`, the linker is executed from the project root.

The directories should be checked against the root project dir instead.

Loptom:
- Introduced a BuildOptions field to indicate commands that need deferring
- Moved project based commands dir checking to after the cwd is set as project root

closes #2980

### DIFF
--- a/src/build/build.h
+++ b/src/build/build.h
@@ -213,7 +213,6 @@ typedef struct BuildOptions_
 	bool test_noleak;
 	bool test_show_output;
 	bool benchmark_csv_report;
-	bool defer_dir_checking;
 	bool print_large_functions;
 	const char *custom_linker_path;
 	uint32_t symtab_size;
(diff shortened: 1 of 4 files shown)


lerno:
You can place this setting in the CompilerState instead:
```c
typedef struct
{
	BuildTarget build;
	PlatformTarget platform;
	Linking linking;
	GlobalContext context;
	const char *obj_output;
	int generic_depth;
	double exec_time;
	double script_time;
} CompilerState;
```

Loptom:
@lerno Hi, thanks for review! Should I just include  `compiler_internal.h` in `build_options.c` to set the field, or move that command filtering to somewhere else, like `compiler.c`?

lerno:
Thank you!

PR #3079, opened 2026-03-29
Add messages to the deprecated termios methods & constants that point to the replacements

Book-reader:
I probably should've done this in the original PR. These might get removed in 0.8.0, so if anyone is still using these (which tbf is pretty unlikely) the new message will point them to the replacements.


lerno:
👍

PR #3094, opened 2026-04-01
Add `@mustinit` Attribute

NotsoanoNimus:
Adds a simple `@mustinit` to explicitly disallow `@noinit` on declarations of a type. The purpose of this attribute has been discussed in Discord, but distills down to _enforcing_ a zero-initialization of a type when that cannot be safely assumed by the type.

```c3
module testrun;
import std::io;
struct Somethingggg @mustinit
{
	int a;
}
fn void main()
{
	Somethingggg h @noinit;   // <--- should throw a compiler error disallowing this
	io::printfn("hello");
}

### DIFF
--- a/releasenotes.md
+++ b/releasenotes.md
@@ -21,11 +21,11 @@
 - Removed deprecated `$assignable`.
 - Enums now no longer directly support `+` and `-` – use ordinals instead.
 - For enums, using `++` and `--` will step through enums with implicit wrap-around.
-- Add a `@mustinit` attribute to enforce zero-initialization of a type. #3094
 - Rename `isz` -> `sz`.
 - Make $sizeof, $alignof and all similar functions return `sz` instead of `usz`.
 - Align literal types with C semantics.
 - Use value promotion instead of signedness promotion to int. So that small unsigned types promote to int, not uint.
+- Add a `@mustinit` attribute to enforce zero-initialization of a type. #3094
 
 ### Stdlib changes
 - `std::collections::RingBuffer` is renamed `RingList`.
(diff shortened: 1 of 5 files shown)


Ecoral360:
I like this feature, but I would name it `@mustinit` to better reflect the intent.

LowByteFox:
Interesting idea, I wonder if and how you want to handle any other arbitrary allocation, e.g. through malloc or other means such as pre-allocating a buffer with `@noinit` and casting.

NotsoanoNimus:
> I like this feature, but I would name it `@mustinit` to better reflect the intent.

The original intent was to _implicitly_ initialize the structure/variable (which would fit for `always`, as it happens behind-the-scenes), but since it has morphed into something which throws a compiler error instead, I do think `must` is a better fit. Great suggestion!

> Interesting idea, I wonder if and how you want to handle any other arbitrary allocation, e.g. through malloc or other means such as pre-allocating a buffer with `@noinit` and casting.

Ah, very interesting to consider "hacking" this system with a cast. I suppose the attribute should be presented, here and in the documentation, as a "guard-rail" for an API, but not something which is going to enforce sneakier approaches (e.g., casting a declared `@noinit` variable into one whose type is `@alwaysinit`). If a developer would like to violate this `@alwaysinit`, they will find a way to one way or another.

Though, yes, I do wonder if memory allocations should check the attributes the assigned pointer type and then forcibly proxy to a `new` (calloc) for `alwaysinit` types. I'm not sure exactly how feasible this is though, because enforcing that on allocations would be coupled with the standard library - in other words, `alwaysinit` would become meaningless with ordinary `malloc`. Also, can reflection even inspect attributes? :thinking:

PR #3106, opened 2026-04-06
fix #3030

### LINKED ISSUE #3030: During crash, when built as static-lib, no C3 source locations are present (Linux)
Unfortunately I am unable to make small reproduction, this could very well be with how my project is built but in nutshell I compile my app as a library and then link it with C/C++ libs and sources using `meson`

```meson
lib = custom_target(
    'OpenPNGStudio C3 library',
    output: 'libOpenPNGStudio.' + ext,
    command: [c3c, 'build', target, '--warn-deprecation=no'],
    build_always_stale: true,
)

opng = declare_dependency(link_with: lib)

# ...

executable(
    'OpenPNGStudio',
    src,
    dependencies: [raylib, sqlite, opng, libavif, libjxl, libwebp, extra_libs],
    include_directories: ['include/', 'include/vendor/'],
    install: true,
    install_dir: 'bin',
    win_subsystem: 'console',
)
```

now if for example I put `*(int*) 0 = 0;` at the beginning of `load_jpegxl` in `src/loaders/jpeg_xl.cpp`, try to load the image (it crashes on a separate thread) and then close the app, I end up with a stacktrace but a lot of C3 stuff is just `(source unavailable)`

ManuLinares:
static-lib is unconditionally forcing `--single-module` on [`src/compiler/compiler.c:526`](https://github.com/c3lang/c3c/blob/master/src/compiler/compiler.c#L526), which merges all compile units into one and loses per-file source location info. Removing that "override" fixes GNU addr2line.

I don't know the implications on removing the **force single-module on static-lib** but this fixes the issue https://github.com/c3lang/c3c/issues/3030. This was introduced in https://github.com/c3lang/c3c/commit/5d026268a75c82d652f5fae7bd5c53714d132e47

### DIFF
--- a/src/compiler/compiler.c
+++ b/src/compiler/compiler.c
@@ -523,10 +523,13 @@ void compiler_compile(void)
 				   "- If you just want to output object files for later linking, use 'compile-only'.");
 	}
 
-	if (compiler.build.type == TARGET_TYPE_STATIC_LIB)
-	{
-		compiler.build.single_module = SINGLE_MODULE_ON;
-	}
+	// Note: We intentionally do NOT force SINGLE_MODULE_ON for static libs here.
+	// Doing so merges all compile units into one, which breaks GNU addr2line's ability
+	// to resolve source locations from DWARF debug info in the resulting archive.
+	// if (compiler.build.type == TARGET_TYPE_STATIC_LIB)
+	// {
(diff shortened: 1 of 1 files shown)


lerno:
The implication is that dynamic dispatch no longer works.

ManuLinares:
> For a static library, features such as dynamic dispatch, formatting etc are completely broken without using single module.

PR #3111, opened 2026-04-06
Add stdlib::regex module - regular expression engine for C3

ChristianReifberger:
## Summary

Adds a stdlib regex engine with a hybrid execution model.

This implementation combines Thompson-style NFA matching for the regular subset with a backtracking path for advanced constructs that require it. The result is a practical, PCRE-like regex engine in the stdlib with a broad supported feature set and explicit documented compatibility boundaries.

Included in this PR:
- public regex API under `std::regex`
- support for captures, named groups, non-capturing groups, alternation, quantifiers, lazy and possessive quantifiers
- anchors and zero-width constructs including `^`, `$`, `\A`, `\z`, `\Z`, `\b`, `\B`, `\G`, lookahead, lookbehind, and `\K`
- advanced grouping and control flow including backreferences, atomic groups, branch reset groups, and conditionals
- escapes and character-class support including `\d`, `\w`, `\s`, `\h`, `\v`, `\R`, POSIX classes, Unicode whitespace properties, and selected Unicode-aware matching behavior
- replacement support including `$&`, `$0`..`$9`, `${N}`, `${name}`, and `$$`
- regex benchmarks
- systematic unit tests, matrix-style API coverage, and deterministic fuzz coverage
- public module documentation and C3 doc-contract annotations for the exported API

## Design Notes

The implementation uses a hybrid matcher:
- regular constructs are handled through Thompson-style NFA execution
- advanced constructs such as backreferences, atomic groups, conditionals, and `\G` route through a dedicated backtracking path

The code is split by responsibility:
- `lib/std/regex/parser.c3`
- `lib/std/regex/nfa.c3`
- `lib/std/regex/engine.c3`
- `lib/std/regex/regex.c3`

This PR is intentionally focused on delivering a practical stdlib regex engine with explicit documented semantics, rather than full PCRE compatibility.

## Compatibility Notes

This engine is intended to be practical and PCRE-like, but it is not a bug-for-bug compatibility target for PCRE2, Oniguruma, .NET, Java, Python, or ECMAScript.

A few important boundaries:
- Unicode support is partial and explicitly documented in `lib/std/regex/regex.c3`
- bare inline `(?i)`, `(?m)`, and `(?s)` are accepted for compatibility syntax, but compile flags are the supported way to control runtime ignore-case, multiline, and dotall behavior
- supported and unsupported syntax is documented in the public module header

## Structure

Main files:
- `lib/std/regex/regex.c3`
- `lib/std/regex/parser.c3`
- `lib/std/regex/nfa.c3`
- `lib/std/regex/engine.c3`

Tests and benchmarks:
- `test/unit/stdlib/regex/regex.c3`
- `test/unit/stdlib/regex/matrix.c3`
- `test/unit/stdlib/regex/fuzz.c3`
- `benchmarks/stdlib/regex/regex.c3`

## Verification

Focused regex suite:
```bash
./build/c3c compile-test -O0 test/unit/stdlib/regex/regex.c3 test/unit/stdlib/regex/matrix.c3 test/unit/stdlib/regex/fuzz.c3

### DIFF
--- a/benchmarks/stdlib/regex/regex.c3
+++ b/benchmarks/stdlib/regex/regex.c3
@@ -1,0 +1,306 @@
+module regex_benchmarks;
+
+import std::regex;
+import std::io;
+
+
+const DEFAULT_ITERATIONS = 50_000;
+
+// A moderately long input string used across several benchmarks.
+const String HAYSTACK = "The quick brown fox jumps over the lazy dog. "
+    "Pack my box with five dozen liquor jugs. "
+    "How vexingly quick daft zebras jump! "
+    "The five boxing wizards jump quickly. "
+    "Sphinx of black quartz, judge my vow.";
+const String EMAIL_INPUT = "Contact us at support@example.com for help.";
+const String BACKREF_HIT = "hello hello";
+const String BACKREF_MISS = "hello world";
+const String BACKREF_FIND_INPUT = "IDs: 42-42 and 71-19";
+const String CONDITIONAL_HIT = "3xdigit";
+const String CONDITIONAL_MISS = "3xother";
+const String GPOS_INPUT = "12345678x";
+
+Regex* re_literal;
+Regex* re_charclass;
+Regex* re_quantifiers;
+Regex* re_alternation;
+Regex* re_groups;
+Regex* re_anchors;
+Regex* re_complex;
+Regex* re_bt_backref;
+Regex* re_bt_atomic;
+Regex* re_bt_conditional;
+Regex* re_bt_find_backref;
+Regex* re_bt_gpos;
+
+fn void bench_setup() @init
+{
+    set_benchmark_warmup_iterations(3);
+    set_benchmark_max_iterations(DEFAULT_ITERATIONS);
+
+    set_benchmark_func_iterations($qnameof(compile_literal),       200_000);
+    set_benchmark_func_iterations($qnameof(compile_charclass),     100_000);
+    set_benchmark_func_iterations($qnameof(compile_quantifiers),   100_000);
+    set_benchmark_func_iterations($qnameof(compile_alternation),    80_000);
+    set_benchmark_func_iterations($qnameof(compile_groups),         80_000);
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by lerno** [comment]
Name types uniquely or make them `@local`. The use of "Thread" here is particularly unfortunate.

**Comment by ChristianReifberger** [comment]
> Name types uniquely or make them `@local`. The use of "Thread" here is particularly unfortunate.

@lerno Thank you, I’ve renamed the types.

I’m still working through the remaining Windows CI failures. If you have additional review feedback in the meantime, feel free to leave it and I’ll fold it in while I’m addressing that.

The goal of this PR is to land the initial stdlib regex engine with broad feature coverage. I’m happy to adjust behavior, API details, or compatibility choices if you think any of those should change before merge.

**Comment by ManuLinares** [comment]
could you take a look?

```c3
import std;

fn void main()
{
	sz size = 10000;
	String input = (String)alloc::new_array(tmem, char, size);
	mem::set(input.ptr, 'a', size);

	file::save("test.txt", input)!!;

	String pattern = "(a|b|c|d|e|f|g|h|i|j)*x";
	Regex* re = regex::tcompile(pattern);

	long c3_t   = @bench() { re.is_match(input); };
	long grep_t = @bench() { process::run({"grep", "-E", pattern, "test.txt"})!!; };

	io::printfn("%-10s: %d ms", "C3 time", c3_t);
	io::printfn("%-10s: %d ms", "Grep time", grep_t);
}

macro long @bench(; @body)
{
	Clock s = clock::now();
	@body();
	return s.to_now().to_ms();
}
```
```bash
c3c compile-run -O1 regex_test.c3 

C3 time   : 21911 ms
Grep time : 0 ms
```

lerno:
So I've looked at this one as well, and it's so much code compared to other implementations, about 3-4 times bigger. And this is a problem for something for the stdlib as previously mentioned. I would again suggest you package this as a .c3l and then make it available to people on github. You can also submit it to c3-showcase where people can easily find it.

PR #3010, opened 2026-03-09
compiler(fix): prevent worker-thread stack overflow during LLVM codegen

fernandolguevara:
Fix worker-thread stack overflow in LLVM codegen. Added 8MB pthread stack and single-thread inline fallback to prevent SIGBUS on stack guard.

PR #3117, opened 2026-04-09
fix(thread-posix): prevent worker-thread stack overflow

fernandolguevara:
#3010


lerno:
I've added the ability to set the stack size now.

PR #3118, opened 2026-04-10
don't override the sigaltstack when address sanitizing: fixes #3115

ManuLinares:
fixes https://github.com/c3lang/c3c/issues/3115


lerno:
Add releasenotes so we can track this fix

lerno:
Awesome!

PR #3119, opened 2026-04-10
gzip Instream/Outstream + example http_server

### LINKED ISSUE #2733: `net::addrinfo` on Windows needs initialized WSA
Several functions in stdlib (`tcp::connect`, `tcp::connect_async`, `tcp::listen`, `udp::connect`, `udp::connect_async`)
call `net::addrinfo` without prior call to `os::start_wsa`. This means that `net::addrinfo` fails if WSA hasn't been initialized before
([documentation of getaddrinfo](https://learn.microsoft.com/en-us/windows/win32/api/ws2tcpip/nf-ws2tcpip-getaddrinfo) says that it can return `WSANOTINITIALISED`).

ManuLinares:
- fix deflate bug
- proper InStream and OutStream in gzip stdlib

Add http_server example:

- with directory listing (like firefox's builtin because I like it)
- gzip compression support  
- partial content (Range header) for media files
- one thread per connection (as simple as it gets)
         
mostly to provide a `python -m http.server $port` example in c3

- Uses `ThreadPool`
- Added a test in `ci_tests.sh`
- Fixed a bug in `lib/std/net/inetaddr.c3` for WIN32

---

Notes;
- gzip still could use some convenience functions like zip
- gzip does not set `mtime` when compressing
- haven't fully tested in 32bit environment

Fixes: https://github.com/c3lang/c3c/issues/2733


lerno:
Could you resolve the merge conflicts and also add something in the releasenotes 🙏

PR #3120, opened 2026-04-10
Implementing Mergesort

sandervdbosch:
I am contributing mergesort as an sorting option to the standard library.

This is a working version of the code, but I am having trouble understanding how to get the implementation on the same level as the other sorting algorithms. 

Suggestions on how to proceed with this are highly appreciated. When the implementation is ready I will also make the unit tests.


sandervdbosch:
The implementation is now completed.
The Mergesort used O(n) memory for the sorting and is almost as fast as quicksort in most cases.

This results in the advice to use Mergesort only on relatively small data amounts.

The one thing I would like to get some feedback on is the memory allocator argument. Normally functions have the allocator as the first argument, but that would interfere with the generic module.

lerno:
Thank you!

PR #3121, opened 2026-04-10
Adding the baseline needed for avr support

Jotrorox:
This commit outlines the base needed to use C3 for embedded development on 8-bit AVR-family microcontrollers like the ATmega328p (mostly used in the Arduino Uno).

I've run the entire test suite with the recompiled compiler and also built a simple Arduino demo using C3 as the only language, not depending on handwritten assembly, etc.

> I know the changes made are quite limited in scope and the support is still minimal. If this is too little I'd appreciate a direction what to add next to make this support more fleshed out. Overall I don't really know what the formally correct way to file this is I just wanted to share this. So if this just isn't enough or I did anything wrong please just tell me what to fix or do next time and thanks a lot for your time reading this and reviewing this pr :)


src/compiler/target.c
```diff
@@ -2167,6 +2181,12 @@ void target_setup(BuildTarget *build_target)
 		error_exit("Xtensa support is not available with this LLVM version.");
 	}
 #endif
+#ifdef AVR_DISABLE
+	if (build_target->arch_os_target == ELF_AVR)
+	{
+		error_exit("AVR support requires building with -DAVR_DISABLE.");
```
lerno: This doesn't make sense to me. This error is just shown IF you build it with -DAVR_DISABLE. And why would you add such a macro to *enable* AVR? I'm confused.

Jotrorox: I'm sorry I forgot to change the exit message to something like `AVR support requires not building with -DAVR_DISABLE.` during my final iteration. I first wanted to mimic the XTENSA_ELF enable logic that the support has to manually be put in, but since avr is a well established target for llvm it shouldn't be an issue if built in by default. I'm sorry for the confusion and if I get anything wrong.

edit: since the CI uses the custom LLVM build I have opened a pull request there adding AVR as a target to make building with AVR enabled by default possible. If this goes too far I'm sorry.

lerno: It's fine, but probably you want it disabled by default as people will use other LLVM builds that don't always support it. Adding AVR as a target for LLVM is good

Jotrorox:
I have added the AVR_ENABLE flag to the CMakeLists.txt file so if the user wants to (not) use it that would work fine. 
The CI/CD has confirmed to still be working on the branch: https://github.com/Jotrorox/c3c/actions/runs/24328011517

If this is something that is wanted in the standard builds of the compilers I'd wait for the next llvm release so the changes (https://github.com/c3lang/llvm-for-c3/pull/4) in the prebuilt llvm are applied.

Note: the android actions fail because the branch is not up to date with those fixes

If there is anything else wanted/needed I'd be more then happy to provide it :)

src/build/builder.c
```diff
@@ -321,6 +321,8 @@ static LinkLibc libc_from_arch_os(ArchOsTarget target)
 		case ELF_X64:
 		case ELF_XTENSA:
 			return LINK_LIBC_OFF;
+		case ELF_AVR:
+			return LINK_LIBC_OFF;
```
Darvisim: If you'd like, you can change this to:

```c
case ELF_EXTENSA:
case ELF_AVR:
    return LINK_LIBC_OFF;
```

Just a little suggestion from my side. :)

Jotrorox: Thanks for the notice and I've applied it in https://github.com/c3lang/c3c/pull/3121/commits/3da9458ffcdb32c38e20df934e96109b204610d8 sorry for not responding earlier I just wanted to not make that single line change a full commit :)

Darvisim: I took a look into the commit and noticed that you've removed the `ELF_AVR` case, which caused the CI to fail at CMake Build step:
```cmake
error: enumeration value 'ELF_AVR' not handled in switch [-Werror,-Wswitch]
  290 |         switch (target)
      |                 ^~~~~~
1 error generated.
```
Was the removal of `ELF_AVR` case intentional?

Jotrorox: I somehow fucked up the squash, I to be honest don't even know how I've managed to do that. So no that wasn't intentional. I just currently am a bit stressed since my finals are next week so I don't know If I'm able to resolve it in the next few days, sorry.

lerno:
Thank you

PR #3124, opened 2026-04-11
stdlib: String and DString optimizations

ManuLinares:
I've used this to benchmark.
[reverse_words.c3.zip](https://github.com/user-attachments/files/26642267/reverse_words.c3.zip)


lerno:
Is it ok @ManuLinares ?

ManuLinares:
> Is it ok @ManuLinares ?

Yes in a more general way. I was optimizing **that** specific benchmark.

To clarify: Yes it is faster using `append_inline` for this benchmark!

Issue #3061, opened 2026-03-25
`exec` in project.json should run relative to project.json path


ManuLinares:
`exec` steps currently assume the working directory is `scripts/`, which is confusing and inconsistent.
In my opinion `exec` should run with the working directory set to the location of `project.json`.
This avoids ambiguity and removes the need for a `scripts/` directory when not used.

---

Also the incongruent `script-dir` with `$exec` naming: 
```
 $ c3c --help | grep script-dir
  --script-dir <dir>         - Override the base directory for $exec.
```

lerno:
Can I close this?

PR #3128, opened 2026-04-12
Set 'exec' cwd to project root and rename 'script-dir`to 'exec-dir' in CLI #3061

ManuLinares:
- Default the 'exec' working directory to the project.json location when `c3c build/dist/run`.
- Rename 'script-dir' to 'exec-dir' in project.json and CLI.
  -  It was weird seeing in help `--script-dir <dir>  - Override the base directory for $exec.` and I think this is more consistent.
- Remove automatic creation of the `scripts/` directory for new projects.
- Update test projects and examples.

TLDR:
1. Project builds: run `exec` relative to the project.json directory.
2. Standalone compiles: run `$exec` relative to your current terminal's working directory.

Addresses: https://github.com/c3lang/c3c/issues/3061

It is "technically" a breaking change but because it's for 0.8.0 I didn't add deprecations.

### DIFF
--- a/releasenotes.md
+++ b/releasenotes.md
@@ -59,7 +59,7 @@
 - Add `compare_to` and `compare_to_ignore_case` to `String`. #3096
 - Add `SortedMap` based on skip lists.
 - Add `OneShotChannel` to `std::thread::channel` for single-send/single-receive thread synchronization.
-- `BufferedChannel` and `UnbufferedChannel` now pointers, create using `create_unbuffered` and `create_buffered`
+- `BufferedChannel` and `UnbufferedChannel` are now pointers, create using `create_unbuffered` and `create_buffered`
 - `RingList` now conforms to `foreach` and adds additional functions.
 - Ini parser and encoder.
 - Updated `ref::new` argument order.
--- a/resources/testproject/project.json
+++ b/resources/testproject/project.json
@@ -15,7 +15,7 @@
   "dependencies": ["clib", "clib2"],
   "cpu-flags": ",,",
   "features": ["ABCD"],
-  "exec": ["scriptme.c3 myarg"],
+  "exec": ["scripts/scriptme.c3 myarg"],
   // c sources
   "targets": {
     "hello_world": {
--- a/resources/testproject/project.json
+++ b/resources/testproject/project.json
@@ -15,7 +15,7 @@
   "dependencies": ["clib", "clib2"],
   "cpu-flags": ",,",
   "features": ["ABCD"],
-  "exec": ["scripts/scriptme.c3 myarg"],
+  "exec": ["scriptme.c3 myarg"],
   // c sources
   "targets": {
     "hello_world": {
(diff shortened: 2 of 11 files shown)


lerno:
I think this is the wrong way to do it. Here's what it should be instead:

1. Keep `script-dir`. This is where you find scripts by default, which is where `exec` and `$exec` will look in for `build_script.c3` etc.
2. Add `exec-dir` which will default to `scripts`, but is then editable.

This will not break anything and would behave as intended with more possibilities for customization. The proposed change in this PR just swaps one default for another and serves as the same amount of lock-in. My idea how to resolve #3061 was always to add more customizability, not change the defaults.

lerno:
Note that the problem with `script-dir` is exactly that of lacking customizability, not that it's fundamentally wrong. In the simple case where you keep the script and the script resources in the `script-dir` it's the natural behaviour. It's when you want to create artifacts using `exec` / `$exec` that things require more flexibility. And here the best solution would be to be able to customize it. It's just that the option was missing.

ManuLinares:
Yes we need to have a proper discussion about this.

`script-dir` is poorly named in my opinion. There is no concept of a `script-dir` in the language. We currently have:

1. In `project.json` we have:
    - `exec`
    - `exec-override`
    - `script-dir` (which effectively acts as the cwd of `exec` and `$exec`, and in my opinion should be named `exec-dir`)
2. In C3 code we have:
    - `$exec`  [1](https://c3-lang.org/generic-programming/compiletime/#exec) [2](https://c3-lang.org/language-fundamentals/modules/#exec) The default cwd of this is hardcoded to the folder named "scripts" and can be changed via the aforementioned `script-dir` 
3. In `c3c` CLI we have the option: 
    - `--script-dir <dir>  - Override the base directory for $exec.`
    - *which in my opinion should also be named `--exec-dir`*

---

My proposal unifies (yes, simplifies) these concepts.

I don't understand why we need to even have a "scripts" dir, if it's required, it should be called "exec-dir" for consistency.
I don't think we need to model "script discovery" and "execution cwd" as separate concepts in C3, because `$exec` does not expose or benefit from that distinction.

- Making the default cwd of any `project.json` execs to the `project.json` directory is consistent with other tools in my experience (and can be overridden).
- Keeping the `c3c` CLI default `exec`/`$exec` cwd as the directory where `c3c` is executed also keeps consistency (and can be overridden).
- I don't think this restricts customizability or maybe I missed something.

This is not intended to convince you, just to lay out my reasoning.

lerno:
script-dir is where you store your scripts. Those may be .sh files, actual executables or .c3 files. The "exec-dir" is where the scripts are executed and presumably places their results.

So for example `$exec` may run code that generates some sources. This can then be placed in a sub-directory to the exec directory, eg `generated-src`. This is natural from my point of view - it mirrors how projects are usually laid out. The scripts are the central thing, where they are executed – secondary.

To me saying "exec-dir" is the central place is to me like saying `/build` is the central thing, and `/src` is secondary and code should just be placed in build.

lerno:
I updated it the way I meant it should be, please check if there is some issue with doing it like this @ManuLinares

lerno:
I'll merge this as is.

Issue #3026, opened 2026-03-12
C3 mem::malloc() is 4 times slower than C malloc()


forchid:
# The version info
```sh
>c3c -V
C3 Compiler Version:       0.7.11 (Pre-release, Mar 12 2026 00:18:09)
Installed directory:       C:/Programs/c3
Git Hash:                  c1aed72aeaab9759fa029a180a845360fb5e6272
Backends:                  LLVM
LLVM version:              22.1.1
LLVM default target:       x86_64-pc-windows-msvc
```
# The test case
The c3 test
```sh
>c3c -O2 compile alloc-c3-perf.c3
Program linked to executable '.\alloc_c3_perf.exe'.

>alloc_c3_perf
ptr -> 0
ptr -> 10000000
ptr -> 20000000
ptr -> 30000000
ptr -> 40000000
ptr -> 50000000
ptr -> 60000000
ptr -> 70000000
ptr -> 80000000
ptr -> 90000000
Time 24.175713s
```
The C test
```sh
>cl -nologo -O2 alloc-c-perf.c
alloc-c-perf.c

>alloc-c-perf
ptr -> 0
ptr -> 10000000
ptr -> 20000000
ptr -> 30000000
ptr -> 40000000
ptr -> 50000000
ptr -> 60000000
ptr -> 70000000
ptr -> 80000000
ptr -> 90000000
Time 6.281000s
```
# The test sources
C3 source
```c3
// alloc-c3-perf.c3
import std;

const N = 100000000;

fn void main() {
  time::Time start_time = time::now();
  
  for(int i = 0; i < N; i++) {
    int *ptr = malloc(int.sizeof);
    *ptr = i;
    if (i % (N / 10) == 0) {
      io::printfn("ptr -> %s", *ptr);
    }
    free(ptr);
  }
  
  time::Time end_time = time::now();
  double sec = end_time.diff_sec(start_time);
  io::printfn("Time %ss", sec);
}
```
C source
```c
// alloc-c-perf.c
#include <stdio.h>
#include <stdlib.h>
#include <time.h>

const int N = 100000000;

int main(void) {
  clock_t start_time = clock();
  
  for(int i = 0; i < N; i++) {
    int *ptr = malloc(sizeof(int));
    *ptr = i;
    if (i % (N / 10) == 0) {
      printf("ptr -> %d\n", *ptr);
    }
    free(ptr);
  }
  
  clock_t end_time = clock();
[...]

PR #3129, opened 2026-04-12
Add LLVM allocation attributes to malloc/free #3062

### LINKED ISSUE #3062: [Enhancement request] Missing LLVM allocation attributes for `libc::malloc` prevent optimizations
This is just a reminder that C3 currently doesn't:
- Elide allocations in loops
- Perform full alias analysis
- Apply equivalent optimizations as in C

This leads to **apparent** performance gaps vs C in benchmarks.

---

Annotate `libc::malloc` with appropriate LLVM attributes to enable optimization parity with C.

## Reference
https://llvm.org/docs/LangRef.html#function-attributes

https://github.com/c3lang/c3c/issues/3026

*No comments needed, this is an "enhancement request" issue*

ManuLinares:
apply `allockind`, `noalias`, etc to standard memory functions. LLVM elides redundant allocations and matches C's optimization behavior. https://github.com/c3lang/c3c/issues/3026

addresses https://github.com/c3lang/c3c/issues/3026
fixes https://github.com/c3lang/c3c/issues/3062

I don't think this applies to **our** allocators but could be expanded to user attributes?

maybe expanding this to a user-facing attribute like @allockind could be nice for custom allocators but from my testing llvm does not apply good optimizations. LLVM is much more aggressive with the standard malloc name than it is with allockind on arbitrary names `¯\_(ツ)_/¯`


forchid:
@ManuLinares When adding the flag `-g0` in C3 0.7.10 or 0.7.11, I find that C3 malloc() is close to the speed of C. See #3026 .

ManuLinares:
Closing this for now, not _really_ needed

PR #3132, opened 2026-04-13
Added std::collections::result to the standard lib

Ecoral360:
As discussed on discord, the `Result` struct will allow errors to carry a payload.

`Result`s should only be used when simple faults are not enough and should not be used as the default. The documentation included with the `Result` struct is clear about this.

The api only includes 2 constructors, 2 accessors, printing, `==` operation (if both OkType and ErrType support it) and the `result.@or_else` macro that allows the result to diverge on error.

### DIFF
--- a/lib/std/collections/result.c3
+++ b/lib/std/collections/result.c3
@@ -1,4 +1,5 @@
 module std::collections::result <OkType, ErrType>;
+import std::io;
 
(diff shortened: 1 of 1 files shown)


lerno:
Next time please don't forget releasenotes and tests.

PR #3133, opened 2026-04-14
feat(vendor-fetch): allow decompress downloaded .c3l

fernandolguevara:
this adds extraction flag support for downloaded `.c3l` in `project fetch -e | vendor-fetch -e`, existing library paths are not replaced without `-f/--force` flag

### DIFF
--- a/c3c.1
+++ b/c3c.1
@@ -127,7 +127,7 @@ Analyze files and generate C headers for public methods.
 .RE
 .PP
 .B c3c vendor-fetch
-[\fB-e\fR|\fB--extract\fR] [\fB-f\fR|\fB--force\fR] \fIlibrary\fR ...
+[\fB-e\fR|\fB--extract\fR] [\fB-f\fR|\fB--force\fR] [\fB-t\fR|\fB--target\fR] \fIlib\fR ...
 .RS
 Fetch one or more libraries from the vendor collection.
 .RE
@@ -173,6 +173,11 @@ Add this library to the compilation.
 \fIdir\fR
 .RS
 Use this as the base directory for the current command.
+.RE
+.PP
+.B -t, --target
+.RS
+For \fBvendor-fetch\fR, allow to specify \fBtarget\fR for fetching dependencies.
 .RE
 .PP
 .B -e, --extract
--- a/c3c.1
+++ b/c3c.1
@@ -127,7 +127,7 @@ Analyze files and generate C headers for public methods.
 .RE
 .PP
 .B c3c vendor-fetch
-\fIlibrary\fR ...
+[\fB-e\fR|\fB--extract\fR] [\fB-f\fR|\fB--force\fR] \fIlibrary\fR ...
 .RS
 Fetch one or more libraries from the vendor collection.
 .RE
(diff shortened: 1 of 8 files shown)


ManuLinares:
*I though I wrote a comment here, github playing tricks, rewriting it now:*

---

Bug: It rewrites the file,even without the `--force`

```bash
$ c3c vendor-fetch raylib
Fetching library 'raylib'...
Downloading [████████████████████████████████████████] 100%
Fetching complete.		

$ ls
raylib.c3l

$ c3c vendor-fetch -e raylib
Fetching library 'raylib'...
Downloading [████████████████████████████████████████] 100%
Fetching complete.		

$ ls
raylib.c3l/
```

---

Rename `<library>` to `<lib` in `--help` to keep indentation nice
```
  dynamic-lib <file1> [<file2> ...]                   Compile files without a project into a dynamic library.
  vendor-fetch [-e|--extract] [-f|--force] <library>... Fetch one or more libraries from the vendor collection.
  project <subcommand> ...                            Manipulate or view project files.
```

---

Better describe what's happening here in the message:
```
				printf("Please delete library directory '%s'...\n", archive_path);
```

fernandolguevara:
@ManuLinares fixed, can you try again?

ManuLinares:
Describe what's changed.

ManuLinares:
Could you, in all honesty answer me, and describe what's changed?

I'm asking overall and specifically this: https://github.com/c3lang/c3c/commit/474f77baa2d656feb31c50a0fb5b7ceb0f33742c#diff-942a1a9975c1f31b7cc2cdbb537b32d462778278fb4f2f23bae6600f0c5279fe

fernandolguevara:
@ManuLinares 
- Added `DownloadProgressState` to fix an issue with the progress bar. In some cases (e.g. `pq`) it wasn’t being displayed properly (see image below).
  imagen
- Added `-e|--extract` `-f|--force` flags to `c3c project fetch`.
  Also refactored it to reuse the internal vendor-fetch logic, since both commands are very similar (the only difference is that `c3c project fetch` fetches only missing dependencies).
- Added `-t|--target` flag to `c3c vendor-fetch | c3c project fetch` to install deps for a specific target.

ManuLinares:
>     * Added `DownloadProgressState` to fix an issue with the progress bar. In some cases (e.g. `pq`) it wasn’t being displayed properly (see image below).

Is this something you can reproduce?

I can't seem to be able to reproduce this one. I don't think we download things in parallel threads which is the only way I see the previous static var maybe not safe.

fernandolguevara:
> Is this something you can reproduce?

yup, with master code, run `c3c vendor-fetch raylib pq vulkan`

imagen

lerno:
I must have forgotten to hit return on the original comment on this, but the flags are incorrectly implemented. The way things work is that flags are common to all targets, one never adds ad-hoc flag handling. As it stands this must be rejected on the grounds that the flags are incorrectly implemented. Right now it's just bulldozing the current build option handling and is dancing on its corpse.

lerno:
I feel we never discussed whether this feature should even go in. It says "decompress downloaded .c3l" but contains "force" and "target" as well.

PR #3135, opened 2026-04-15
Fix swapped arguments in aligned_offset for vector codegen

ManuLinares:
#### testcase:
```c3
module vector_align_test;

struct Foo
{
	char[<5>] v2 @align(4);
}

fn void main()
{
	Foo f;
	f.v2 = { 1, 2, 3, 4, 5 };
}
```
```
c3c compile vector_align_test.c3 --emit-llvm && grep "store <" llvm/linux-x64/vector_align_test.ll
```

#### current:
```
  store <8 x i8> <i8 0, i8 0, i8 0, i8 0, i8 0, i8 undef, i8 undef, i8 undef>, ptr %f, align 4, !dbg !21
  store <5 x i8> <i8 1, i8 2, i8 3, i8 4, i8 5>, ptr %f, align 4, !dbg !22
```

#### fixed:
```
  store <8 x i8> <i8 0, i8 0, i8 0, i8 0, i8 0, i8 undef, i8 undef, i8 undef>, ptr %f, align 4, !dbg !21
  store <8 x i8> <i8 1, i8 2, i8 3, i8 4, i8 5, i8 undef, i8 undef, i8 undef>, ptr %f, align 4, !dbg !22
```

PR #3138, opened 2026-04-15
Better align `.clang-tidy` to the c3c codebase

ManuLinares:
Changes the `.clang-tidy` and `.clang-format` to better align to the c3c codebase and by default catch more with less static analysis.

- I caught at least [one bug](https://github.com/c3lang/c3c/pull/3135)
- Added some `// NOLINT` but very specific ones to not clog the other analysis.
- Also enforced scope braces on some macros for better static analysis and overall format niceness. It reads better but the static analysis is a plus.

Nothing mayor, just tidying up stuff for better static analysis in general and less noise (better defaults).

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,11 +1,17 @@
+Language: C
+
 IndentWidth: 4
-UseCRLF: false
-IndentCaseLabels: true
-UseTab: ForIndentation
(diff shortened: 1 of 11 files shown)


lerno:
Great!

PR #3139, opened 2026-04-15
build: remove git binary dependency for hash extraction

ManuLinares:
Parse `.git/HEAD` and `packed-refs` directly via cmake instead of invoking `git rev-parse`. This drops `git` as a strict build dependency.

I was doing a bash script but that defeated the purpose of removing a dependency. This is better (got the idea from mold: https://github.com/rui314/mold/blob/main/lib/update-git-hash.cmake)


lerno:
👍

PR #3141, opened 2026-04-15
CI: support continuous releases for `*.*.x` branches

ManuLinares:
CI `main.yml` now also triggers on commits to `*.*.x` branches:

  - Commits to a `*.*.x` branch will now generate a prerelease tag and title matching the branch name + `-latest` (for example `v0.7.x-latest`).
  - Commits to master will continue to generate `latest-prerelease-tag`.


.github/workflows/main.yml
```diff
@@ -2,7 +2,7 @@ name: CI
 
 on:
   push:
-    branches: [ master, dev, ci_testing, experiments ]
+    branches: [ master, dev, ci_testing, experiments, '*.*.x' ]
```
Book-reader, line 5:
```suggestion
    branches: [ master, dev, ci_testing, experiments, 'v[01].[0-9]+.x' ]
```
This isn't likely to matter all that much, but being more explicit to exactly what branches it should match can't hurt.
The `*.*.x` pattern matches both the `0.6.x` and `v0.6.x` branches for example (though both branches existing might be an accident)

ManuLinares: We might actually **do** want to trigger for those as-well, in fact I thought on making this a broader "*.x" so we could track very minor hotfixes "v0.8.7.2x" but seemed unnecessary.

Book-reader: > We might actually do want to trigger for those as-well [...]

As in track branches named both `vX.X.x`  and `X.X.x`?

> [...] in fact I thought on making this a broader "*.x" so we could track very minor hotfixes "v0.8.7.2x" but seemed unnecessary

I think lerno has said that he would prefer to overwrite a release than release a `0.X.X.X` version, so I don't know if that's likely to be an issue in the future. `*.*.x` might match `v0.8.7.2.x`  anyway unless `*` is a non-greedy match (`.*?` instead of `.*` in regex), it isn't clear from the [github docs](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#filter-pattern-cheat-sheet)

In any case, it probably doesn't matter, so feel free to close this suggestion.

lerno:
Nice!

PR #3142, opened 2026-04-15
Limits the size of the sortedmap skip list per-node forward slice to the actual level of the node

corleypc:
Reduces the allocated forward slice in sortedmap skip list nodes from the fixed max level (32 * 8 = 256 bytes on 64-bit platforms) to the actual rolled level. Half the nodes (asymptotically) live on the bottom level and only require a single forward item and in practical use no node hits the max level.

Also caps level growth on insertion to no more than current level + 1, which avoids (otherwise harmless) looping through empty levels caused by potential outlier rolls in small maps.

### DIFF
--- a/lib/std/collections/sortedmap.c3
+++ b/lib/std/collections/sortedmap.c3
@@ -34,7 +34,7 @@ struct SkipNode
 	Key key;
 	Value value;
 	int level;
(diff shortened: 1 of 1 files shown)


lerno:
Thank you!

PR #3145, opened 2026-04-16
fix(target_setup): fallback cpu llvm value when available

fernandolguevara:


ManuLinares:
Thanks, 

1. This is not needed, already handled in `src/compiler/target.c:1984`:
  ` compiler.platform.cpu ? compiler.platform.cpu : ""`
2. Breaks cross-compilation.
3. _No description provided._ ?

Please read the [CONTRBUTING.md](https://github.com/c3lang/c3c/blob/master/CONTRIBUTING.md) guidelines regarding [AI usage in contributions](https://github.com/c3lang/c3c?tab=contributing-ov-file#5-ai-usage-in-contributions).

fernandolguevara:
@ManuLinares thanks! 

1. fallback value and what you pointed handled in `src/compiler/target.c:1984` isn't the same.
2. fixed now!

lerno:
It has a TODO?

ManuLinares:
> @ManuLinares thanks!
> 
>     1. fallback value and what you pointed handled in `src/compiler/target.c:1984` isn't the same.
> 
>     2. fixed now!

This is what I was trying to avoid. [Do not use PR reviews as an AI feedback loop](https://github.com/c3lang/c3c?tab=contributing-ov-file#5-ai-usage-in-contributions)

I must ask again, why is it that you don't provide any description? On PRs or on changes done to PRs?

fernandolguevara:
@lerno

> It has a TODO?

do you have any suggestion, on what to do when llvm is not available?

fernandolguevara:
> > @ManuLinares thanks!
> > ```
> > 1. fallback value and what you pointed handled in `src/compiler/target.c:1984` isn't the same.
> > 
> > 2. fixed now!
> > ```
> 
> This is what I was trying to avoid. [Do not use PR reviews as an AI feedback loop](https://github.com/c3lang/c3c?tab=contributing-ov-file#5-ai-usage-in-contributions)
> 
> I must ask again, why is it that you don't provide any description? On PRs or on changes done to PRs?

@ManuLinares this isn’t AI-generated code. If there’s a concrete concern, let’s focus on that.

ManuLinares:
I don't understand what is "the problem" being solved? This is why I ask for a _description_

fernandolguevara:
@ManuLinares since we already own `compiler.platform.cpu` it should be the source of truth and its value propagated to the underlying layers.

ManuLinares:
> do you have any suggestion, on what to do when llvm is not available?

When is not LLVM available?

> @ManuLinares since we already own `compiler.platform.cpu` it should be the source of truth and its value propagated to the underlying layers.

what? I'm confused by your words.

lerno:
Isn't it more reasonable to simply have a "native-cpu" built option?

lerno:
I'll close this, we need to have something that works correctly with everything. Providing a CPU like this won't do that, it needs many more things.

PR #3156, opened 2026-04-19
src/utils/json.c: fix linking warning in mold

ManuLinares:
mold is a little stricter, so this prevents:
```
[99/99] Linking CXX executable c3c
mold: warning: symbol type mismatch: error
>>> defined in CMakeFiles/c3c.dir/src/utils/json.c.o as STT_OBJECT
>>> defined in /usr/lib/libc.so.6 as STT_FUNC
```


lerno:
Thank you!

PR #3158, opened 2026-04-21
Add free list allocator and buddy allocator to stdlib

nyr24:
- add free list allocator to stdlib
- add buddy allocator to stdlib


lib/std/core/mem.c3
```diff
@@ -12,7 +12,7 @@ const ulong MB = KB * 1024;
 const ulong GB = MB * 1024;
 const ulong TB = GB * 1024;
 
-faultdef OUT_OF_MEMORY, INVALID_ALLOC_SIZE;
+faultdef OUT_OF_MEMORY, INVALID_ALLOC_SIZE, POINTER_NOT_OWNED;
```
lerno: Pointer not owned should be a panic, not an error?

nyr24: ok, I'll fix it

nyr24: it should be fixed now

lerno:
You need to look into the bug with the allocator.

nyr24:
Yeah, I'm trying to do that currently

nyr24:
@lerno I've fixed that!

lerno:
What happens in the bst heap allocator when you call remove_node on a node that has children? Can that happen?

nyr24:
Yes, good catch, I will fix this

nyr24:
@lerno it has been fixed

nyr24:
Apparently not, bug still exists, marking as draft for now

nyr24:
This bug was addressed, but there seems to be compiler internal errors in failing checks

lerno:
Allocators are notoriously hard to get right, and I think these are not yet battle hardened enough. Please provide them as a library instead, get users and track bug reports and then when they're clearly solid and has seen real life, let's look at integrating them again. Right now they're a little too buggy still.

Issue #3163, opened 2026-04-22
Collection::object.get_enum not working properly


Blu-ea:
https://github.com/c3lang/c3c/blob/1fd63d2ef12583e41b37048dae5201b654a40a28/lib/std/collections/object.c3#L400
```c3
// Original Function 
macro String? Object.get_enum(&self, $EnumType, String key)
{
	Object value = self.get(key)!;
	if ($EnumType::typeid != value.type) return TYPE_MISMATCH~;
	return ($EnumType)value.i;
}
```

- `self.get(key)!` return a pointer, but the variable `value` is not.
```c3 
fn Object*? Object.get(&self, String key);
```
- The return type of the function is `String?`, shouldn't it be the `enum?`?   
- This also applies to [`get_enum_at`](https://github.com/c3lang/c3c/blob/1fd63d2ef12583e41b37048dae5201b654a40a28/lib/std/collections/object.c3#L410)

---

Even when I modify it myself, I cannot get it to work out because of the `EnumType::typeid != value.type` being always false. 

```json
{
    "enum_key1": 1,
    "enum_key2": "SECOND",
    "enum_key3": "third"
}
```

```c3
// my modification 
macro String? Object.get_enum(&self, $EnumType, String key)
{
	Object* value = self.get(key)!;  // adding the pointer
	if ($EnumType::typeid != value.type) return TYPE_MISMATCH~;
	return "Type found !"; // changing the return to a static string
}

enum MyEnum : (String value) {
	FIRST {"first"},
	SECOND {"second"},
	THIRD {"third"}
}

fn void main(){
	String file_content = (String) file::load(mem, "file.json")!!;

	Object* a = json::parse(mem, file_content)!!;
	
	io::printn(a.get_enum(MyEnum, "enum_key1")!!); // Will return builtin::TYPE_MISMATCH
	io::printn(a.get_enum(MyEnum, "enum_key2")!!); // Will ALSO return builtin::TYPE_MISMATCH
	io::printn(a.get_enum(MyEnum, "enum_key3")!!); // And so does this one return builtin::TYPE_MISMATCH
}
```

I don't know if it's an issue on my part here or if the standard library is in fault.

lerno:
Should be fixed now or?

Blu-ea:
Now using the `get_value()`, i have an issue when loading an `object` with `json::parse()`

```json
{
	"enum_key1": 1,
}
```

```c3
enum MyEnum : int {
	FIRST,
	SECOND,
[...]

PR #3164, opened 2026-04-22
Add Object.to_value. Fix Object.get_enum and Object.get_enum_at

Ecoral360:
Fixed the `Object.get_enum` and `Object.get_enum_at`

Added a `Object.to_value` to retrieve a value from an object directly.

See issue #3163


lerno:
is_enum looks wrong?

Ecoral360:
Yes you are right, I fixed it and added a test

lerno:
Good!

Issue #3172, opened 2026-04-25
Possible `Path` additions


cmann1:
## Accept multiple parts when appending to a path

```c3
fn Path? Path.append_m(self, Allocator allocator, String... parts, String ext = "") {}
```

```c3
Path p = path::temp("root");
p.tappend("a", "b", "c", ext:"txt");
io::printn(p); // > root/a/b/c.txt
```

Allowing the constructors to accept multiple parts might also make sense?

```cpp
Path p = path::temp("root", "a", "b", "c.txt");
```

## Relative to

```c3
<* Returns true if this path is relative to another. *>
fn bool Path.is_relative_to(self, Path other) {}

<*
 Returns this path relative to another.
 If walk_up is true, `..` may be added, otherwise an error is thrown.
*>
fn Path? Path.relative_to(self, Path other, bool walk_up = false) {}
```

PR #3181, opened 2026-04-28
Add support for multiple parts to `Path.append` (#3172)

cmann1:
This is only the change to `append` - the `relative_to` method seems more involved so I don't know if I'll attempt that.

How much does the initial capacity matter - is a rough guess like this ok, or would it be preferable to sum the lengths beforehand?


lerno:
It would be nice to get a better guesstimate for the temp allocator so there is no realloc on the temp allocator.

lerno:
I forgot to request a test for it. Could you add one please?

lerno:
Thank you!

PR #3185, opened 2026-04-29
Implementing a preset option for targets in libraries

Jotrorox:
This change is (for now) just a proposal: it implements a way to define presets for certain target configurations by the library authors. This is especially useful for developers working with libraries targeting microcontroller boards like the Arduino or ESP32 family. The user won't have to fiddle with those configurations, like what to enable, what to disable, what CPU architecture to use, or whether to not set an entry point.

This, at least in my opinion, provides a huge boost in ease of use without taking away from the customizability of C3 projects. Everything can be overridden by the user by just setting the configuration in the user's project.json file.

I just wanted to put this out here and get some feedback on whether this is wanted or not or how this could be changed for a better dx.

**Example `manifest.json`:**
```json
{
  "provides": "arduino",
  "presets": {
    "arduino_uno": {
      "cpu": "atmega328p",
      "target": "elf-avr",
      "opt": "Oz",
      "linker": "builtin",
      "use-stdlib": false,
      "link-libc": false,
      "no-entry": true
    }
  },
  "targets": { ... }
}
```

**Example `project.json`:**
```json
{
  "dependencies": ["arduino"],
  "dependency-search-paths": ["lib"],
  "targets": {
    "pot_fade": {
      "type": "executable",
      "preset": "arduino:arduino_uno",
      "name": "pot_fade",
      "extension": ".elf"
    }
  }
}
```


lerno:
Interesting!

Jotrorox:
The current state of this PR:

- Added `preset` support for project targets using the `library:preset` format.
- Presets are resolved from a dependency’s `manifest.json`, using the configured `dependencies` and `dependency-search-paths`.
- Preset references are validated for format and naming, the referenced library must also be listed as a dependency.
- Preset configuration is loaded before the target’s own configuration, so users can still override anything locally in `project.json`.
- Library manifests can now expose a top-level `presets` map.
- Library manifests can now expose `features`, and enabled project target features can apply additional library target configuration. Those features are uppercase constants
- Manifest loading has been factored out so presets can be resolved from both directory libraries and `.c3l` archives.
- The project schema has been extended to document the new target fields and add 2 missing ones.

I think that should be the most important things that changed.
I would really appreciate some feedback on that changes and the proposal overall :)

ManuLinares:
Would it be possible to add a few tests?

Maybe a synthetic lib or two in `resources/testproject/`, some .c3 file in there that uses (or not) the features, and some logic in `scripts/tools/ci_tests.sh` to test this in the CI (so we validate all platforms).

Jotrorox:
@ManuLinares Would that be enough testing or should I add more? If I should add more could you please provide one or two ideas of tests you'd like to see. And thanks a lot for the note :)

ManuLinares:
This looks good.

I would remove all of the weird dashes `–` (Unicode U+2013) from the codebase but that's a @lerno call

Jotrorox:
The en-Dashes are used quite a bit around the code base so I thought that they are fine. Of course I could of course remove them if wanted :) 

And yeah you are right they can look weird, but I like them quite a bit :3

> Just a few lines where they are used (the first few returned by `rg $'\u2013'`):
> lib/std/encoding/codepage.c3: 36, 39, 50
> src/compiler/sema_expr.c: 7048, 8387
> src/compiler/sema_decls.c: 422, 3713

ManuLinares:
compiler output is a tool interface, not prose, and should minimize encoding risk, ambiguity, and downstream parsing issues

Jotrorox:
I removed all occurrences of the en-Dash in my pr and I'd b e open to also create a quick pr removing the other ones in error msgs etc. If this is wanted.

Jotrorox:
A huge limitation and also a unexpected outcome of this feature was that you couldn't properly "compose" features in presets or in the project.json as the user later on. This would be fixed now but to keep the current overriding semantics I had do add new "override" fields. If that is too far or not wanted I can also revert the commit. I just don't think that it is the expected outcome that if you have 2 features that each define sources, that one just silently overrides the other.

ManuLinares:
Sorry, I don't understand, this is incomplete right?

Maybe you can explain what were you trying to achieve with examples.

Jotrorox:
I'm sorry for the incomplete/weirdly written comment.
What I was trying to fix is the behavior when multiple features are enabled for the same target.

For example, imagine a library has two optional features:
```json
{
  "features": {
    "feature_a": {
      "sources": ["src/feature_a/**"]
    },
    "feature_b": {
      "sources": ["src/feature_b/**"]
    }
  }
}
```

If a user enables both features, I think most ppl. would expect both source directories to be used:
```jsonc
{
  "features": ["feature_a", "feature_b"] // -> Expected: "sources": ["src/feature_a/**", "src/feature_b/**"]
}
```

But with the previous behavior, applying feature_b would overwrite the sources from feature_a, so only the last applied feature would remain. That means two otherwise independent features could not be composed.

The goal of this commit was to make feature application additive by default for list-like fields such as sources, c-sources, etc.

To preserve the old behavior where replacing the previous value is actually wanted, I added explicit *-override fields. So:
```json
{
  "sources": ["extra/source"]
}
```
means “append these sources”, while:
```json
{
  "sources-override": ["only/these/sources"]
}
```
means “replace the previously configured sources”.

The reason I added this is that I think additive composition is the more obvious default for features. If two enabled features both define sources or dependencies, silently dropping one of them does not seem like the expected outcome to me.

I hope this explains it a bit better :)

ManuLinares:
"make feature application additive", yes.

<del> but the added sources-override, etc, seems to add complexity, and you haven't implemented it at all from what I see? am I missing something? </del> I was missing something.

Will have a look at this at a later time, I need to see how this works in complex scenarios

lerno:
We're going to add this, but let's think about the name for now. "presets" might be not be conventional, so let's evaluate a few other alternatives:

- profiles
- configurations
- configs
- templates
- flavors
- defaults

It would look like:

```
"profile": "arduino:uno"
"configuration": "arduino:uno"
"config": "arduino:uno"
"template": "arduino:uno"
"flavor": "arduino:uno"
"recipe": "arduino:uno"
```

Jotrorox:
Thanks a lot for the info and yeah that could be an issue you are right. 

In my personal opinion `configuration` and `config` sound to generic.
`flavor` and `recipe` don't quite match, which would leave `template` or `profile`.

I personally would say `template` matches the use case quite well :)

lerno:
Let's use "template" then.

lerno:
I think `features` here is the doubtful feature. From what I understand, passing a feature flag now also implicitly modifies compilation and selection in various places. I suppose that the idea was to control the libraries by way of that. But this overloads the harmless "let's insert a constant into the source code" to be something that controls compilation. So that one has to be thought through.

I think the template is ok as it is, except change to "/". Add release notes. Remove `features` or rework it.

Jotrorox:
Okk, I personally think that `features` is a really nice feature, but you are right the name should be different. I'd propose maybe changing it to `extras` (kinda like pips `extras_require`). 
To maybe explain my reasoning behind wanting to add that flag is that it lets a library author bundle optional source files, dependencies, and build settings behind a single named switch (for example `wifi`, `bluetooth`) that the lib consumer can turn on in their own `project.json` . So optional functionality stays out of the build until it's actually wanted, without forcing libv users to hand-assemble the sources, flags, and dependencies it needs themselves.

lerno:
Yes, but as is it feels a bit bolted on, esp with features already being basically a switch. So I think the way it's presented needs to be improved somewhat.

Jotrorox:
Yeahh I kinda get what you mean, I'd remove it for this pr and look into how to properly add it

Jotrorox:
I think that should cover everything, the merge issue should also be fixed :) @lerno

Issue #3188, opened 2026-04-30
Docgen: enum/constdef values are missing


lerno:
Currently enum/constdef values are not shown.

lerno:
Still not completely done.

It should look like

```
[Params]
type  name
int   foo 

[Values]
name   ordinal   parameter
BAR      0          342
BAZ      1          333
```
For enums and for constdef:
```
[Values]
name   value
BAR    "hello"
BAZ    "world"
```

Jotrorox:
Sorry for the huge comment. But I wanted to get some feedback if that is what you mean, sorry if I'm getting your vision wrong but I tried to make it resemble it as close as possible.  

> Each example is one image of the rendered docs and the code snippet that made it happen.

<details>
<summary>
Example 1:
</summary>

```c3
<* An enum with explicit ordinal values. *>
enum HttpStatus : ushort
{
	OK,
	NOT_FOUND,
	SERVER_ERROR,
}
```

<img width="735" height="508" alt="Image" src="https://github.com/user-attachments/assets/77505972-12af-4f0f-b164-eb59e804c079" />

</details>

<details>
<summary>
Example 2:
</summary>

```c3
<* A plain enum with auto-assigned ordinals and per-value docs. *>
enum Color : int
{
	<* Represents the color red *>
	RED,
	<* Represents the color green *>
	GREEN,
	<* Represents the color blue *>
	BLUE,
}
```
<img width="734" height="608" alt="Image" src="https://github.com/user-attachments/assets/c4d8fde9-b129-4c5c-97c2-34a8aca0cab6" />

</details>

<details>
<summary>
Example 3:
</summary>

```c3
<* An enum with associated values for each constant. *>
enum Status : int (int code, String msg)
{
	OK { 0, "ok" },
	WARN { 1, "warn" },
	ERROR { 2, "error" },
}
```
<img width="736" height="646" alt="Image" src="https://github.com/user-attachments/assets/d0b5dab5-1fe1-4331-a019-d72a9733216a" />

</details>

<details>
<summary>
Example 4:
</summary>

```c3
<* A constdef with explicit values. *>
constdef Priority : int
{
	LOW = 1,
	MEDIUM = 5,
	HIGH = 10,
}
```
<img width="736" height="512" alt="Image" src="https://github.com/user-attachments/assets/662c66fb-cdb9-419b-8fe2-990c30c2b9e2" />

</details>

lerno:
[...]

PR #3190, opened 2026-04-30
Adressing #3188 - Missing enum/constdef docgen

Jotrorox:
This pull request addresses the #3188 by adding support for parsing and then converting enums and constdefs with the docgen tooling.

It supports doc Comments by putting them in the description, like it already is with functions/methods. 

> As always I'm pretty new to this code base and contributing to compilers, if there is anything I'd need to do please me up. 

### Examples:
```c
<*
 * Color represents a simple test struct to showcase that this stuff is working
 *>
enum Color : int
{
	RED,
	GREEN,
	BLUE,
	WHITE,
	BLACK
}

<*
 * Status codes used throughout no application.
 *>
constdef Status : int
{
	OK = 0,
	WARN = 1,
	ERR = 2,
	FATAL = 99
}
```
20260430_15h17m55s_grim
20260430_15h17m28s_grim

### DIFF
--- a/src/compiler/docgen.c
+++ b/src/compiler/docgen.c
@@ -34,6 +34,40 @@ static void write_expr_source_json(FILE *file, Expr *expr)
 	json_write_string(file, scratch_buffer_to_string());
 }
 
+static void write_const_value_json(FILE *file, Expr *expr)
+{
+	if (!expr || expr->expr_kind != EXPR_CONST)
+	{
+		fputs("null", file);
+		return;
+	}
+	switch (expr->const_expr.const_kind)
+	{
+		case CONST_INTEGER:
+			scratch_buffer_clear();
+			scratch_buffer_append(i128_to_string(expr->const_expr.ixx.i, 10,
+			                                     type_kind_is_signed(expr->const_expr.ixx.type), false));
+			json_write_string(file, scratch_buffer_to_string());
+			break;
+		case CONST_FLOAT:
+			scratch_buffer_clear();
+			scratch_buffer_append_double(expr->const_expr.fxx.f);
+			json_write_string(file, scratch_buffer_to_string());
+			break;
+		case CONST_STRING:
+			json_write_string(file, expr->const_expr.bytes.ptr);
+			break;
+		case CONST_BOOL:
+			json_write_string(file, expr->const_expr.b ? "true" : "false");
+			break;
+		default:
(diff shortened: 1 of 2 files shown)


ManuLinares:
Could you discard those indentation changes?

Jotrorox:
I think this should've reverted all of those formatting changes. And I'm sorry my editor just thought that formatting parts of the code with 2 space indenting would be a good Idea.

ManuLinares:
> I think this should've reverted all of those formatting changes. And I'm sorry my editor just thought that formatting parts of the code with 2 space indenting would be a good Idea.

No worries, it happens. Sorry for be nitpicky here but there are still 2 spaces indentations added weirdly like `function buildMembersHtml(vd, baseDocs) {` has 1 tab and 2 spaces following, etc...

Jotrorox:
It should all be fixed now. The formatting changes in the C file are because of clang-format and adhere to the configured .clang-format file, if I should revert them aswell I'd be glad to do so. I'm sorry for the two new commits and the force pushes, but yeah, it should all be working fine now. 

And don't worry about the nitpick; I'm still new to this codebase, and I have to learn things like this. But I think it could be easier if the .editorconfig file were extended to also have rules for the .html files.

src/compiler/docgen.c
```diff
@@ -520,19 +584,19 @@ static void emit_normal_attrs(FILE *file, Decl *decl)
...
+		if (has_attrs)                        \
+		{                                     \
+			fputs(",", file);                 \
+		}                                     \
+		else                                  \
+		{                                     \
+			fputs(",\"attributes\":[", file); \
+			has_attrs = true;                 \
+		}                                     \
+		fputs("\"@" name "\"", file);         \
```
ManuLinares:
```suggestion
#define EMIT_ATTR(flag, name)                  \
	if (flag)                                  \
	{                                          \
		if (has_attrs)                         \
		{                                      \
			fputs(",", file);                  \
		}                                      \
		else                                   \
		{                                      \
			fputs(",\"attributes\":[", file);  \
			has_attrs = true;                  \
		}                                      \
		fputs("\"@" name "\"", file);          \
```

lerno:
Thank you

lerno:
Note that it isn't completely fixed though. It should separate params and values, they are separate. They should also show the values for each enum value.

PR #3196, opened 2026-05-03
Migrating towards AVR as a enabled by default target

Jotrorox:
As the title says. It inverts the current logic around the AVR_ENABLE flag to make AVR a by default supported backend for c3.

I tested it with and without the `-DC3_AVR_DISABLE=ON` flag and the output was as expected.

> For anyone that hasn't read the chat in the discord:
> But why do this? AVR microcontrollers are probably the most widespread kind of microcontrollers there are on planet earth for beginner use, most notably the Arduino family of microcontrollers. This would add official support in the downloadable binary version of the compiler for those kinds of things. Which would in turn allow new users of the language to dive into an easy path to building things on the embedded platform they are already really comfortable with.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -130,7 +130,7 @@ set(C3_OPTIONS
         C3_LLD_DIR
         C3_ENABLE_CLANGD_LSP
         LLVM_CRT_LIBRARY_DIR
-        C3_AVR_ENABLE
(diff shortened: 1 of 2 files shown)


ManuLinares:
The CI passes, (openbsd is an external problem not related to this)

🚀️

PR #3208, opened 2026-05-07
fix/chore: resolve GCC -fanalyzer warnings

ManuLinares:


src/compiler/codegen_internal.h
```diff
@@ -20,9 +20,10 @@ static inline bool abi_type_is_valid(AbiType type);
 
 static inline LoweredType *type_lowering(Type *type)
 {
-	while (1)
+	while (type)
```
lerno: Should be asserts, not defensive programming

src/build/project_creation.c
```diff
@@ -298,9 +298,9 @@ void create_project(BuildOptions *build_options)
 	if (str_eq(build_options->project_name, "."))
 	{
 		build_options->project_name = get_cwd_project_name();
-		if (!check_name(build_options->project_name))
+		if (!build_options->project_name || !check_name(build_options->project_name))
 		{
-			error_exit("The parent directory (which is '%s') is not a valid project name.", build_options->project_name);
+			error_exit("The current directory is not a valid project name.");
```
lerno: Can even get_cwd_project_name return null? It shouldn't, that should be an early error in that case.

src/compiler/codegen_internal.h
```diff
@@ -89,13 +90,15 @@ static inline LoweredType *type_lowering(Type *type)
 				return type;
 		}
 	}
+	return type_void;
 }
 
 static inline LoweredType *type_lowering_abi(Type *type)
 {
-	while (1)
+	while (type)
```
lerno: Same with this, assert, don't loop

src/compiler/codegen_internal.h
```diff
@@ -161,11 +164,12 @@ static inline LoweredType *type_lowering_abi(Type *type)
 				return type;
 		}
 	}
+	return type_void;
 }
 
 static inline bool abi_type_match(AbiType type, Type *other_type)
 {
-	other_type = other_type->canonical;
+	other_type = other_type ? other_type->canonical : NULL;
```
lerno: You see the pattern

src/compiler/parse_global.c
```diff
@@ -2959,7 +2959,7 @@ static inline Decl *parse_func_definition(ParseContext *c, FunctionParse parse_k
 	if (!parse_func_macro_header(c, func)) return poisoned_decl;
 	if (func->name[0] == '@')
 	{
-		RETURN_PRINT_ERROR_AT(false, func, "Function names may not use '@'.");
+		RETURN_PRINT_ERROR_AT(poisoned_decl, func, "Function names may not use '@'.");
```
lerno, line 2962: First actual bug

src/compiler/parser.c
```diff
@@ -182,7 +182,7 @@ bool parse_stdin(void)
 			.dir_path = "",
 	};
 #define BUF_SIZE 65536
-	char buffer[BUF_SIZE];
+	char buffer[BUF_SIZE] = {0};
```
lerno: Horray, zero out 64 kb every time. No thank you GCC

src/compiler/parser.c
```diff
@@ -197,6 +197,7 @@ bool parse_stdin(void)
 			if (buffer == data)
 			{
 				data = malloc(capacity);
+				if (!data) error_exit("Out of memory parsing stdin.");
```
lerno: Yeah, right.

src/compiler/sema_stmts.c
```diff
@@ -3466,7 +3466,8 @@ bool sema_analyse_function_body(SemaContext *context, Decl *func)
 		context->call_env.ensures = has_ensures;
 		bool is_naked = func->func_decl.attr_naked;
 		if (!is_naked) sema_append_contract_asserts(assert_first, body);
-		Type *canonical_rtype = type_no_optional(rtype)->canonical;
+		Type *base_rtype = rtype ? type_no_optional(rtype) : type_void;
```
lerno: No, fix elsewhere don't hide the error

src/compiler/symtab.c
```diff
@@ -108,6 +108,7 @@ void symtab_init(uint32_t capacity)
 
 	size_t size = capacity * sizeof(SymtabEntry*);
 	symtab.bucket = malloc(size);
+	if (!symtab.bucket) error_exit("Out of memory initializing symtab.");
```
lerno: No

src/utils/file_utils.c
```diff
@@ -897,6 +897,7 @@ bool execute_cmd_failable(const char *cmd, char **result, const char *stdin_stri
 		free((char*)cmd);
 		cmd = scratch_buffer_to_string();
 		FILE *f = fopen("__c3temp.bin", "w");
+		if (!f) return false;
```
lerno: Error handling?

src/utils/taskqueue.c
```diff
@@ -46,6 +46,7 @@ void taskqueue_run(int threads, Task **task_list)
 		return;
 	}
 	pthread_t *pthreads = malloc(sizeof(pthread_t) * (unsigned)threads);
+	if (!pthreads) error_exit("Out of memory starting task queue.");
```
lerno: No

src/utils/taskqueue.c
```diff
@@ -102,6 +103,7 @@ void taskqueue_run(int threads, Task **task_list)
 {
 	ASSERT(threads > 0);
 	HANDLE *handles = malloc(sizeof(HANDLE) * (unsigned)threads);
+	if (!handles) error_exit("Out of memory starting task queue.");
```
lerno: No

lerno:
The calls to `malloc`: use `MALLOC` instead. Should be fine in those cases I think?

src/utils/taskqueue.c
lerno: Maybe this can't be allocated on the arena, but still, should be separately handled in a uniforma manner.

ManuLinares:
GCC -fanalyzer is hard to appease, check src/compiler/parser.c:200, but it did find two actual bugs: a bad return type in parse_global.c and a missing null-terminator in the stdin parser. I've also cleaned up the defensive code with ASSERT and MALLOC to match the project style hopefully.

PR #3209, opened 2026-05-08
[C Backend] Correctly emit string slices

matthew-nagy:
As pointed out in the compiler-development channel in the c3c discord, the c backend emits string slices incorrectly
Before example (when compiling resources/example/hash.c3)
`__c3_slice1 ___var_2 = "Hello World!";`
New output
`__c3_slice1 ___var_2 = { "Hello World!", 12 };`


matthew-nagy:
Right, fixed the portable printfing. Lets hope the tests pass on this one

PR #3211, opened 2026-05-08
[C Backend] implement function pointer type declarations

matthew-nagy:


src/compiler/c_codegen.c
```diff
@@ -168,25 +168,34 @@ static bool c_emit_type_decl(GenContext *c, Type *type)
...
-				c_emit_type_decl(c, t.type);
-			}
+			FunctionPrototype* prototype = type->function.prototype;
 			int id = ++c->typename;
-			PRINTF("typedef %s(*__c3_fn%d)(", c_type_name(c, proto->return_info.type), id);
-			FOREACH_IDX(i, ParamInfo, t, proto->param_infos)
+
+			c_emit_type_decl(c, prototype->return_result);
+
+			for(size_t i = 0; i < prototype->param_count; i++)
```
lerno: Please use `for (size_t` and not `for(size_t`, same with `while` `if` etc to follow the code standard.

PR #3221, opened 2026-05-11
chore: ignore __cmake_systeminformation

fernandolguevara:


ManuLinares:
why, what's that?

fernandolguevara:
@ManuLinares [cmake](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Source/cmake.cxx#L3795) s**t

ManuLinares:
This seems very specific to your machine. No need to add ignores.

fernandolguevara:
why? don’t we all use cmake? 

PD: make my life easy

ManuLinares:
It would be helpful to include context for this change (for example IDE/tooling that generates this file). I don't see __cmake_systeminformation in my environment, and it appears to be related to CMake/IDE-generated artifacts rather than a standard file.

In general, adding a short explanation or opening an issue to document which generated files should be ignored would help avoid ambiguity.

PR #3225, opened 2026-05-15
fix(codegen): slice len as %llu

fernandolguevara:
Fix https://github.com/c3lang/c3c/pull/3209 macos build, format slice len as %llu


lerno:
Fix already pushed.

fernandolguevara:
@lerno thanks!

PR #3231, opened 2026-05-17
[C Backend] Add if and for loop control flow

matthew-nagy:
Add code to emit if and for loops. As c3 while loops are internally recorded as for loops, it adds support for while loops too.

The backend emits all loops as while loops. This is so that expressions that take multiple lines in C can still be used as the condition and increment expressions.

I intend to do some more work on the loops when I try add support for (labeled) break/continue. My thought so far is that I could emit a label after emitting the c3 body, but before the increment and condition are emitted, which is what any continue jumps to. Other labels would be emitted for labeled breaks.

### DIFF
--- a/src/compiler/c_codegen.c
+++ b/src/compiler/c_codegen.c
@@ -330,6 +330,7 @@ static VariableId c_emit_temp(GenContext *c, CValue *value, Type *type)
 	return c->id_gen;
 }
 
+static void c_emit_stmt(GenContext *c, Ast* ast);
 static void c_emit_expr(GenContext *c, CValue *value, Expr *expr);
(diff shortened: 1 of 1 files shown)


src/compiler/c_codegen.c
```diff
@@ -330,6 +330,7 @@ static VariableId c_emit_temp(GenContext *c, CValue *value, Type *type)
 	return c->id_gen;
 }
 
+static void c_emit_stmt(GenContext *c, Ast* ast);
```
matthew-nagy: Ah, just found that this is defined earlier in the file. Will remove when I have a moment

PR #3232, opened 2026-05-17
[C Backend] Implement expression emitters commonly used with control …

matthew-nagy:
…flow

If this PR is too big I can break it down into some smaller ones, but none of this is particularly in depth except the binary expression emitter.

### DIFF
--- a/src/compiler/c_codegen.c
+++ b/src/compiler/c_codegen.c
@@ -331,6 +331,8 @@ static VariableId c_emit_temp(GenContext *c, CValue *value, Type *type)
 }
 
 static void c_emit_expr(GenContext *c, CValue *value, Expr *expr);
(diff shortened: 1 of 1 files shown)


src/compiler/c_codegen.c
```diff
@@ -402,6 +404,34 @@ static void c_emit_const_expr(GenContext *c, CValue *value, Expr *expr)
 	}
 	PRINT("/* CONST EXPR */\n");
 }
+static void c_emit_cond_expr(GenContext *c, CValue *value, Expr *expr)
+{
+	Expr **list = expr->cond_expr;
+	size_t size = vec_size(list);
```
lerno: Note that you can just use `unsigned` for all of these. `vec_size` returns unsigned.

Issue #3233, opened 2026-05-18
Bad project.json file causes segfault and computer crash


aneez-06:
## Issue

I was trying to install sdl3 through `c3c vendor-fetch sdl3` and it caused a segmentation fault.
It loads until 100% and my system just slows down for like  3 to 5 minutes and shows a segmentation fault  (core dumped) message.

I first created a project with `c3c init` and removed some unwanted directories.
And I played around with the project.json file trying to learn how the build system works and testing some things to link shared libraries and stuff.

And then i tried to install sdl3 through the `c3c vendor-fetch` command and it caused a segmentation fault.

## How to Reproduce:

```bash
c3c init test
cd test
rm -r README.md LICENSE scripts test resources docs project.json
cat <<EOF > project.json
{
    "langrev": "1",
    "warnings": [ "no-unused" ],
    "version": "0.1.0",
    "sources": [ "src/**" ],
    "build-dir": "build",
    "output": "build",
    "targets": {
        "test": {
            "type": "executable",
        }
    },
    "cpu": "generic",
    "opt": "03",
}
EOF

c3c vendor-fetch sdl3
```

## Cause
I found out that not having 
```json
"dependency-search-paths": [ "lib" ],
"dependencies": [ ],
```
these two options in the project.json file was causing the issue.

ManuLinares:
@aneez-06 could you test this now?

aneez-06:
I tested it.
And there is **no** segmentation fault now. But it is not perfect.

Not having this line `"dependency-search-paths": [ "lib" ],` now does not cause a segmentation fault, but the package is not fetched into the lib directory and is put in the root of the project.

It adds the `"dependencies": [ ],` line to the end of the json file and adds the name of the package which I fetched (eg sdl3). But `"dependency-search-paths": [ "lib" ],` is not added and the package ends up being put in the root of the project.

### Project.json file [Before]

```jsonc
{
	// ---------- Changed the order of these commands -------------
	// Warnings used for all targets.
	"warnings": [ "no-unused" ],
	// Language version of C3.
	"langrev": "1",
	// Directories where C4 library files may be found.
	// "dependency-search-paths": [ "lib" ], // ------------ commented this line for testing ---------
  // Libraries to use for all targets.
  // "dependencies": [ ], //-------------- commented this line for testing ---------
  // Authors, optionally with email.
  "authors": [ "Doe John <john.doe@example.com>" ], // --------- changed this for testing ---------
  // Version using semantic versioning.
[...]

PR #3235, opened 2026-05-18
fix: handle missing "dependencies" key in project.json

ManuLinares:
https://github.com/c3lang/c3c/issues/3233

null checks and add "dependencies" key if missing from project.json


lerno:
Thanks!

PR #3241, opened 2026-05-19
add `mmap_file` for WIN32 and httpserver example improvements

ManuLinares:
Can't add many features for this example of course

- hardened against directory traversal, and a bunch of other optimizations, it's still small and pretty.
- added description
```c3
<*
 This is a simple static HTTP server example. It listens on a TCP port, parses basic
 HTTP/1.1 requests, handles Keep-Alive connections, serves static files with Range
 and ETag support, applies dynamic Gzip compression, and displays directory listings.
*>
```


lib/std/core/os/mem_vm.c3
```diff
@@ -249,6 +250,56 @@ fn void*? mmap_file(Fd fd, sz size, sz offset = 0, VirtualMemoryAccess access =
...
+ @param shared : "if True changes are shared/written back, else private (copy-on-write)"
+ @return? mem::OUT_OF_MEMORY, RANGE_OVERFLOW, UNKNOWN_ERROR, ACCESS_DENIED, INVALID_ARGS, io::NO_PERMISSION, io::FILE_NOT_VALID, io::WOULD_BLOCK, io::FILE_NOT_FOUND
+ @return "Pointer to the mapped region"
+*>
+fn void*? mmap_file(Fd fd, sz size, sz offset = 0, VirtualMemoryAccess access = READ, bool shared = false) @if (env::WIN32)
+{
+	Win32_HANDLE h_file = (Win32_HANDLE)win32::_get_osfhandle(fd);
+	if (h_file == (void*)uptr::max) return io::FILE_NOT_VALID~;
+
+	bool is_write = access == WRITE || access == READWRITE;
```
lerno: EXECWRITE? Why not use to_win32 and tweak it?

lib/std/core/os/mem_vm.c3
```diff
@@ -249,6 +250,56 @@ fn void*? mmap_file(Fd fd, sz size, sz offset = 0, VirtualMemoryAccess access =
...
+			case win32::ERROR_INVALID_HANDLE:    return io::FILE_NOT_VALID~;
+			case win32::ERROR_NOT_ENOUGH_MEMORY: return mem::OUT_OF_MEMORY~;
+			default:                             return UNKNOWN_ERROR~;
+		}
+	}
+	defer (void)win32::closeHandle(h_map);
+
+	Win32_DWORD offset_high = (Win32_DWORD)(offset >> 32);
+	Win32_DWORD offset_low = (Win32_DWORD)(offset & 0xFFFFFFFF);
+	void* ptr = win32::mapViewOfFile(h_map, map_access, offset_high, offset_low, (Win32_SIZE_T)size);
```
lerno, line 309: What about the size = 0 case?

ManuLinares: https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-mapviewoffile

```
[in] dwNumberOfBytesToMap

The number of bytes of a file mapping to map to the view. All bytes must be within the maximum size specified by CreateFileMapping. If this parameter is 0 (zero), the mapping extends from the specified offset to the end of the file mapping.
```

For zero it says is valid, and maps from the offset to the end. I could change the contract comment to reflect this?
>  @param size : "Number of bytes to map, will be rounded up to page size"

ManuLinares: for our api consistency we could, on `size == 0` do aligned_alloc_size which does one `win32::allocation_granularity`, I don't know what's the correct path here.

ManuLinares: my understanding is that POSIX mmap requires the length argument be strictly greater than zero, we are rounding up now to 1 pagesize when `mmap_file`
maybe in posix we should also `if (size == 0) return INVALID_ARGS~;`

LowByteFox:
👀 something quite useful for me

lerno:
Thank you!

PR #3242, opened 2026-05-21
Fix --no-entry on Windows

as2te:
`--no-entry` behavior is incorrect on Windows.

The `/NOENTRY` linker arg is only valid alongside `/DLL`, to build a resource-only DLL.
It is currently added to any non-DLL build instead.

Which leads to `lld-link: error: /noentry must be specified with /dll` when trying to, e.g. build an `exe` with externally defined `main`.


lerno:
If so then I think the PR is a little incorrect: /NOENTRY shouldn't be added for DLLs either.

as2te:
Sure.

lerno:
Thank you

PR #3244, opened 2026-05-21
Support explicit pipe bindings for process::spawn

Bqrry4:
As the title say.
Could also be improved by defining an explicit Pipe object, yet not sure about this.
The tests are provided to reflect the usage.


lerno:
Looks like tests do not pass?

Bqrry4:
Fixed the test for the windows pipe.

lerno:
Thank you!

PR #3248, opened 2026-05-22
Remove dummy interface for termios

hovsater:
This is an attempt to fix some inconsistencies with the "dummy interface" that's provided by `lib/std/libc/termios.c3`. Specifically, it was added in commit b0e104bfd0d7794e01fabcfd3789c6efa0bbad18, but then changes in 487057d03c9ba6bd1b7a73811653f31a64c63114 and 5b39dc97f434fc06f2a2bbd87b573cac78aefa0b forgot to make the necessary changes to that interface as well.

With that said, I have some follow up questions:

- Is there a point in maintaining  stubs for platforms not having an implementation? Especially since the `Termios` structure is incompatible anyway?

### DIFF
--- a/lib/std/libc/termios.c3
+++ b/lib/std/libc/termios.c3
@@ -1,4 +1,4 @@
-module libc::termios @if(env::LIBC &&& env::POSIX);
+module libc::termios;
 
 fn int send_break(Fd fd, int duration) => tcsendbreak(fd, duration);
 fn int drain(Fd fd) => tcdrain(fd);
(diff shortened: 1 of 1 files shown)


hovsater:
After discussing with @lerno on Discord, we decided to remove the dummy interface altogether. See https://github.com/c3lang/c3c/pull/3248/commits/9d9acdb5e439d71ddc5730da33fef1bd18d2f346.

hovsater:
@ManuLinares and I decided that we'll merge all work into a single PR. See #3254 instead.

PR #3251, opened 2026-05-22
Move libc termios bindings from POSIX to Linux

hovsater:
**Note: Based on #3248. Merge that one first.**

This moves the current `termios` implementation out of `lib/std/libc/os/posix.c3` and into `lib/std/libc/os/linux.c3`. The definitions were not generic POSIX definitions. They match Linux/glibc specific `termios` constants, control-character indexes, and `struct termios` layout. `lib/std/libc/termios.c3` is also narrowed from POSIX to Linux so the API is only exposed for the platform it currently matches.


hovsater:
I'll rebase this branch on `master` once #3248 is merged.

hovsater:
@ManuLinares and I decided that we'll merge all work into a single PR. See #3254 instead.

PR #3253, opened 2026-05-22
Changed `Rect` edge inclusivity behavior

cmann1:
`contains_point` is now exclusive on the maximum edge


lerno:
If so, then we don't really need the fix when adding points, right?

cmann1:
You mean the changes to `@ensure return.contains_point_inclusive(point)`?

Since the added point could become the new maximum edge, and `contain_point` is now exclusive on that side, it would fail.

lib/std/math/shapes.c3
```diff
@@ -218,7 +228,7 @@ fn Rectangle Rectangle.expand(self, Num amount)
  @param point : "The point to expand the rectangle with"
  @return "The resulting rectangle"
 
- @ensure return.contains_point(point)
+ @ensure return.contains_point_inclusive(point)
```
lerno: So maybe we could make this not even a requirement and also drop the "nexhigher" part. Then have min/max with Box which is inclusive?

cmann1: Yeah I think that makes sense

cmann1:
Hopefully all of that in one PR is ok.

If you do want `Box` instead or whatever naming convention it can be changed.

lerno:
Shouldn't we just call it "Box"?

cmann1:
Sure if that's what you prefer.

`Rectangle` has the `Rect` alias as a convenience - what should the naming for box be?
`BoundingBox` for the generic base version, and `Box` for the float shorthand?
Or do you want `Box` for the generic, but then I'm not sure what the float shorthand could be.

lerno:
Thank you, and sorry for the delay!

PR #3254, opened 2026-05-22
Update `termios` bindings

hovsater:
This includes several improvements to the existing `termios` bindings.

- The dummy interface was removed as requested by @lerno.
- The existing bindings were moved from ` lib/std/libc/os/posix.c3` into ` lib/std/libc/os/linux.c3` as they were Linux specific.
- The type of the `c_cc` member in the `Termios` struct was adjusted for the Linux bindings as agreed upon with @ManuLinares.
- New Darwin specific bindings were added.


ManuLinares:
I don't know about the `alias Cc = char`, maybe just make it a `char` directly instead

hovsater:
> I don't know about the `alias Cc = char`, maybe just make it a `char` directly instead

I don't mind changing it. The reason for the alias is because that's what both Odin and Zig does (since `termios` exposes a `cc_t` type). On Linux, `Cc` is also used for multiple fields (`c_cc[NCCS]` and `c_line`).

PR #3255, opened 2026-05-22
feat(json): pretty print

fernandolguevara:
Add json pretty print

```c3
User user = { .username = "admin", .created_at = 611667900 };
String json = json::marshal(tmem, user, { .chars = "\t" });
assert(json == "{\n\t\"username\": \"admin\",\n\t\"created_at\": 611667900\n}");
```


lib/std/encoding/json_marshal.c3
```diff
@@ -10,6 +10,10 @@
 module std::encoding::json;
 import std::io;
 
+struct Indent {
+	usz spaces;
```
lerno: Better do this as an enum:
1. No pretty print
2. Tab
3. 1 space
4. 2 spaces
5. 3 spaces
6. 4 spaces
7. 5 spaces

With no pretty print, don't add `\n` either.

lerno:
Thank you!

PR #3257, opened 2026-05-23
add peekNamedPipe

Zathyy:


ManuLinares:
https://learn.microsoft.com/en-us/windows/console/getstdhandle

it's from the [Console API Functions](https://learn.microsoft.com/en-us/windows/console/console-functions)

Zathyy:
ah look at that, reverted it :)

PR #3260, opened 2026-05-23
Fix crash when reflecting certain type properties of `untypedlist`

Book-reader:
accessing the `kind`, `qname`, or `alignment` would reach unreachable statements in their respective type property functions.
image
image
image
image

`untypedlist::kind` now returns `TypeKind.UNTYPEDLIST`, `untypedlist::qname` now returns `untypedlist`, and `untypedlist::alignment` now returns `1`.

I went with `untypedlist::alignment` returning `1` instead of making it an error for consistency with `untypedlist::size` and `void::size`/`void::alignment`, lmk if this should be changed.


lerno:
Untypedlist should not return a size.

Book-reader:
I'm not sure what happened with the android CI there, but `untypedlist::size` and `untypedlist::alignment` are now disallowed.
image
image

lerno:
Thank you!

PR #3263, opened 2026-05-25
fix(compiler): init analysis locals

fernandolguevara:
a3129537604ea1f75a58c77f1fe5594ac9f463b7


lerno:
See if you can suppress the flag on the *incorrect* analysis instead. This kind of code just makes me want to cry.

PR #3264, opened 2026-05-25
Fix `std::core::test` module documentation

hovsater:
- In commit 5c77c9a754ff1796f728e214dbd0de312eb687d6 `fault` was changed to `faultdef` but fault usage was left scoped under `MathError`.
- In C3 0.6.6, the `void!` type was dropped and it states that `@test` functions should return `void` and force unwrap were necessary.


lerno:
Thank you

PR #3266, opened 2026-05-26
docgen: more fixes

ManuLinares:
- Add `--emit-stdlib=<yes|no>` to the docgen: Omits printing the stdlib from the output.

---
handle TYPE_COMPRESSED_*

example: https://c3-lang.org/standard-library/docs.html#std::thread::channel::create_one_shot

before:

```json
          "return_type": {
            "name": "OneShotChannel?",
            "uid": "std::thread::channel::OneShotChannel"
```

now:
```json
          "return_type": {
            "name": "OneShotChannel*?",
            "uid": "std::thread::channel::OneShotChannel"
          },
```

before was showing this signature in the web:
`fn OneShotChannel? create_one_shot(Allocator allocator) <Type>`
now correctly shows
`fn OneShotChannel*? create_one_shot(Allocator allocator) <Type>`

---------------

handle varargs `...`

added ` "is_vararg": true` to the json

example:
https://c3-lang.org/standard-library/docs.html#std::encoding::pem::encode_many
before it was showing this signature in the web:
`fn String? encode_many(Allocator allocator, char[][] bodies, String[] tags, PemHeader[][] pem_headers, bool use_crlf = false)`
now correctly shows
`fn String? encode_many(Allocator allocator, char[][] bodies, String[] tags, PemHeader[]... pem_headers, bool use_crlf = false)`

---

We correctly omit showing empty "contract" column for generic macro like `docs.html#std::collections::linkedlist::linked_list_alloc_node`, and on methods we are correctly showing the type of &self, as the field type.

---

---

---

TODO: We are missing showing the type for a generic method: for example for `docs.html#std::collections::linkedlist::LinkedList.last` we should be showing `LinkedList{Type}`. this could be handled in the frontend

we already have the info in the json:
```json
{
  "name": "last",
  "kind": "method",
  "uid": "std::collections::linkedlist::LinkedList.last",
  "file": "lib/std/collections/linkedlist.c3:131:21",
  "is_generic": true,
  "generic_parameters": [
    "Type"
  ],
  "return_type": {
    "name": "Type?"
  },
  "members": [
    {
      "name": "self",
      "is_ref": true
    }
  ],
  "docs": null
}
```


lerno:
Thank you!

PR #3268, opened 2026-05-26
Remove return in expr_rewrite_const_string_from_raw function

Darvisim:
Likely shouldn’t have returned a value in that function, because of which on Windows cmake build step, it shows:
```cmake
D:\a\c3c\c3c\src\compiler\expr.c(1083): warning C4098: 'expr_rewrite_const_string_from_raw': 'void' function returning a value
```


Darvisim:
Closing this, because this has been implemented in commit 0ff7564

PR #3278, opened 2026-05-29
linux: fix musl linking and warn on dynamic glibc cross-compilation

ManuLinares:
### `c3c ... -linux-libc=musl`

- warn user if they are in a glibc system compiling with `-linux-libc=musl` that they should use the builtin linker
- warn user if they are dynamically linking in a glibc system.
- warn user if they are statically linking against glibc, reminding them to use musl instead.


lerno:
Thank you

PR #3280, opened 2026-05-29
Implement block parameter name validation

salmatamil2025-alt:
Adds a function to validate block parameter names against reserved keywords.

### DIFF
--- a/My work
+++ b/My work
@@ -1,0 +1,31 @@
+// Fix compiler error message for keyword block parameter names
+
+function validateBlockParameterName(name) {
+  const reservedKeywords = [
+    "if",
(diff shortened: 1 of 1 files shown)


lerno:
Explain yourself

lerno:
Let me know why I shouldn't ban you from this repo.

hovsater:
I mean, don't even engage with this. Just ban.

lerno:
@hovsater I was thinking that maybe something was accidentally doing this. Sometimes someone sets their LLM to do more things than it should. But I already banned it actually. Just forgot to close the PR

PR #3285, opened 2026-05-30
compiler: error visibility modifiers on enum/constdef members

fernandolguevara:


lerno:
This should be an error, not a warning. Simply don't pass &visibility on enum and constdef values

fernandolguevara:
@lerno done

lerno:
Thank you

PR #3290, opened 2026-06-01
fix(compiler): tag error messages showed wrong function names; corrected to `has_tag`/`get_tag`

fernandolguevara:


lerno:
Thank you!

Issue #1403, opened 2024-09-03
Implement Reflection invocation


ygorpontelo:
Would be great to have [this](https://c3-lang.org/references/docs/anyinterfaces/#reflection-invocation) implemented.

Having interfaces helps a lot, but sometimes i just want to call the method if it exists.

lerno:
Poor feature getting postponed, but it will happen. I just want 0.6.6 to be stable for now 😄

PR #3297, opened 2026-06-03
feat(compiler): add interface introspection

fernandolguevara:
Add interface introspection support

Adds compile-time type properties for inspecting declared interfaces and checking structural interface method conformance.

New type properties:

- `Type::interfaces`: Returns the interfaces directly declared on a type.
- `Type::all_interfaces`: Returns declared interfaces recursively, including inherited parent interfaces & deduplicates.
- `Type::satisfies(InterfaceType.method)`: Checks whether a type satisfies an interface method/signature structurally.

#1403

**NOTE:**  I’m opening this as a draft because the API naming and exact semantics may need discussion, and I’m open to suggestions.


lerno:
This is a feature not discussed or previously accepted. There are very subtle considerations that need to be resolved before adding them. As you see, the problem is not adding them, it is adding them correctly. Instead of making a PR, please turn this into an issue. We can then see what should be in and not.

"Feature request disguised as PR" which is circumventing the way we usually do changes to the languages (through issues and Github discussions). You can can keep this around and make a new PR with the relevant parts that are accepted.

Issue #3300, opened 2026-06-04
Musl release links against llvm20


h8d13:
I'm guessing not a lot of people use alpine/void musl builds but thought it might be useful regardless (maybe for docker too!)

Your readme mentioned help about getting stuff running on odd Linuxes so:

> Explicit split packages = fun.

```
apk add \
llvm22-dev \
llvm22-static \
llvm22-gtest \
lld22-dev \
libxml2-dev \
curl-dev \
clang \
cmake \
samurai \
musl-dev
```

or whichever versions you want to link against.

Then: 

`cmake -B build -G Ninja -DCMAKE_BUILD_TYPE=Release`

Check for success exit. 

`cmake --build /path/to/build`

---

```
./build/c3c --version
C3 Compiler Version:       0.8.1 (Pre-release, Jun  4 2026 03:01:37)
Installed directory:       /home/alpm/memory-store/h8d13/c3c/build
Git Hash:                  1cfd9a4b773c133edce92641d87360f4f006aaaf
Backends:                  LLVM
LLVM version:              22.1.3
LLVM default target:       x86_64-alpine-linux-musl
```

Thanks for even having the musl release regardless :)

ManuLinares:
Hi, I don't understand if there is a proposal here.

https://github.com/c3lang/c3c/#compiling-on-linux

For building on musl distros (alpine, void) just `cmake -B build -S . -DC3_FETCH_LLVM=ON -DCMAKE_BUILD_TYPE=Release` because we ship our own llvm https://github.com/c3lang/llvm-for-c3/releases

Our CI does build against [llvm 20](https://github.com/c3lang/c3c/blob/master/.github/workflows/main.yml#L22) and if you see the way we install them is the same as what you wrote: https://github.com/c3lang/c3c/blob/master/.github/workflows/main.yml#L358

So I just want to know what is the goal of this issue, is there any part of the documentation that you want updated? Are any instructions wrong? If so, do tell, thanks!

h8d13:
Thanks for the details :) Just didn't find the specific docs to Alpine (I now see the deps in the CI) so thought it would be nice to put somewhere perhaps. 

If I'm not mistaken non-musl gets llvm-22+ ?

And not really an issue I guess: just saw that the main readme says that getting stuff to run on weird linuxes is appreciated :D I got the release and was surprised it used v20 instead. Which led me to build instead I guess

ManuLinares:
[...]

PR #3301, opened 2026-06-04
ci: update alpine deps

ManuLinares:
I also made an upstream request to add c3c to alpine aports (their package repository)

https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/103453

---

well I guess the compiler compiles for these architectures, so good 🤟️.
image


h8d13:
Thnx ! Closed #3300 :)

PR #3306, opened 2026-06-04
release: add MANUAL.md to the released bundles #2331

### LINKED ISSUE #2331: Offline Doc/Handbook accompany with c3c distribution package
Is it possible to provide c3c manual/doc/handbook something like [https://c3-lang.org/](url) provides in the distributionn package?Sometime it is not so convenient or even it fails to connect to the page.It does help a lot if there is an offline manual available.

ManuLinares:
from https://c3-lang.org/all.md

https://github.com/c3lang/c3c/issues/2331#issuecomment-4626620200

closes https://github.com/c3lang/c3c/issues/2331


lerno:
Thank you!

PR #3309, opened 2026-06-07
feat: add typo suggestions for types, contracts and top-level keywords

### LINKED ISSUE #1420: Improve error messages when keywords are misspelled
Currently when a keyword like `interface` are misspelled, the error is difficult to connect to a typo:

> Error: A type name was expected, but this looks a variable or function name (as it doesn't start with an uppercase letter).

This error message can be replaced by contextually checking if the token is close (see [edit distance](https://en.wikipedia.org/wiki/Edit_distance)) to a keyword.

in the case of keywords used for top level statement, since `IDENT` tokens cannot be used in global variable declarations, (as builtin types like `int` are keywords and custom types are `IDENT_TYPE`) when encountering an `IDENT` token in `parse_top_level_statement`, before calling `parse_global_declaration`, we can first go through the keywords we could expect and calculating the edit distance to the value and showing a message which explicitly suggests a typo.

R-Camacho:
- Introduce `SuggestionKeywords` to hold lists of keywords.
- Add `suggestion_init()` to populate these lists.
- Implement typo suggestions for:
- Top-level declarations;
- Base types;
- Contract parameter names.
- Move `damerau_levenshtein_distance` to `stringutils.c`.
- Add `str_find_closest` to `stringutils.c`.
- Add tests for implemented suggestions.

Closes #1420


lerno:
Pushed to 0.8.2

src/compiler/compiler.c
```diff
@@ -19,6 +19,7 @@
 
 
 CompilerState compiler;
+SuggestionKeywords suggestion_keywords;
```
lerno: Should be folded into Compiler since it's a global

src/compiler/compiler.c
```diff
@@ -45,6 +46,36 @@ static const char *out_name(void)
 
 #define START_VMEM_SIZE (sizeof(size_t) == 4 ? 1024 : 4096)
 
+static void suggestion_init()
+{
+	suggestion_keywords.base_types = VECNEW(const char *, 32);
+	suggestion_keywords.top_level_keywords = VECNEW(const char *, 32);
+
```
lerno: Should be done inside of the general symtab setup

src/compiler/sema_decls.c
```diff
@@ -3948,17 +3948,31 @@ static inline bool sema_analyse_doc_header(SemaContext *context, DeclId docs,
 			continue;
 		}
 		Decl *param = NULL;
+		const char **param_names = VECNEW(const char*, 16);
```
lerno: Memory intensive. It's not reasonable to create a dynamic array every time you do a lookup.

lerno (changes requested):

R-Camacho:
Hi @lerno, thanks for the review! We have applied your suggestions.

lerno:
I've looked at this one so many times and it just make me unsure whether it should go in or not. It does what it says, but it's essentially hacked on top of the code. So I'm going to reject this one unfortunately. Maybe it can return in some form, but not now.

PR #3312, opened 2026-06-08
macOS: a second C3 dylib loaded into a non-C3 host would crash

### LINKED ISSUE #3298: Second C3 dynamic-lib dlopen'd into a non-C3 host crashes in dl_reg_callback on macOS
_Hi! I'm not very fluent in writing english texts, so I'm sorry - I've had asked an LLM to help me polish some parts of this message._

On macOS, when two independently built C3 `dynamic-lib`s are `dlopen`'d into the same process by a **non-C3 host** (each `.so` therefore carrying its own C3 runtime), loading the **second** one segfaults inside `std::core::machoruntime::dl_reg_callback`. The first module loads and runs fine; the crash happens as the second module's runtime starts up.

This blocks using C3 to write plugins/modules for C hosts that load several of them (Tarantool in my case).

## Env

```
❯ c3c --version
C3 Compiler Version:       0.8.0
Installed directory:       /opt/homebrew/Cellar/c3c/0.8.0/libexec
Git Hash:                  unknown
LLVM version:              22.1.5
LLVM default target:       arm64-apple-darwin25.4.0
```

## Linux

Everything works fine on Linux (Archlinux rolling).

## Repro

`a.c3`:
```c3
module a;
import std::core::mem;
fn int run() @export("run") { char[] x = mem::new_array(char, 4); mem::free(x.ptr); return 1; }
```

`b.c3`: identical but `module b;` and returns `2`.

Build two separate dylibs:
```
c3c dynamic-lib a.c3      # -> liba.dylib
c3c dynamic-lib b.c3      # -> libb.dylib
```

Host `host.c` (plain C, no C3 runtime):
```c
#include <dlfcn.h>
#include <stdio.h>
int main(void) {
    void* a = dlopen("./liba.dylib", RTLD_NOW); int(*ra)(void)=dlsym(a,"run");
    printf("a=%d\n", ra());                       //  <-- WORK FINE
    void* b = dlopen("./libb.dylib", RTLD_NOW); int(*rb)(void)=dlsym(b,"run");
    printf("b=%d\n", rb());                       //  <-- CRASH
    return 0;
}
```

## Investigation

I am not completely sure, since this is my first time looking at the codebase, so:

* On Linux: Each image's `@init` constructors and `@dynamic`-method registration are emitted by the compiler straight into that image's `.init_array`. The dynamic loader runs each image's array when that image loads, in isolation.
* On macOS: per-image init slot (`__mod_init_func`), and it runs `__c3_runtime_startup`. But instead of just initializing itself, that startup calls `_dyld_register_func_for_add_image(&dl_reg_callback)` — a process-global hook. `dl_reg_callback` then walks every image in the process, reading and mutating each one's `__c3ctor`/`__c3_dynamic` tables (`type.dtable = dm`).

I've drafted a small patch that works fine for me, but I'm not completely sure that this is the right way to fix this problem (also I've failed to write this test, should they be under `ci_tests.sh`?). Tell me if it's fine to open a PR with this patch.

Basically we're saving header position and using it to iterate over constructors/destructors that are related only to current dylib.

<details>

```diff
diff --git a/lib/std/core/private/macho_runtime.c3 b/lib/std/core/private/macho_runtime.c3
index 10ebb98f1..303a17e1f 100644
--- a/lib/std/core/private/macho_runtime.c3
+++ b/lib/std/core/private/macho_runtime.c3
@@ -92,7 +92,7 @@ fn Section64*? find_section(SegmentCommand64* command, char* sectname)
 	return NOT_FOUND~;
 }
 
-macro find_segment_section_body(MachHeader* header, char* segname, char* sectname, $Type)
+macro find_segment_section_body(MachHeader* header, sz vmaddr_slide, char* segname, char* sectname, $Type)
 {
 
 	Section64*? section = find_section(find_segment(header, segname), sectname);
@@ -100,13 +100,18 @@ macro find_segment_section_body(MachHeader* header, char* segname, char* sectnam
 	{
 		return ($Type[]){};
 	}
-	$Type* ptr = (void*)header + section.offset;
+	$Type* ptr = (void*)(uptr)(section.addr + (ulong)vmaddr_slide);
 	return ptr[:section.size / $Type::size];
 }
 
-alias DyldCallback = fn void (MachHeader* mh, sz vmaddr_slide);
+extern char __dso_handle;
 
-extern fn void _dyld_register_func_for_add_image(DyldCallback);
+fn sz image_slide(MachHeader* header)
+{
+	SegmentCommand64*? text = find_segment(header, "__TEXT");
+	if (catch text) return 0;
+	return (sz)((uptr)header - (uptr)text.vmaddr);
+}
 
 
 struct DlInfo
@@ -161,7 +166,8 @@ fn void runtime_startup() @public @export("__c3_runtime_startup")
 {
 	if (runtime_state != NOT_STARTED) return;
 	runtime_state = INIT;
-	_dyld_register_func_for_add_image(&dl_reg_callback);
+	MachHeader* self = (MachHeader*)&__dso_handle;
+	dl_reg_callback(self, image_slide(self));
 	assert(runtime_state == INIT);
 	runtime_state = RUN_CTORS;
 	Callback* ctor = ctor_first;
@@ -216,36 +222,19 @@ struct TypeId
 
 fn void dl_reg_callback(MachHeader* mh, sz vmaddr_slide)
 {
-	sz size;
-	assert(runtime_state == INIT || runtime_state == READ_DYLIB, "State was %s", runtime_state);
-	foreach (&dm : find_segment_section_body(mh, "__DATA", "__c3_dynamic", DynamicMethod))
+	assert(runtime_state == INIT, "State was %s", runtime_state);
+	foreach (&dm : find_segment_section_body(mh, vmaddr_slide, "__DATA", "__c3_dynamic", DynamicMethod))
 	{
 		TypeId* type = dm.type;
 		dm.next = type.dtable;
 		type.dtable = dm;
-		DynamicMethod* m = dm;
-		while (m)
-		{
-			m = m.next;
-		}
 	}
-	foreach (&cb : find_segment_section_body(mh, "__DATA", "__c3dtor", Callback))
+	foreach (&cb : find_segment_section_body(mh, vmaddr_slide, "__DATA", "__c3dtor", Callback))
 	{
 		append_xxlizer(&dtor_first, cb);
 	}
-	foreach (&cb : find_segment_section_body(mh, "__DATA", "__c3ctor", Callback))
+	foreach (&cb : find_segment_section_body(mh, vmaddr_slide, "__DATA", "__c3ctor", Callback))
 	{
 		append_xxlizer(&ctor_first, cb);
 	}
 }
```

</details>

Thank you very much for this project, this is great language and I wish you a good luck.

bigbes:
A process-global dyld add-image hook walked every loaded image and mutated each one's `__c3ctor`/`__c3dtor`/`__c3_dynamic` tables, assuming a single C3 runtime per process; with two C3 dylibs in one host the second hook walked the first's image and segfaulted. Section pointers were also derived from the file offset instead of `vmaddr + slide`.

Initialise only our own image via `__dso_handle` (as ELF does with `.init_array`) and address sections by `section.addr + vmaddr_slide`.

Also I've spend some time to investigate if code in 7736ab391e74cf395884bd6fad695b55ca20a9f8 is even needed - looks like it's not; so I've written number of tests for both linux/macOS that tests this functionality in-depth eadd583ed35c44baa972f3480f5fa1806df11a55 on both linux and macOS.

Closes #3298

### DIFF
--- a/lib/std/core/private/macho_runtime.c3
+++ b/lib/std/core/private/macho_runtime.c3
@@ -152,7 +152,6 @@ enum StartupState
 	INIT,
 	RUN_CTORS,
 	READ_DYLIB,
-	RUN_DYLIB_CTORS,
 	RUN_DTORS,
 	SHUTDOWN
 }
@@ -222,18 +221,12 @@ struct TypeId
 
 fn void dl_reg_callback(MachHeader* mh, sz vmaddr_slide)
 {
-	sz size;
-	assert(runtime_state == INIT || runtime_state == READ_DYLIB, "State was %s", runtime_state);
+	assert(runtime_state == INIT, "State was %s", runtime_state);
 	foreach (&dm : find_segment_section_body(mh, vmaddr_slide, "__DATA", "__c3_dynamic", DynamicMethod))
 	{
 		TypeId* type = dm.type;
 		dm.next = type.dtable;
 		type.dtable = dm;
-		DynamicMethod* m = dm;
-		while (m)
-		{
-			m = m.next;
-		}
 	}
 	foreach (&cb : find_segment_section_body(mh, vmaddr_slide, "__DATA", "__c3dtor", Callback))
 	{
@@ -243,15 +236,4 @@ fn void dl_reg_callback(MachHeader* mh, sz vmaddr_slide)
 	{
 		append_xxlizer(&ctor_first, cb);
 	}
-	if (runtime_state != READ_DYLIB) return;
-	runtime_state = RUN_DYLIB_CTORS;
-	Callback* ctor = ctor_first;
-	ctor_first = null;
-	while (ctor)
-	{
-		ctor.xtor();
-		ctor = ctor.next;
-	}
-	assert(runtime_state == RUN_DYLIB_CTORS);
-	runtime_state = READ_DYLIB;
 }
--- a/lib/std/core/private/macho_runtime.c3
+++ b/lib/std/core/private/macho_runtime.c3
@@ -92,7 +92,7 @@ fn Section64*? find_section(SegmentCommand64* command, char* sectname)
 	return NOT_FOUND~;
 }
 
-macro find_segment_section_body(MachHeader* header, char* segname, char* sectname, $Type)
+macro find_segment_section_body(MachHeader* header, sz vmaddr_slide, char* segname, char* sectname, $Type)
 {
 
 	Section64*? section = find_section(find_segment(header, segname), sectname);
@@ -100,13 +100,18 @@ macro find_segment_section_body(MachHeader* header, char* segname, char* sectnam
 	{
 		return ($Type[]){};
 	}
-	$Type* ptr = (void*)header + section.offset;
+	$Type* ptr = (void*)(uptr)(section.addr + (ulong)vmaddr_slide);
 	return ptr[:section.size / $Type::size];
 }
 
-alias DyldCallback = fn void (MachHeader* mh, sz vmaddr_slide);
+extern char __dso_handle;
 
-extern fn void _dyld_register_func_for_add_image(DyldCallback);
+fn sz image_slide(MachHeader* header)
+{
+	SegmentCommand64*? text = find_segment(header, "__TEXT");
+	if (catch text) return 0;
+	return (sz)((uptr)header - (uptr)text.vmaddr);
+}
 
 
 struct DlInfo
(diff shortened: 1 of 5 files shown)


lerno:
Thank you for this.

bigbes:
Thank you for your time, @lerno!

Issue #3296, opened 2026-06-03
Compiler option to specify custom entry point in freestanding environments


rickyadastra:
Currently and to my understandings, specifying a `main` function lead to it being wrapped so that it conforms with the actual entry point function the platform expects. On the contrary, using `--no-entry` produces an executable without entry point. 

### Proposal
My proposal is to have something like `--entryfn my_main` so that:
1. the compiler would not require a proper main function and won't try wrapping it either, accepting `my_main` as is;
2. the linker would be told to about the custom entry point so that it links it correctly.

### Use cases
In general this might be useful to compile freestanding binaries without having the signature constraint on the entry point. Another even more specific is compiling EFI binaries, that require producing a proper PE (Portable Executable) with the firmware expecting a custom signature; using `--no-entry` is a solution, but it requires a separate linking pass. 

---

Alternatively, the `main` function checking and wrapping could be skipped if `--use-stdlib=no` is present, since the wrappers are defined there.

lerno:
You can just pass the entry-point directly to the linker using `-z <argument>`. This sends the argument to the linker.

rickyadastra:
It doesn't work unfortunately, I tried it already. 

The `/ENTRY` options clashes with `/NOENTRY` passed by the compiler for having `--no-entry`. And without it, the compiler complains about the missing/malformed main function.

lerno:
Try with the 0.8.1 alpha, I think that should already be fixed: https://github.com/c3lang/c3c/releases/tag/latest-prerelease-tag

rickyadastra:
Tried to build a minimal EFI application on 0.8.1 pre-release (32e8634), compiled with: 
```bash
build/c3c compile zefi.c3 
	--use-stdlib=no 
	--link-libc=no 
	--no-entry 
	--target windows-x64 
	-z /ENTRY:efi_main 
	-z /SUBSYSTEM:EFI_APPLICATION 
	-z /OUT:z.efi 
	--print-linking

# Output link arguments are:
/OUT:./out.exe /SUBSYSTEM:CONSOLE /DEBUG:FULL /tmp/c3cXTtL7X5/obj/windows-x64/zefi.obj # generated
/ENTRY:efi_main /SUBSYSTEM:EFI_APPLICATION /OUT:z.efi # custom
```

The generated binary format is correct (`PE32+ executable (EFI application) x86-64`) and so it works, as it should. My bad for not checking the latest build before asking.

[...]

PR #3319, opened 2026-06-11
Add Windows subsystem option

rickyadastra:
This PR references #3296 in regards to adding support for more Windows subsystems ([comment](https://github.com/c3lang/c3c/issues/3296#issuecomment-4626967254)).

It adds the `--win-subsystem <subsystem>` option and passes the value to the Windows linker. It also adds `win-subsystem` as a project and target option to use in project config files.

On the compiler side, it replaces `win.use_win_subsystem` with `win.subsystem`, defaulting to `"CONSOLE"`. If a valid WinMain is present, it is updated to `"WINDOWS"`; if the option is present, it is updated to that value instead.


lerno:
Since there are a limited number of subsystems, use the enum approach to set it instead. Have a default (which means autodetecting it to either Console or Windows) otherwise using what it's set to. If something is incompatible (e.g. using WinMain while setting it to Console) issue an error. Alternatively have a default which selects between Console and Windows and let the setting only give the others as options.

rickyadastra:
Replaced the string option with the enum approach. There's `WIN_SUBSYSTEM_DEFAULT` which becomes either `WIN_SUBSYSTEM_WINDOWS` or `WIN_SUBSYSTEM_CONSOLE` when the main function is analyzed in sema. If the user explicitly sets the subsystem via the option, an error is issued for incompatible combinations: WINDOWS without WinMain, and non-WINDOWS with WinMain.

There's only a couple of things that I'm not 100% convinced about:
- the help and list-project-properties descriptions get long if they list every subsystem
- if `WIN_SUBSYSTEM_DEFAULT` gets through to the linker pass ([here](https://github.com/c3lang/c3c/pull/3319/changes/ad15ed5ebe1b250450aca86cf018af892a69dd23#diff-16f296e6ddb7e1ad9f8becaa2ffd64042a850f071f39da59cdf6c097205f33a3R83), e.g if `--no-entry` is specified), no subsystem arg is appended

lerno:
I was unable to push fixes to this pull request. Moved to #3346 instead.

PR #3320, opened 2026-06-11
CI: add windows-aarch64 artifacts

ManuLinares:
Uses https://github.com/c3lang/llvm-for-c3/releases/tag/llvm_22.1.9

- stdlib doesn't properly support windows-aarch64
- No testing


lerno:
Thank you!

PR #3322, opened 2026-06-12
feat: implement array::contains_slice

AtomicSynth:
Implements `array::contains_slice` so it's possible to look up a sequence of elements in a slice/array.

### DIFF
--- a/lib/std/core/array.c3
+++ b/lib/std/core/array.c3
@@ -9,14 +9,7 @@ import std::collections::pair, std::io;
  @require @kindof(array) == SLICE || @kindof(array) == ARRAY
  @require @typematch(array[0], element) : "array and element must have the same type"
 *>
-macro bool contains(array, element)
-{
-	foreach (&item : array)
-	{
-		if (*item == element) return true;
-	}
-	return false;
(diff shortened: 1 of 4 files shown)


lerno:
I was thinking – maybe you could do "find index of slice" and then contains_slice uses that one?

lerno:
Also, check what contains do on other places I think an empty needle is always true.

AtomicSynth:
I have used implementation of `String.contains` for inspiration. It uses `@ok` macro and `String.index_of`. As for `array::index_of` it works only for an element and `array::contains` searches instead of doing same thing as `String.contains`.

Can make `array` module follow `String`.

Only `String` is implemented like this... `List.contains` searches instead of using `List.index_of`, etc.

AtomicSynth:
> I was thinking – maybe you could do "find index of slice" and then contains_slice uses that one?

Could `array::index_of_slice` work?

lerno:
Yes

AtomicSynth:
I have added `array::index_of_slice` and also matched String.contains, which just calls `@ok` on index_of.

PR #3348, opened 2026-06-26
Add Support for Dynamically Linked LLD on Linux

VisenDev:
Prior to this, I couldn't get c3c to build on nobara (a fedora based distro), because it does not ship the static (.a) versions of the needed lld libraries.


ManuLinares:
Adding ".so" dynamic libs to the ` if(NOT C3_LINK_DYNAMIC)` path is wrong. Should build in nobara with provided build steps as discussed in the issue, or `-D C3_LINK_DYNAMIC=ON`

This PR should be closed.

Issue #3350, opened 2026-06-27
Broken gzip behaviour on missing seek.


lerno:
If seek doesn't exist gzip crashes. If it exists but fails (e.g. returning a "NOT_IMPLEMENTED") then it's just ignored.

The better solution would be if incomplete deflate/inflate would in some way return the overread bytes so that they either can be pushed back or reused. I have added a test that exposes the error but it's currently disabled.

PR #3352, opened 2026-06-28
fix gzip behaviour on missing seek #3350

ManuLinares:
https://github.com/c3lang/c3c/issues/3350

plus a tiny refactoring


lerno:
Thank you

PR #3354, opened 2026-06-29
Add `types::is_nullable` and `values::@is_nullable`

guineapiguuhh:


lerno:
What do you need this for? To test if you can set it to `null`?

guineapiguuhh:
My plan is to use it together with generics for null checks

Exemple:
```c3
module maybe <Type>;
struct Maybe 
{
    Type value;
    bool has_value;
}

fn Maybe just(Type value)
{
    $if types::is_nullable(Type):
        if (value == null) assert(false);
        ...
    $else
        ...
    $endif
}
```

lerno:
Just use `$defined(value == null)`

lerno:
To expand on this: originally I had a lot of helper macros, but this requires people to figure out exactly what macros they're supposed to call as naming also isn't always clear. Using `$defined` avoids this: it's compact and obvious to the reader.

guineapiguuhh:
Wow, `$define` is really nice

guineapiguuhh:
Thanks!

PR #3355, opened 2026-06-30
Ability to fetch macOS SDK

LowByteFox:
As I've been showing in the discord, I've been working on macOS SDK fetcher, currently it's written in C3 but I'm slowly start to rewrite it to C and have it later be added to `c3c`.

Making this PR ahead of time so my progress can be tracked, suggestions given including fixes.

Unsure when this will land in `c3c` but hopefully in following releases.


src/utils/fetch_sdk/macos_sdk.c
```diff
@@ -0,0 +1,382 @@
...
+#else
+	sleep(5);
+#endif
+
+	download_file(pkm, "", metadata_out, false);
+
+	size_t size;
+	const char *content = file_read_all(metadata_out, &size);
+
+	content = strstr(content, "<pkg-info");
```
lerno, line 153: Check that content is non-null?

LowByteFox: yes... I don't check for nulls, in other places too

let me fix that 😅

lerno: I've updated some now.

src/utils/fetch_sdk/macos_sdk.c
```diff
@@ -0,0 +1,387 @@
...
+	char *end = strchr(version, '"');
+	*end = 0;
+
+	const int major = (int) strtol(version, &version, 10);
+	version++;
+	const int minor = (int) strtol(version, &version, 10);
+
+	return (Version) { major, minor };
+FAILED:
+	return (Version) { 0, 0 };
```
lerno, line 169: The code should also check for this so that it's not added.

src/utils/fetch_sdk/macos_sdk.c
```diff
@@ -0,0 +1,387 @@
...
+}
+
+static Sdk get_sdk(const char *array_tag)
+{
+	Sdk sdk = {};
+
+	char *base_pkm = strstr(array_tag, BASE_PKM);
+	char *sdk_pkg = strstr(array_tag, SDK_PKG);
+	if (!base_pkm || !sdk_pkg) return sdk;
```
lerno, line 178: I added early exit, but this means the Sdk should be checked before add

src/utils/fetch_sdk/macos_sdk.c
```diff
@@ -0,0 +1,387 @@
...
+
+	SuCatalog catalog;
+	sucatalog_init(&catalog, sucatalog_out);
+
+	const char *file_arr;
+	while ((file_arr = sucatalog_next(&catalog)) != NULL)
+	{
+		if (!strstr(file_arr, "CLTools_macOS")) continue;
+
+		const Sdk sdk = get_sdk(file_arr);
```
lerno, line 226: Check that it's not zero here

src/utils/pbzx.c
```diff
@@ -0,0 +1,96 @@
...
+		fread(in_buf, sizeof(uint8_t), min, file->file);
+
+		if (!plain && strncmp((char *) in_buf, "\xfd""7zXZ\0", 6) != 0)
+		{
+			error = true;
+			goto cleanup;
+		}
+		while (len)
+		{
+			if (plain)
```
lerno, line 56: This looks like a bug: if it's "plain" then we only do cpio_push, return to the to of the while loop, continue again. `len` is never updated.

src/utils/xar.c
```diff
@@ -0,0 +1,91 @@
...
+	dst[consumed] = 0;
+
+	const size_t off = ftell(header->file);
+	const char *start = (char*) dst;
+	while ((start = strstr(start, "<file")))
+	{
+		char *end = strstr(start, "</file");
+		*end = 0;
+
+		const char *name = strstr(start, "<name>");
```
lerno, line 66: No checks on `name` before use here

src/utils/sucatalog.c
```diff
@@ -0,0 +1,25 @@
...
+}
+
+const char *sucatalog_next(SuCatalog *catalog)
+{
+	const char *start = strstr(catalog->start, "<array>");
+	if (!start) return NULL;
+
+	start += 7;
+	char *end = strstr(start, "</array>");
+	*end = 0;
```
lerno, line 20: Possible null deref on invalid data. You can just error_exit here.

LowByteFox:
Ignore NetBSD, seems to be going through identity crisis.

src/utils/fetch_sdk/macos_sdk.c
```diff
@@ -0,0 +1,394 @@
...
+
+	if (!tmp_dir_base) error_exit("Failed to create temp directory");
+	if (verbose_level >= 1) printf("Temp dir: %s\n", tmp_dir_base);
+
+	sucatalog_out = file_append_path(tmp_dir_base, "sucatalog");
+	metadata_out = file_append_path(tmp_dir_base, BASE_PKM);
+
+	if (options->macos_list_sdks)
+	{
+		list_sdks(hardcoded, ELEMENTLEN(hardcoded));
```
lerno: This seems to create MacOSX.sdk which it leaves on error or when just listing_sdks. Is that correct?

src/utils/fetch_sdk/macos_sdk.c
```diff
@@ -0,0 +1,394 @@
...
+		{
+			const Sdk *sdk_i = hardcoded + i;
+			if (sdk_i->version.major == major && sdk_i->version.minor == minor)
+			{
+				select = i;
+				break;
+			}
+		}
+
+		if (select == -1)
```
lerno, line 296: I don't think select == -1 can ever happen, because select is set to the last element prior to this?

lerno: int64_t select = ELEMENTLEN(hardcoded) - 1;

src/utils/fetch_sdk/macos_sdk.c
```diff
@@ -0,0 +1,394 @@
...
+
+	/* target MacOSX.sdk will be moved there */
+	file_delete_dir(sdk_output);
+
+#ifdef _WIN32
+	char *sdk = str_printf(".\\SDKs\\%s", sdk_path);
+#else
+	char *sdk = realpath("SDKs/MacOSX.sdk", NULL);
+#endif
+	rename(sdk, "MacOSX.sdk");
```
lerno, line 377: Consider checking result sdk and rename. Maybe file_delete_dir should be deferred until that happens?

LowByteFox: The `CLTools_macOS_SDK.pkg` archive contains symlink to unpacked SDK from `CLTools_macOSNMOS_SDK.pkg`, on Windows this is read into memory to prevent cloning the entire SDK.

Apple groups these within a single [array in sucatalog](https://github.com/LowByteFox/c3c/blob/fetch-macossdk/src/utils/sucatalog.c#L12-L26).

The SDK is moved to prevent further copies and also unpacked on the same filesystem.

src/utils/fetch_sdk/macos_sdk.c
LowByteFox: Made sure it happens, including when dynamically fetched SDK list.

src/utils/fetch_sdk/macos_sdk.c
LowByteFox: Moved it at a later step, so it creates work dir when it's about to download and unpack everything.

LowByteFox:
Don't merge it yet, I'll try to simplify the fetching logic so it's simpler and consistent across platforms.

src/utils/fetch_sdk/macos_sdk.c
LowByteFox: I have simplified the logic, now it works more cleanly including Windows.

PR #3356, opened 2026-07-01
fix bug in unzipper.c when `c3c build`

ManuLinares:
in addition, update miniz to 3.1.2 [latest upstream](https://github.com/richgel999/miniz/releases)


lerno:
Thank you!

Issue #3358, opened 2026-07-02
[compiler] Runaway memory leak and silent crash when debug info generation is on


fesoliveira014:
Consider the following code:

```c3
module bug;

struct Device { VTable* vt; }
struct VTable { BeginFn begin; }
alias  BeginFn = fn Cmd? (Device*);
struct Cmd    { Device* device; int x; }

fn void use_cmd(Cmd* c) { (void)c; }  

fn Cmd? make(Device* d) {
    return { .device = d, .x = 1 };
}

fn void main() {
    Device d;
    use_cmd(null);
    Cmd? c = make(&d);
    (void)c;
}
```
Here we declare a vtable that has a function pointer `begin` that points to a function that returns a `Cmd` or a fault. Compiling this code with debug info turned on (`c3c compile-only bug.c3` by default does the trick) causes the compilation to segfault after running for a while and a runaway memory allocation. WSL generated a nice 27GB memory dump due to this, taking over 300GB of my SSD while I tried to debug the issue before noticing I was out of space. When we turn off debug info (`-g0`) the issue does not happen.

I am not versed in compilers and LLVM, so I used an LLM agent to help me debug the issue and find a root cause. The following was generated by one:

- c3c memoizes debug metadata per type: `type->backend_debug_type `(raw `LLVMMetadataRef`)
- Recursive structs: cache gets a temporary forward node first; real node built after members
- Struct member chain reaches `fn Cmd? (Dev*) `→ `TYPE_OPTIONAL `case (`llvm_codegen_debug_info.c:621`) caches `Cmd?` = whatever `Cmd` resolves to right now = the temporary
- Struct completes → `LLVMMetadataReplaceAllUsesWith(temp, real)` (`llvm_codegen_debug_info.c:81`) RAUWs then frees the temp
- RAUW fixes references inside LLVM's metadata graph only; c3c's raw cache pointer invisible to it → `cache[Cmd?]` dangles
- Next `fn` signature using `Cmd?` cache-hits freed node → invalid entry in `DISubroutineType` → LLVM assert (debug builds) / SIGSEGV, sometimes with runaway memory (release)
- Order-dependent: `Cmd?` first resolved outside `Cmd`'s construction → real node cached → no crash.
- -g0 fine: debug-info code never runs. line-tables still crashes: still emits function type metadata
- Affects 0.8.0 + 0.8.1

lerno:
This should be fixed now, please test.

fesoliveira014:
[...]

PR #3359, opened 2026-07-02
Add segfault handler on linux to prevent hangs caused by systemd-coredump

Book-reader:
Currently when the compiler crashes on many linux distros it will silently hang for up to a minute as the core dump is generated, this PR makes it print a proper error message and exit immediately to prevent confusion and make testing crashes faster.

tested using the example from #3358

before:
```console
$ time c3c compile-run 3358.c3
________________________________________________________
Executed in   48.13 secs    fish           external
   usr time    0.10 secs    1.27 millis    0.10 secs
   sys time    7.54 secs    0.06 millis    7.54 secs

fish: Job 1, 'time c3c compile-run 3358.c3' terminated by signal SIGSEGV (Address boundary error)
```
after:
```console
$ time c3c compile-run 3358.c3
⚠️ The compiler encountered an unexpected error: "Segmentation Fault".

- Function: segfault_handler(...)
- Source file: /build/f4pxfrxwsd94z838hhqzj6wkhlnvrp0k-source/src/main.c:37

🙏 Please consider taking the time to file an issue on GitHub (https://github.com/c3lang/c3c/issues/new), so that we can get it fixed.
________________________________________________________
Executed in  421.61 millis    fish           external
   usr time  308.86 millis    0.59 millis  308.27 millis
   sys time  109.73 millis    1.88 millis  107.85 millis
```

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -11,6 +11,9 @@
 extern bool SetConsoleCP(uint32_t codepage);
 extern bool SetConsoleOutputCP(uint32_t codepage);
 #endif
+#ifdef __linux__
+#include <signal.h>
+#endif
 
 bool debug_log = false;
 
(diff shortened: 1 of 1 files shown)


lerno:
Can you test that this works on the musl version too?

Book-reader:
seems to work there as well

image

lerno:
Thank you!

PR #3360, opened 2026-07-02
fix for test gzip_concat_nonseekable_9byte_overread() and two more tests

ManuLinares:
- add test gzip_seekable_backtrack()
- add test gzip_seekable_backtrack_1byte()
- add test gzip_read_after_close()


lerno:
Thank you!

PR #3364, opened 2026-07-04
Detect conflicting typedef redefinitions in .c3i files (#3336)

### LINKED ISSUE #3336: Compiler won't complain about redefinitions on .c3i files. C3 Compiler Version:       0.8.2 (Pre-release, Jun 13 2026 13:22:06)
```c3
// mylib.c3i
typedef MyType = void;
typedef MyType = int[<4>]; // Won't complain 
alias MyType = char; // Error: 'MyType' would shadow a previous declaration.
```
Which is an issue based on: 
> Or wait, not that is fine, it's really just in the case they're different and both in the interface file that it should be an issue. 

 _Originally posted by @lerno in [#3321](https://github.com/c3lang/c3c/issues/3321#issuecomment-4692041462)_

frapank:
Fixes #3336.

`sema_compare_weak_decl()` only validates DECL_FUNC and DECL_VAR, so DECL_TYPEDEF falls into `default: break` with no check at all - two conflicting typedefs of the same name in a .c3i just get silently merged instead of erroring.

Added a DECL_TYPEDEF case comparing the resolved underlying type (decl->distinct->type), same approach as the existing DECL_VAR check. Left DECL_TYPE_ALIAS alone since weak_anything.c3t relies on alias being overridable with an unrelated type.

I saw this is assigned to you and on the 0.8.2 milestone already, so you're probably already on it, figured I'd throw this up anyway in case it saves time. Feel free to close if redundant.


lerno:
It's something I'm actively thinking about. It makes more sense that they're not checked if the `@weak` allows it to be overwritten.

frapank:
Fixed it so an explicit @weak skips the check (that's the "allowed to be overwritten" case), while plain .c3i redefinitions still get flagged like before

lerno:
I'll reject this for now. We need a more clear way to uniformly handle .c3i vs .c3 vs extern vs weak. And at this time it can't be done well. So, we'll revisit this when we get around to the .c3i.

PR #3367, opened 2026-07-06
Set CPU for AVR architecture in target.c

### LINKED ISSUE #3366: Add CPU support for AVR to allow more advanced AVR instruction sets
The current addition of AVR support is pretty cool. In playing with it, the LLVM backend was rejecting valid ASM instructions for my ATMega chips. The LLVM backend needs to be told what CPU should be used for compilation to enable up the correct instruction set.

After some digging around, it was discovered that this could be added with a single line:
```c3
compiler.platform.cpu = compiler.build.cpu;
```

This should be added to "target.c", "target_setup" function, in the ARCH_TYPE_AVR case where the API is set. My private custom build with this change applied to the top of 0.8.1 works perfectly and I haven't noticed any other issues.

shiftylogic:
Fixes #3366.

CPU setting should come from the project.json "cpu" property for the given target.

### DIFF
--- a/src/compiler/target.c
+++ b/src/compiler/target.c
@@ -2307,6 +2307,7 @@ void target_setup(BuildTarget *build_target)
 			break;
 		case ARCH_TYPE_AVR:
 			compiler.platform.abi = ABI_AVR;
+			compiler.platform.cpu = compiler.build.cpu;
 			compiler.platform.tls_supported = false;
 			break;
 		case ARCH_TYPE_WASM32:


lerno:
Unfortunately this is not the right way to do it. Rather than setting the API directly we should be setting capabilities. You can roughly see how this works with x64 and aarch64

shiftylogic:
No worries. Given how many chips exist that you would need to cover for AVR, this is not a task I'm interested in. I'll just keep using a custom build with this hack.

NOTE: Regardless of features, etc... I do believe you will still need to pass the actual CPU used down to LLVM. Maybe I'm wrong. Someone more LLVM knowledgeable will be a better candidate to fix this.

PR #3369, opened 2026-07-08
add `std::encoding::regex` implementation

ManuLinares:
- based on https://git.musl-libc.org/cgit/musl/tree/src/regex/regcomp.c
- added tests and benchmarks against libc regex

Added `std::encoding::regex`:
- `compile`, `free`, `exec`, `matches`, `first_match`, `find_all`, `replace`, `replace_fn`, `split`, `escape`, `error_string` (and `t*` variants).


data-man:
But why it in `encoding`?

fesoliveira014:
Well, regex kinda is a type of encoding in which we look and parse certain types of patterns. Not sure what better place would fit it to be honest

data-man:
> Not sure what better place would fit it to be honest

Just `std::regex` like many other programming languages.

lerno:
```c3
fn void main()
{
	String pattern = "f";
	Regex re = regex::compile(pattern)!!;
	defer re.free();
	Match[] m = re.exec("éfé")!!;
	io::printn(m);
}
```
This prints `[�]`, because match offsets seem to slice by bytes, not codepoints.

lib/std/encoding/regex.c3
```diff
@@ -0,0 +1,3188 @@
...
+
+<*
+ Compile a POSIX regex pattern into a `Regex`.
+
+ @param pattern  : "The regex pattern string."
+ @param cflags   : "Compile flags: EXTENDED | ICASE | NOSUB | NEWLINE."
+ @param allocator : "Allocator used to persist the compiled regex. Defaults to the heap allocator."
+ @return "The compiled `Regex`, or a compile-error fault (BADPAT, ECOLLATE, ECTYPE, EESCAPE, ESUBREG, EBRACK, EPAREN, EBRACE, BADBR, ERANGE, OUT_OF_MEMORY, BADRPT)."
+*>
+fn Regex? compile(String pattern, CompileFlags cflags = {}, Allocator allocator = mem)
```
lerno: Stdlib pattern is to start with the allocator and it's non-implicit. This should be followed here as well.

lerno:
`é` as the pattern does not work at all.

lerno:
```c3
fn void main()
{
	String pattern = `\(.\)\1`;
	Regex re = regex::compile(pattern)!!;
	defer re.free();
	Match[] m = re.exec("éé")!!;
	io::printn(m);
}
```
This unexpectedly yields no match.

lib/std/encoding/regex.c3
```diff
@@ -0,0 +1,3188 @@
...
+}
+
+
+fn void? re_compile(Regex* preg, String regex, CompileFlags cflags, Allocator allocator = mem) @private
+{
+	@pool()
+	{
+		String regex_nt = regex;
+		// Ensure the pattern is null-terminated: the parser uses *s == '\0' as its stop condition.
+		if (regex.len == 0 || regex.ptr[regex.len] != '\0')
```
lerno: This is unsafe practice. If you really want zero termination, make a zero terminated copy to be safe.

lib/std/encoding/regex.c3
```diff
@@ -0,0 +1,3188 @@
...
+fn void? re_exec(Regex* preg, String string, sz nmatch, RegMatch* pmatch, ExecFlags eflags) @private
+{
+	@pool()
+	{
+		Tnfa* tnfa = preg.__tnfa;
+		if (!tnfa) return BADPAT~;
+
+		// Ensure target string is null-terminated for backtrack and parallel execution
+		String string_nt = string;
+		if (string.len == 0 || string.ptr[string.len] != '\0')
```
lerno: Same problem as before with read-after-end.

lerno:
Matching more than one group fails. It always only captures one group. However, match always returns as many slots as you set as max, it's just that those are empty(!)

lerno:
The interface as-is, lacks a lot:
find-all, interating over matches, replace, split

lerno:
Match being implicitly tmem, but then containing references to things that are allocated differently seems like a bad choice.

lib/std/encoding/regex.c3
```diff
@@ -0,0 +1,3188 @@
...
+constdef ExecFlags : int
+{
+	NOTBOL = 1 << 0,
+	NOTEOL = 1 << 1,
+}
+
+
+
+faultdef
+	OK,
```
lerno: "OK"? This seems like a bad idea to keep.

lib/std/encoding/regex.c3
ManuLinares: changed to String.zstr_copy() now

ManuLinares:
I've addressed all comments except the missing helpers (`find-all`, `replace`, `split`, and iterating over matches).

The matching regressions were a translation error on my part during the final stages of the musl port. These are now resolved, and I've added more regression tests.

Regarding the return slots and capture behavior:
- I renamed `nmatch` to `max_captures`
- By default, `max_captures = 0` dynamically detects and returns exactly the number of capture groups defined in the pattern (plus group 0).
- If the caller explicitly passes a larger number than the pattern defines, the wrapper now automatically clamps and return slice size to `re_nsub + 1`.
- Any skipped groups in the pattern still return as empty slots so the indexing doesn't break.

And will continue to work on this.

ManuLinares:
Done pushing changes

lerno:
Thank you for your work. I'm marking this experimental, because I haven't gone through the API deeply, and this allows us to change it later without having to maintain backwards compatibility.

PR #3371, opened 2026-07-09
crash fix: support ASM_ARG_ADDR for aarch64

### LINKED ISSUE #3085: "ldrp" instruction while compiling inline asm makes compiler crash
The followed snippet (compiled with `c3c run --use-stdlib=no`) threw an error

```c3
fn short key_mac() @naked {
    asm {
        sub $sp, $sp, 16;  
        mov $x0, 0;        
        mov $x1, $sp;     
        mov $x2, 1;        
        mov $x16, 3;       
        svc 0;             
        ldrb $w8, [$sp];     // <--- Without this line compiling succeeded
        add $sp, $sp, 16;
        ret;               
    }
}
fn int main()
{
    key_mac();

    return 0;
}
```

The error:

```
⚠️ The compiler encountered an unexpected error: "TODO reached".

- Function: codegen_create_aarch64_arg(...)
- Source file: /private/tmp/c3c-20260330-41267-tw3hdp/src/compiler/codegen_asm.c:118

🙏 Please consider taking the time to file an issue on GitHub (https://github.com/c3lang/c3c/issues/new), so that we can get it fixed.
```

welrox:
Fixes #3085.

# additional example
create `arg_mem.c3`:
```c3
fn int main()
{
    asm {
        nop;
        ldr $x0, [$x1];            
        ldr $x0, [$x1 + 5];
        ldr $x0, [$x1 - 5];
        ldr $x0, [$x1 + $x2];
        ldr $x0, [$x1 + $x2 * 8];
        nop;
    }

    return 0;
}
```
Compile with `c3c compile arg_mem.c3 --use-stdlib=no` (assuming aarch64 is the default target)
## before
fails to compile:
```
⚠️ The compiler encountered an unexpected error: "TODO reached".

- Function: codegen_create_aarch64_arg(...)

[...]
```

## after

compiles successfully. Running `objdump -d` on the resulting executable:

```
[...]

Disassembly of section __TEXT,__text:

0000000100000348 <_main>:
100000348: a9bf7bfd     stp     x29, x30, [sp, #-0x10]!
10000034c: 910003fd     mov     x29, sp
100000350: d503201f     nop
100000354: f9400020     ldr     x0, [x1]
100000358: f8405020     ldur    x0, [x1, #0x5]
10000035c: f85fb020     ldur    x0, [x1, #-0x5]
100000360: f8626820     ldr     x0, [x1, x2]
100000364: f8627820     ldr     x0, [x1, x2, lsl #3]
100000368: d503201f     nop
10000036c: 2a1f03e0     mov     w0, wzr
100000370: a8c17bfd     ldp     x29, x30, [sp], #0x10
100000374: d65f03c0     ret
```

No more crash, and the above sequence of assembly instructions between the two `nop`s is exactly what we would expect. With these changes, aarch64 matches the ASM_ARG_ADDR support that x86 currently has.

### DIFF
--- a/src/compiler/codegen_asm.c
+++ b/src/compiler/codegen_asm.c
@@ -115,7 +115,42 @@ static inline void codegen_create_aarch64_arg(AsmInlineBlock *block, unsigned in
 			}
 			return;
 		case ASM_ARG_ADDR:
-			TODO
+			scratch_buffer_append_char('[');
+            if (arg->base)
+            {
+                codegen_create_aarch64_arg(block, input_offset, exprptr(arg->base));
+            }
+            if (arg->offset)
+            {
+                scratch_buffer_append_char(',');
+                if (arg->neg_offset) scratch_buffer_append_char('-');
+                scratch_buffer_append_unsigned_int(arg->offset);
+            }
+            if (arg->idx)
+            {
+                scratch_buffer_append_char(',');
+                codegen_create_aarch64_arg(block, input_offset, exprptr(arg->idx));
+                scratch_buffer_append_char(',');
+                switch (arg->offset_type)
+                {
+                    case ASM_SCALE_1:
+                        scratch_buffer_append("lsl #0");
+                        break;
+                    case ASM_SCALE_2:
+                        scratch_buffer_append("lsl #1");
+                        break;
+                    case ASM_SCALE_4:
+                        scratch_buffer_append("lsl #2");
+                        break;
+                    case ASM_SCALE_8:
+                        scratch_buffer_append("lsl #3");
+                        break;
+                    default:
+                        UNREACHABLE_VOID
+                }
+            }
+			scratch_buffer_append_char(']');
+            return;
 		case ASM_ARG_MEMADDR:
 			TODO
 	}


ManuLinares:
Hi @welrox, please do read the [Contributing Guidelines](https://github.com/c3lang/c3c?tab=contributing-ov-file#5-ai-usage-in-contributions)

welrox:
Hi @ManuLinares, thank you for the comment. I have read the guidelines.

I'm not entirely sure so please let me know as I'm new here, but I suppose you want me to explain what I did. What I did for this PR, is I looked at [the already existing relevant code for x86](https://github.com/c3lang/c3c/blob/master/src/compiler/codegen_asm.c#L50C3-L85C11). I simply used that same implementation as a base and changed it so that it uses the aarch64 asm syntax. The point was mainly to fix #3085 by allowing memory operands to be used in aarch64 asm blocks.

welrox:
For more details, I spent some time debugging the compiler by setting breakpoint at the crash site [`codegen_create_aarch64_arg`](https://github.com/c3lang/c3c/blob/21e61164e8b0073f61b089fb76a0e716a7ec6f12/src/compiler/codegen_asm.c#L90) and inspecting variables while reading [`parse_asm_addr`](https://github.com/c3lang/c3c/blob/21e61164e8b0073f61b089fb76a0e716a7ec6f12/src/compiler/parse_stmt.c#L227) to understand how the mem operand is parsed

lerno:
Thanks. I'm not looking at this in detail, but until I have time with the big asm refactoring, patches like this is fine.

PR #3372, opened 2026-07-10
add `std::os::argparse` module

ManuLinares:
- fix small docs.html markdown parsing error

notes appreciated.

docs attached for easy viewing of the API only: [docs.html](https://github.com/user-attachments/files/29932632/docs.html)


lib/std/os/argparse.c3
```diff
@@ -0,0 +1,712 @@
...
+Passing a pointer to a fixed-size array collects sequential occurrences of that option flag into the array. If the array size is exceeded, the parser returns `argparse::OUT_OF_BOUNDS`.
+
+```c3
+String[3] input_files = {};
+
+argparse::@parse(args,
+	"f", "file", "Target files to process", &input_files
+);
+// Passing "-f file1.txt -f file2.txt" populates input_files[0] and input_files[1]
+```
```
fesoliveira014, line 149: Hmm, I think the most ergonomic way here would be `-f file1.txt file2.txt` instead of repeated `-f`. Ideally the parser would know through the array pointer that we are expecting a list of up to the size of the array, and either error if more than array::size elements are passed.

lib/std/os/argparse.c3
```diff
@@ -0,0 +1,712 @@
...
+If you pass a pointer to an `enum`, the parser case-insensitively maps CLI string values to the enum member names, falling back to matching integer ordinal indices.
+
+```c3
+enum LogLevel { DEBUG, INFO, WARNING, ERROR }
+LogLevel level = LogLevel.INFO;
+
+argparse::@parse(args,
+	"l", "level", "Select logging level", &level
+);
+// Values like "debug", "DEBUG", "2" are mapped automatically.
```
fesoliveira014, line 161: very neat

lib/std/os/argparse.c3
```diff
@@ -0,0 +1,712 @@
...
+```c3
+int port = 8080;
+
+argparse::@parse(args,
+	"p", "port=env:PORT", "Port number (defaults to $PORT)", &port
+);
+```
+
+### 2. Custom Validation Callbacks
+You can pass a callback function of type `fn void? (String)` as an option assignee. The callback is invoked with the parsed option value. Returning a fault from this function aborts option parsing and propagates the fault out of `@parse`.
```
fesoliveira014: this is very useful, I don't see this in a lot of argparsers

lib/std/os/argparse.c3
```diff
@@ -0,0 +1,712 @@
...
+By default, non-boolean variables require an argument (e.g. `--port 8080`). Appending `?` to the option name makes its argument optional. If the flag is provided without an argument, boolean targets are set to `true`, and other types retain their defaults or are parsed if a value is present.
+```c3
+"v?", "verbose?", "Toggle verbose output", &verbose
+```
+
+### 3. Incremental Counters (`+`)
+Appending `+` to an option name increments the target numeric variable every time the flag is specified.
+```c3
+"v+", "verbose+", "Increase verbosity", &verbosity
+// Passing "-v -v --verbose -v" results in verbosity = 4
```
fesoliveira014: I think this also suffers from the same issue as array one, most people would expect this to accept `-v, -vv, -vvv, etc` instead of `-v` repeated N times.

fesoliveira014 (commented):
overall the design is very nice! I left some comments on ergonomics

lib/std/os/argparse.c3
ManuLinares: this already works:
```
$ c3c compile-run test/unit/stdlib/os/argparse.c3 -- -l 2 as -vv  a sdasd -v -vvvvv -f asd Aas -67

Program linked to executable './argparser_test'.
Launching ./argparser_test -l 2 as -vv a sdasd -v -vvvvv -f asd Aas -67

Verbosity level: 8
Scale count:     1.000000
App name:        
Logger level:    WARNING
Target files:    asd, , 
Parsed Positional Arguments: 
as a sdasd Aas -67 

```

lib/std/os/argparse.c3
ManuLinares: Most command-line parsing libraries avoid this pattern because it introduces several critical ambiguities and fragile behaviors. You can still do this using the low level API or simply using:
```
### 6. Positional Arguments Capture
If a final trailing argument (not part of a triplet or quartet) is provided to `@parse`, it acts as a positional target array. Unhandled positional arguments are stored there in order.

```c3
String[10] positional_args = {};

argparse::@parse(args,
	"v", "verbose", "Verbose mode", &verbose,
	&positional_args // Stores up to 10 positional arguments
);

// and you parse positional_args[]
```

```

File: lib/std/os/argparse.c3 (hunk 4 above)
Comment (ManuLinares) [reply lib/std/os/argparse.c3]: I've added this to the documentation:
```
The parser supports both standard stacked short flags (e.g., `-vvv`) as well as repeated occurrences of flags.
```

File: lib/std/os/argparse.c3 (hunk 1 above)
Comment (fesoliveira014) [reply lib/std/os/argparse.c3:149]: Gotcha, that is fair. Having the option to do it with low level lib is good enough!

Comment (ManuLinares) [reply lib/std/os/argparse.c3:149]: 🤘 Trade-offs, I've updated the documentation a bit

**Comment by ManuLinares** [comment]
I guess this needs a review and a name change.

```
std::cli::args;
std::cli::flags;
std::env;
std::args
std::flags;
std::io::args;
```

fesoliveira014 (approved):
looks really good, thanks for this!

ManuLinares:
I like the API and it's simplicity but I'll put in in draft so other APIs can be considered.

ManuLinares:
https://github.com/ManuLinares/argparse

PR #3379, opened 2026-07-14
Fix timed `tcp::connect` to use `getsockopt(SO_ERROR)` for the connect result

omarluq:
`connect_with_timeout_from_addrinfo` subscribed the poll to `POLLOUT` but tested the result against `POLLWRNORM`, a different bit, so a completed non-blocking connect was never recognized. Every timed connect fell through to `os::socket_error()`, which read the stale `EINPROGRESS` `errno` and mapped it to `io::GENERAL_ERROR` so a timed connect never returned a socket and never surfaced the real failure.

The fix reads the actual result with `getsockopt(SO_ERROR)` after poll: 0 means connected, otherwise convert the error (e.g. `ECONNREFUSED` → `CONNECTION_REFUSED`).


ManuLinares:
doesn't this skips any remaining addresses in addrinfo?

omarluq:
> doesn't this skips any remaining addresses in addrinfo?

yes on a definite connect error it return right away instead of trying the remaining addresses i went with return on first error for simplicity happy to change it to record the error and `continue` then return the last error if you prefer that

omarluq:
seems the new test surfaced an unrelated pre-existing bug there's no OpenBSD `Errno` table so it just inherits Linux's errno numbers (`ECONNREFUSED` is 111 there vs OpenBSD's real 61) so OpenBSD socket errors get mis-decoded to `io::GENERAL_ERROR` no matter what this fix does

this is probably broader than one constant since OpenBSD has no errno table at all so happy to do a separate PR adding one with correct values if the core team wants that or glad to leave it to you and just file an issue whatever you prefer

lerno:
Thank you for this.

PR #3381, opened 2026-07-16
Improve error message for invalid self parameter usage

FullStackDev200:
Improve the compiler diagnostic emitted when & is used on a non-method parameter.

Previously the compiler only reported:

```
Self parameters are only allowed on methods.
```

The new diagnostic also suggests the likely fix by recommending Type* instead of Type&, making the function non-method:

```
Self parameters `&` are only allowed on methods. Did you mean perhaps want '*arg'?
```

### DIFF
--- a/src/compiler/sema_decls.c
+++ b/src/compiler/sema_decls.c
@@ -1223,7 +1223,7 @@ static inline bool sema_analyse_signature(SemaContext *context, Signature *sig, 
 	{
 		if (!method_parent)
 		{
-			RETURN_SEMA_ERROR(params[0], "Self parameters are only allowed on methods.");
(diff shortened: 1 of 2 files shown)


src/compiler/sema_decls.c
```diff
@@ -1223,7 +1223,7 @@ static inline bool sema_analyse_signature(SemaContext *context, Signature *sig,
 	{
 		if (!method_parent)
 		{
-			RETURN_SEMA_ERROR(params[0], "Self parameters are only allowed on methods.");
+			RETURN_SEMA_ERROR(params[0], "Self parameters `&` are only allowed on methods. Did you mean perhaps want '*%s'?", params[0]->name);
```
lerno: "Did you mean perhaps want" -> "Did you perhaps mean" or "Did you perhaps want"

PR #3387, opened 2026-07-20
Add RFC 3986 url parse tests

omarluq:
@ManuLinares following up on our Discord chat: this PR adds RFC 3986 URL parse tests to `test/unit/stdlib/net/url.c3`, using cases from peer libraries (Ruby's `Addressable`, Go's `net/url`, Python's `urllib`)

Most pass, but 5 fail:

<table>
<tr>
 <th>Input
 <th>Expected
 <th>Result
 <th>Cause
<tr>
 <td><code>//example.com</code>
 <td><code>host='example.com'</code>, <code>path=''</code>
 <td><code>host=''</code>, <code>path='//example.com'</code>
 <td>authority is gated on <code>scheme != "urn"</code> instead of a leading <code>//</code>, so a schemeless <code>//host</code> is parsed as path
<tr>
 <td><code>//user@example.com/path?a=b</code>
 <td><code>username='user'</code>, <code>host='example.com'</code>, <code>path='/path'</code>
 <td><code>username=''</code>, <code>host=''</code>, <code>path='//user@example.com/path'</code>
 <td>same authority gate: no scheme means the <code>//</code> authority is never parsed
<tr>
 <td><code>//example.com:80</code>
 <td><code>host='example.com'</code>, <code>port=80</code>
 <td><code>host=''</code>, <code>port=0</code>, <code>path='//example.com:80'</code>
 <td>same authority gate
<tr>
 <td><code>http://@example.com/</code>
 <td><code>host='example.com'</code>, <code>username=''</code>
 <td>faults <code>INVALID_USER</code>
 <td>empty userinfo hits <code>if (!username.len) return INVALID_USER~</code>, but RFC 3986 §3.2.1 permits it
<tr>
 <td><code>http://:@example.com/</code>
 <td><code>host='example.com'</code>
 <td>faults <code>INVALID_USER</code>
 <td>same empty-userinfo guard (empty user and password)
</table>

Happy to send the fixes too if you'd rather they go in here

### DIFF
--- a/lib/std/net/url.c3
+++ b/lib/std/net/url.c3
@@ -100,7 +100,6 @@ fn Url? parse(Allocator allocator, String url_string)
 			{
 				String[] userpass = userinfo.tsplit(":", 2);
 				username = userpass[0];
-				if (!username.len) return INVALID_USER~;
 				url.username = decode(allocator, username, HOST) ?? INVALID_USER~!;
 				if (userpass.len > 1) url.password = decode(allocator, userpass[1], USERPASS) ?? INVALID_PASSWORD~!;
 			};
@@ -185,7 +184,7 @@ fn sz? Url.to_format(&self, Formatter* f) @dynamic
 	if (self.host.len > 0) len += f.print("//")!;
 
 	// Add username and password if they exist
-	if (self.username)
+	if (self.username || self.password)
 	{
 		@stack_mem(64; Allocator smem)
 		{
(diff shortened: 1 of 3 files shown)


ManuLinares:
Assuming the tests are correct, could you use lib/std/core/test.c3 functions/macros instead of asserts?

and if you can, most minimal fix would be fine, if not these tests could be added as `@if(false)` until another PR or fix is proposed of course.

omarluq:
> Assuming the tests are correct, could you use lib/std/core/test.c3 functions/macros instead of asserts?
> 
> and if you can, most minimal fix would be fine, if not these tests could be added as `@if(false)` until another PR or fix is proposed of course.

I applied minimal fixes for the failing cases

lerno:
Thank you.

PR #3391, opened 2026-07-21
Change the c function to be invoked by _erff: erf -> erff

### LINKED ISSUE #3390: Float overload of the error function is broken
Calling `math::erf` with a float argument (or calling `math::_erff` directly) results in erratic behavior.

Image

I plotted the function from -2 to 2 in the image above.

# Solution

The issue is that the implementation is calling the C erf function (double by default) but passing a float without converting. This can be fixed by simply changing

```c3
extern fn float _erff(float x) @MathLibc("erf");
```

to

```c3
extern fn float _erff(float x) @MathLibc("erff");
```

in line 931 of `math.c3`.

Darvisim:
Closes: #3390

The `_erff` function was calling the C `erf` function (which expects a `double`) instead of `erff` function (which expects a `float`) which caused an erratic behaviour whenever a `float` was passed to the `erf` macro.

### DIFF
--- a/lib/std/math/math.c3
+++ b/lib/std/math/math.c3
@@ -928,7 +928,7 @@ extern fn double _erf(double x) @MathLibc("erf");
 extern fn double _lgamma(double x) @MathLibc("lgamma");
 extern fn double _tgamma(double x) @MathLibc("tgamma");
 
-extern fn float _erff(float x) @MathLibc("erf");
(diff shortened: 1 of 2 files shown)


lerno:
Can you add a releasenote for it.

lerno:
Perfect, thank you!

PR #3399, opened 2026-07-25
Fix Url parser accepting control characters in the host

omarluq:


ManuLinares:
isn't this part of https://github.com/c3lang/c3c/pull/3387/ ?

omarluq:
> isn't this part of https://github.com/c3lang/c3c/pull/3387/ ?

both bugs are in url parser but functionally unrelated, I think it can be a part of that PR or independent, happy to be fold it in if you prefer that

PR #3415, opened 2026-08-03
fixes needed for building c3c with emcmake (wasm32-emscripten)

ManuLinares:
This is so I can build the playground [(live here)](https://manulinares.github.io/c3-playground/)

source code: https://github.com/ManuLinares/c3-playground/

### DIFF
--- a/src/compiler/compiler.c
+++ b/src/compiler/compiler.c
@@ -751,7 +751,7 @@ void compiler_compile(void)
 		{
 			const char *cc = compiler.build.cc ? compiler.build.cc : default_c_compiler();
 			if (!file_executable_in_path(cc)) system_linker_available = false;
-			if (compiler.platform.os == OS_TYPE_EMSCRIPTEN && (!file_executable_in_path(cc) || !strstr(cc, "emcc")))
+			if (compiler.platform.os == OS_TYPE_EMSCRIPTEN && compiler.build.linker_type != LINKER_TYPE_BUILTIN && (!file_executable_in_path(cc) || !strstr(cc, "emcc")))
 			{
 				error_exit("\"emscripten\" target requires Emscripten to be installed on your system; \"%s\" was not found.", cc);
 			}
--- a/src/utils/vmem.c
+++ b/src/utils/vmem.c
@@ -97,7 +97,8 @@ static inline void* mmap_allocate(Vmem *vmem, size_t to_allocate)
 	return ptr;
 }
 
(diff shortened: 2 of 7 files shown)


src/utils/vmem.c
```diff
@@ -97,7 +97,8 @@ static inline void* mmap_allocate(Vmem *vmem, size_t to_allocate)
 	return ptr;
 }
 
-size_t max = 0x10000000;
+// 32-bit / WASM has a 2GB total address limit; cap per-arena virtual memory to 128MB.
+size_t max = sizeof(size_t) == 4 ? 128 : 0x10000000;
```
lerno: Better restrict this to WASM

src/compiler/compiler.c
```diff
@@ -751,7 +751,7 @@ void compiler_compile(void)
 		{
 			const char *cc = compiler.build.cc ? compiler.build.cc : default_c_compiler();
 			if (!file_executable_in_path(cc)) system_linker_available = false;
-			if (compiler.platform.os == OS_TYPE_EMSCRIPTEN && (!file_executable_in_path(cc) || !strstr(cc, "emcc")))
+			if (compiler.platform.os == OS_TYPE_EMSCRIPTEN && compiler.build.linker_type != LINKER_TYPE_BUILTIN && (!file_executable_in_path(cc) || !strstr(cc, "emcc")))
```
lerno, line 754: Will it work with linker=builtin?

ManuLinares: nope, unless the user provides their own SDK like the emscripten libc sdk but at that point just use `emcc`

lerno: So shouldn't it be an error regardless? Does it need the check?

ManuLinares: My thinking is, 

In the playground we have a virtual filesystem. Right now c3c is trying to find `emcc` which in the browser it fails to do so:

image

So I thought two options.

1. Mocking a dummy `emcc` in the virtual filesystem so this check passes (which I currently don't do)
2. Allowing the option `--linker=builtin` for target `emscripten`

---

I prefer option (2) because it will allow us maybe ship the "C3 playground" SDK (libc), which is very small and doesn't need emcc at all. But this was just an idea, 

I think we can also make this check, and give a Warning instead of an Error, thus removing the hard requirement of having emcc in the path

lerno:
Thank you!

PR #3424, opened 2026-08-08
feat: Adds `lidt` and `lgdt` instructions to inline assembly for x86 targets 

ssimb:
Adds the `lgdt` and `lidt` instructions support to inline assembly for x86 targets


lerno:
Please add a mention of the addition to the release notes, and then I can merge it.

lerno:
Thank you!

PR #3425, opened 2026-08-08
Add `range::slice` macro

m0tholith:


lerno:
I don't know if range::slice(array, range) might be better?

lerno:
Ping

lerno:
You can extract the check to be external (the one that it's an ExclusiveRange/Range) and place it in the macro `@require` instead. And add to the releasenotes?

lerno:
Preconditions should also check that the start >= 0 and the end is < len

m0tholith:
This new version splits the macro into `range::slice` and `range::exclusive_slice`, with bounds checking and type checking. I found that if the two macros were one, the contract would be too confusing for anybody to read, so I split them. If this looks good I'll merge everything with release notes

lerno:
You should test it for slicing rather than indexing `array[:0]` rather than`array[0]`. Add releasenotes and I'll take it from there.

ManuLinares:
why does the tests say `	int[] array = { 5, 6, 9, 1, 2, 8, 4 };`? isn't `int[]` a slice?

lerno:
Thank you

PR #3429, opened 2026-08-09
fix: write back the hi-offset GEP in llvm_coerce_expand_hi_offset

### LINKED ISSUE #3428: [riscv] FP+int structs corrupted when passed or returned by value
### Summary

On RISC-V targets (riscv64 verified, riscv32 shares the code path), a struct
containing one floating-point and one integer field (e.g. `{ float f; int i; }`)
has its integer member zeroed when the struct is passed to or returned from a
function by value. The float member is correct.

This is the RISC-V psABI "FP+integer pair" case: such structs are passed /
returned in one FPR and one GPR, and c3c models that with
`ABI_ARG_EXPAND_COERCE`.

### Versions

- c3c 0.8.2 (installed), and current master (`fe559eace`) - same code path.

### Minimal repro

```c3
// exit 0 = correct, exit 1 = make_pair().i wrong, exit 2 = take_pair(p) wrong
module pair_abi;

struct Pair
{
	float f;
	int i;
}

fn Pair make_pair()
{
	Pair p = { .f = 3.5f, .i = -19 };
	return p;
}

fn int take_pair(Pair p)
{
	return p.i;
}

fn int main(String[] args)
{
	Pair p = make_pair();
	if (p.i != -19) return 1;
	if (take_pair(p) != -19) return 2;
	return 0;
}
```

Expected on riscv64: exit 0
Actual: exit 1 (`make_pair().i` comes back 0; the same corruption applies to
by-value parameters, checked by `take_pair`).

The same code exits 0 on x86-64. A C equivalent compiled with clang
`--target=riscv64-linux-gnu` also passes, so the bug is in c3c's codegen, not
LLVM. Full repro + a qemu verification script are attached to this issue.

### Root cause

`llvm_coerce_expand_hi_offset` in `src/compiler/llvm_codegen_expr.c` computes
the GEP that advances the address to the `hi` member of an
`ABI_ARG_EXPAND_COERCE` pair, but **discards the result** - `*addr` is never
updated. All four call sites (parameter store/load, return store/load, and the
two call-argument paths) then read/write the `hi` value at offset 0, i.e. on
top of the `lo` member.

The generated LLVM IR shows it directly (dead GEP):

```llvm
%ptradd2 = getelementptr inbounds i8, ptr %p, i64 4
%1 = load i32, ptr %p, align 4            ; should be ptr %ptradd2
```

### Suggested fix

```c
*addr = llvm_emit_const_ptradd_inbounds_raw(c, *addr, type_size(info->coerce_expand.hi) * info->coerce_expand.offset_hi);
```
and the same `*addr =` for the packed branch. See the PR.

### Impact

Only RISC-V is affected (`abi_arg_new_expand_coerce_pair` is called only from
`src/compiler/abi/c_abi_riscv.c`). Any C3 code that passes or returns
mixed FP/int structs by value (e.g. per-tensor quantization parameters) is
silently corrupted on RISC-V.

storious:
# PR: fix riscv64 ABI corruption of FP+int structs passed/returned by value

Fixes #3428.

## Problem

`llvm_coerce_expand_hi_offset` computes the address of the `hi` member of an
`ABI_ARG_EXPAND_COERCE` pair but drops the GEP result, so `*addr` keeps
pointing at `lo`. All four call sites then place/read the `hi` value at
offset 0. On RISC-V this corrupts any `{float, int}`-style struct passed or
returned by value (the integer member arrives as 0).

## Fix

```diff
 	if (info->coerce_expand.packed)
 	{
 		*align = type_min_alignment(*align, *align + info->coerce_expand.offset_hi);
-		llvm_emit_const_ptradd_inbounds_raw(c, *addr, info->coerce_expand.offset_hi);
+		*addr = llvm_emit_const_ptradd_inbounds_raw(c, *addr, info->coerce_expand.offset_hi);
 		return type2;
 	}
 	*align = type_min_alignment(*align, *align + llvm_store_size(c, type2) * info->coerce_expand.offset_hi);
-	llvm_emit_const_ptradd_inbounds_raw(c, *addr, type_size(info->coerce_expand.hi) * info->coerce_expand.offset_hi);
+	*addr = llvm_emit_const_ptradd_inbounds_raw(c, *addr, type_size(info->coerce_expand.hi) * info->coerce_expand.offset_hi);
 	return type2;
```

## Verification

- Minimal repro (`pair_abi.c3`) fails on riscv64 before the fix
  (`make_pair().i = 0`) and passes after (`ok`), verified under qemu-riscv64.
- The same repro passes on x86-64 before and after (no regression).
- clang-compiled C equivalent passes on riscv64, confirming the issue is in
  c3c's codegen, not LLVM.
- Full rvinfer (an inference runtime exercising this ABI path with
  quantization structs) now produces correct results on riscv64.

## Scope

Only `c_abi_riscv.c` uses `abi_arg_new_expand_coerce_pair`, so this affects
RISC-V targets only.

### DIFF
--- a/releasenotes.md
+++ b/releasenotes.md
@@ -40,7 +40,6 @@
 - ThreadGroup with function returning `void` was broken.
 - Timed `tcp::connect` always failed with `io::GENERAL_ERROR` instead of the real result.
 - Compile time struct with zeroed union member access causes compiler error #3382.
-- RISC-V structs with mixed FP and integer fields were corrupted when passed or returned by value. #3428
 - Generic methods checked before the generic type is fully registered.
 - Math function `_erff` invoked C `erf` function instead of `erff` function #3391
 - Defining local constants inside a macro causes it to fail to @const fold. #3397
@@ -57,6 +56,7 @@
 - Several uses of InStream didn't properly handle io::EOF.
 - Fixes to memory handling during zip loading.
 - Multireader reading after a final empty read would crash.
+- RISC-V structs with mixed FP and integer fields were corrupted when passed or returned by value. #3428
 
 ## 0.8.2 Change list
 
--- a/releasenotes.md
+++ b/releasenotes.md
@@ -38,6 +38,7 @@
 - ThreadGroup with function returning `void` was broken.
 - Timed `tcp::connect` always failed with `io::GENERAL_ERROR` instead of the real result.
 - Compile time struct with zeroed union member access causes compiler error #3382.
+- RISC-V structs with mixed FP and integer fields were corrupted when passed or returned by value. #3428
 - Generic methods checked before the generic type is fully registered.
 - Math function `_erff` invoked C `erf` function instead of `erff` function #3391
 - Defining local constants inside a macro causes it to fail to @const fold. #3397
--- a/src/compiler/llvm_codegen_expr.c
+++ b/src/compiler/llvm_codegen_expr.c
@@ -442,11 +442,11 @@ LLVMTypeRef llvm_coerce_expand_hi_offset(GenContext *c, LLVMValueRef *addr, ABIA
 	if (info->coerce_expand.packed)
 	{
 		*align = type_min_alignment(*align, *align + info->coerce_expand.offset_hi);
-		llvm_emit_const_ptradd_inbounds_raw(c, *addr, info->coerce_expand.offset_hi);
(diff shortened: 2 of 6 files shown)


lerno:
Please add a mention in the releasenotes as well.

lerno:
Thank you.

PR #3430, opened 2026-08-10
feat(range): add to_array methods

fernandolguevara:


lerno:
What is this used for?

fernandolguevara:
@lerno 
I use this for plot axis values, I also need kind of steprange where next value isn't +1 logic, ex: range that starts at 1/1/1990, ends 1/1/1991, step 1 week

lerno:
Do you need to create a range first, or could it be better be `range::array(mem, from, to)`?

fernandolguevara:
I prefer to pass around a Range and expand it when it necessary

lerno:
Thank you

PR #3434, opened 2026-08-12
Introduce C3_ARM_DISABLE to allow building without ARM backend support

WeirdGumi:


lerno:
Why?

lerno:
I'm just going to close this because there is no rationale or discussion. In general, I don't want to encourage platforms where the maintainers of LLVM decide in their almighty wisdom that only a few platforms are worthy of inclusion.

WeirdGumi:
> Why?

Some distributions provide LLVM builds that do not include all targets.

https://github.com/eweOS/packages/blob/llvm/PKGBUILD#L237

lerno:
This is why we offer our own packaged LLVM libraries. There is no need to use broken LLVM packages.

PR #3437, opened 2026-08-13
Stack features options and attributes

rickyadastra:
This PR adds build options and attributes to set stack argument probing and stack protector levels. 
- `--stack-protector=<option>` sets the stack smashing protector level, available options are:
  - `none`, disable stack protector
  - `basic` (default), enable stack protector as needed
  - `strong`, enable stack protector as needed with stricter rules
  - `all`, enable stack protector for all functions
- `--stack-probe=<option>` sets the stack argument probing level, available options are:
  - `none`, disable stack probing
  - `call` (default), enable standard stack probing by calling host-defined functions
  - `inline`, enable stack probing by inlining probe routines in the function prologue
- `--stack-probe-size <number>` sets the minimum stack size to enable stack probing for a function

It also adds the following attributes for functions and lambdas:
- `@stackprotector(level)` sets the stack protector level, accepts the same values as `--stack-protector`
- `@stackprobe(level)` sets the stack probe level, accepts the same values as `--stack-probe`
- `@nostackprotector` is a shorthand for `@stackprotector("none")`
- `@nostackprobe` is a shorthand for `@stackprobe("none")`

---
~Though I'm not 100% about my implementation of the new attribute data, especially of where I [added them](https://github.com/c3lang/c3c/blob/8d4a531f59252836636b8f674215e4e8e7313c1f/src/compiler/compiler_internal.h#L314-L315).~ Moved [here](https://github.com/rickyadastra/c3c/blob/196e4421bfce3aaa05412d41d7f8802bc11f04f0/src/compiler/compiler_internal.h#L412-L413), it should be better.


lerno:
Will you please update the releasenotes with this as well?

rickyadastra:
Done, I rebased onto master as well.

lerno:
The generated code is wrong for MacOS at least. This is what clang produces on Godbolt:

```
attributes #0 = { noinline nounwind optnone ssp uwtable(sync) "frame-pointer"="non-leaf-no-reserve" "no-trapping-math"="true" "stack-protector-buffer-size"="8" "target-cpu"="apple-m1" "target-features"="+aes,+altnzcv,+ccdp,+ccidx,+ccpp,+complxnum,+crc,+dit,+dotprod,+flagm,+fp-armv8,+fp16fml,+fptoint,+fullfp16,+jsconv,+lse,+neon,+pauth,+perfmon,+predres,+ras,+rcpc,+rdm,+sb,+sha2,+sha3,+specrestrict,+ssbs,+v8.1a,+v8.2a,+v8.3a,+v8.4a,+v8a" "tune-cpu"="apple-m5" }
```

As you see, the stack-protector-buffer-size is there, but in the test cases they're not. You can see this in user_defined_attributes.c3t for example. I suspect that the correct way to know the defaults is to look through the clang source code.

rickyadastra:
`user_defined_attributes.c3t` and the other tests changed in d8f6ccd8f66c67a8c7847f172c794dcc6542ae59 didn't have stack protector enabled prior. The only codegen change I made was to output `"stack-protector-buffer-size"` only when `ssp` is enabled, as it's ignored otherwise (see [here](https://llvm.org/doxygen/StackProtector_8cpp_source.html#l00439), L439-453). 

Currently, `ssp` is inserted only when backtrace is enabled, while `"stack-protector-buffer-size"` is output regardless.
https://github.com/c3lang/c3c/blob/67536edf13bb7b311bfe4f965d9172599a2da6ee/src/compiler/llvm_codegen.c#L1246-L1251 
With [these changes](https://github.com/c3lang/c3c/pull/3437/changes#diff-e1e04ae0247dd7c5b4b725833bc6534c2a513b33b61e956fdc89fce9c6acf855L1249-L1251), the `ssp` attribute can be set on or off (or changed to either `sspstrong` or `sspreq`) but the default behavior remains unchanged: if backtrace is enabled, enable `ssp` as well, and `"stack-protector-buffer-size"` becomes tied to it.

lerno:
It is wrong. ssp should be on regardless of stacktrace requirements, at least on MacOS.

lerno:
You can investigate with godbolt what's the right thing on other platforms. Or use Clang's source

rickyadastra:
Got it now.

What about safe mode though? Should ssp depend on it or should it just stick with platform defaults unless explicitly set?

lerno:
Safe mode is just controlling inserts of checks.

lerno:
And with that I mean things like null checks, array overflow etc. Not other safety measures. So for example ASAN and safe mode are different things.

rickyadastra:
I sorted out the platform defaults for the stack protector: now both MacOS and iOS have `ssp` turned on, and OpenBSD has `sspstrong` by default. Other platforms have it disabled. I also updated MacOS tests with the new IR attributes. 

There are two specific edge cases I'd like to point out: 
- `stack-protector-buffer-size` is currently emitted only when ssp is enabled; this is different from Clang where it's always emitted regardless of stack protector level
- naked functions with ssp enabled might generate buggy canary checks by accessing an unallocated stack frame; Clang permits it (e.g. a naked function targeting MacOS outputs both the `ssp` and `naked` IR attributes) 

Should we follow Clang's implementation in both cases?

lerno:
The clang behaviour might be a bug. It could be worth checking the diff between GCC and Clang for the asm. So it's like that by default on Clang or? If it's permitted, but not default, that seems ok.

rickyadastra:
I did a test with `-fstack-protector-all`, and GCC was inconsistent: on x86-64 it preferred the naked behavior over producing any stack protector checks, while on AArch64 it completely ignored the naked attribute and generated a full prologue/epilogue with the canary checks. I re-tested Clang as well and, while the IR has both the attributes, the generated asm always contains the naked function we would expect with no stack canary checks. 

I think it's fine to keep both attributes if Clangs does it as well, but skipping ssp emission in codegen if a function is marked as naked might be a more robust solution.

lerno:
Ok, let's do it like this:
1. `@naked` makes ssp always turn off *except* for when it's explicitly enabled on the function.
2. Apply the same to sanitizers, so that sanitizers are also only added when explicitly enabled on the function.
3. The overall ssp flag would default to whatever Clang defaults to for the os. But a global flag to enable it would, as I said in (1) not affect `@naked` unless it's explicitly added as an attribute alongside `@naked`

I think this would be the least amount of footguns. Does it sound reasonable?

PR #3438, opened 2026-08-14
Update win-subsystem accepted values in project properties

rickyadastra:
This PR updates the `win-subsystem` values listed in project properties to match the actual strings accepted by the parser, also aligning them to the build option output.


lerno:
Thank you.

Issue #3383, opened 2026-07-17
[Bug] Incorrect highlighting on unreachable code warning


BWindey:
When getting a warning about unreachable code, the compiler tries to highlight which code will never execute. The highlighting logic however seems to be incorrect in this case, demonstrated by this piece of code:

```cpp
fn void main() {
	return;
	dstring::temp(
		"this is a very long long string that I need to type to demonstrate "
		"this bug in highlighting the part of the code where the error occured "
		"so that it hopefully can be fixed"
	);
}
```

```
$ c3c compile bug.c3
1: fn void main() {
2:      return;
3:      dstring::temp(
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Warning: This code will never execute.

1: fn void main() {
2:      return;
              ^
Note: No code will execute after this statement.
```

It seems like it already has concatenated the string and placed it on the same line as `dstring::temp(` when calculating how long the `^^^` should be, this is not desirable tough, it should stick to the actual line of code as written.

BWindey:
This same highlighting error also happens when a contract is not met:

```cpp
<* @require x[1] == '-' *>
fn void foo(String x) {}

fn void main() {
	foo(
		"this is a very long long string that I need to type to demonstrate "
		"this bug in highlighting the part of the code where the error occured "
		"so that it hopefully can be fixed"
	);
}
```

```
 5: fn void foo(String x) {}
 6: 
 7: fn void main() {
 8:     foo(
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: @require "x[1] == '-'" violated.
```

[...]

PR #3440, opened 2026-08-14
fix: #3383 incorrect highlighting on error reports

Bqrry4:
Implements the format suggested in the issue.
Can be tested with the following:
```c3
<* @require x > 1000000 *>
fn void baz(int x) {}


fn void main() {
	1 + "basic one";
	// Simple line that is elided with '|' -----------------------------------------------------------------------------------------------;
	/* Line that is too long: */ 0 + "I can't wait to figure out the correct place to call free() for every byte of memory I allocate today!";
	/* Line that that has an error far to the right with a \t---------------------------------------------------------------------------	*/baz( (1293 + 392) - (384 * 389)  + (0132 * (9314 + 91 - 1321)) );
	/* Line that that has an error far to the right and is too long  --------------------------------------------------------------------- */baz(/*--------------------------------------------------------------------------------*/ (1293 + 392) - (384 * 389)  + (0132 * (9314 + 91 - 1321)) );

	// Multiline
	baz(
		(1293 + 392) - (384 * 389)
		+ (0132 * (9314 + 91 - 1321))
	);
}```

### CODE REVIEW

**Comment by lerno** [comment]
Looking good for the most part, but let's do this as well:

1. For the simple line that is elided with '|' add a ... at the end instead of `|`

2. In this case, it looks like the line is repeated:

```
9: 	/* Line that that has an error far to the right with a \t-------------------------------------------------------...
 9+137: ----------------	*/baz( (1293 + 392) - (384 * 389)  + (0132 * (9314 + 91 - 1321)) );
                        	  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(/Users/lerno/Projects/c3c/resources/testfragments/debugstuff.c3:9:137) Error: @require "x > 1000000" violated.
```

Can we do:
```
9: ... ------------	*/baz( (1293 + 392) - (384 * 389)  + (0132 * (9314 + 91 - 1321)) );
                        	  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ (col 137)
```
Instead perhaps?

**Comment by Bqrry4** [comment]
> Can we do:
> 
> ```
> 9: ... ------------	*/baz( (1293 + 392) - (384 * 389)  + (0132 * (9314 + 91 - 1321)) );
>                         	  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ (col 137)
> ```
> 
> Instead perhaps?

How should this look when the span is way too long and must be elided?
```
9: ... ------------	*/baz( (1293 + 392) - (384 * 389)  + (0132 * (9314 + 91 - 1321)) ); /*---------*/...
                        	  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
Some ideas for a different placements of the column indicator:
```
9+137: ... ------------	*/baz( (1293 + 392) - (384 * 389)  + (0132 * (9314 + 91 - 1321)) ); /*---------*/...
                        	  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
```
9: (col 137)... ------------	*/baz( (1293 + 392) - (384 * 389)  + (0132 * (9314 + 91 - 1321)) ); /*---------*/...
                        	  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
```
9: ... ------------	*/baz( (1293 + 392) - (384 * 389)  + (0132 * (9314 + 91 - 1321)) ); /*---------*/...
                        	  ^(col 137)~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Imho, the first one with the addition of `...` at the start and without repeating the line as it was before should look alright.

**Comment by lerno** [comment]
I kind of like all of them. Maybe we should ask on the discord to get more opinions

**Comment by Bqrry4** [comment]
Updated as requested, pls check.

**Comment by lerno** [comment]
We had a discussion on the discord, should that +137 be removed?

**Comment by lerno** [comment]
A tiny bug still:
```
  9: 	// Simple line that is elided with '|' ----------------------------------------------------------------------- ...
 10: 	/* Line that is too long: */ 0 + "I can't wait to figure out the correct place to call free() for every byte o ...
 11: 	/* Line that that has an error far to the right with a \t----------------------------------------------------- ...
 12: ... ---------------- */baz(/*--------------------------------------------------------------------------------*/ ...
                            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(/Users/lerno/Projects/c3c/resources/testfragments/debugstuff.c3:12:139) Error: @require "x > 1000000" violated.
```

When eliding from both sides, the line is 2 characters too short compared to the other lines.

**Comment by Bqrry4** [comment]
Not a bug in particular, but a behaviour from the way the current printing is performed. Both lines in fact are MAX_WIDTH bytes, but because of the row prefixes the tab expansion is of a different visual length. A solution could be assuming a tab width of 4 and render the line by doing manual expansion, or 8 as per default of ANSI terminals.
Tho, it might look weird when the tab is on the first column and it's shown as `tab length + 1 padded` spaces, but we could skip the padded space when the tab is present(?).
As a note the same goes for unicode characters, visually it will show the line as shorter.

**Comment by lerno** [comment]
Oooh ok. Should we try to fix that? For example we could print all tabs as 2, 3 or 4 characters everywhere.

**Comment by Bqrry4** [comment]
Should work now, well somewhat...
This can be used to get a look of it.
```c3
<* @require x > 1000000 *>
fn void baz(int x) {}

fn void main() {
	1 + "basic ââ";
	
	/* tab n â */ 	0 + "Î 	can't wait to figure out the correct place to call free() for every byte of memory I allocate today!";
	/* ă It was popularised thanks to these sheets and more recently with desktop publishing software like Aldus PageMaker and Microsoft Word including versions of Lorem Ipsumââ. */baz(	(1293 + 392) - (384 * 389) + (0132  * (9314 + 91 - 1321)) );
	/* ne that that has an error far to the right and is too long  --------------------------------------------------------------------- ââ*/baz(/*--------------------------------------------------------------------------------*/ 	(1293 + 392) - (384 * 389)  + (0132 * (9314 + 91 - 1321)) );

	// Multiline
	baz/*â*/(
	(1293 + 392) - (384 * 389)
		+ (0132 * (9314 + 91 - 1321))
	);
}
```
Currently, the width of a unicode char is assumed to be of 1, tho there are chars with width of 2. There is `wcwidth` in glibc, but if we want to support it for multiplatform i suppose we must have a lookup table of some sort for utf-8?

Also as a note, I avoided using the column from the provided location as it doesn't reflect a proper column in a visual sense. An editor and the diagnostic will show different columns for the same source, because the lexer just computes it assuming ascii. I am aware that c3 doesn't allow any unicode in the code, but it can appear in comments/string literals, so if there's the need for the `(file:row:column)` to match what is displayed in a text editor. this should be addressed, but i think in a separate pr.

Bqrry4:
To add to the side note, the current column affects the lsp as well.
Example:
```c3
fn void bar(String x, int y) {}
fn void main() {
	int? a = 2;
	bar("âppl€s", a!);
}
```
It should point the rethrow error to `a` but it shows it at `;`.

lerno:
Thanks a lot for this.

PR #3445, opened 2026-08-16
Add extern declaration open() function from unistd.h and added O_flags

qwuser234:
In std::os::posix::files added extern declaration posix function open(); and added fcntl.h flags for Linux(glibc). 

If possible, future OS flags will be auto-generated via a script. Execution timing is TBD: either at library authoring time or dynamically on each compile with std::os::posix::files


lerno:
You can probably get quite far by asking LLMs for this kind of information, even though they're unreliable (best bet is to ask them to explicitly get the information from headers and DON'T do any guessing)

lerno:
I am waiting for the additional constants to be added then. And don't forget to add a note in the releasenotes.

lerno:
What is happening with this one?

lerno:
It also appears that it doesn't compile.

lib/std/os/posix/files.c3
```diff
@@ -54,4 +55,43 @@ macro String readdir_cname() @local @const
...
+const O_TRUNC     = 0o01000;
+const O_APPEND    = 0o02000;
+
+const O_NONBLOCK  = 0o04000;
+const O_NDELAY    = O_NONBLOCK;
+
+const O_SYNC      = 0o4010000;
+const O_ASYNC     = 0o0020000;
+const O_RSYNC     = O_SYNC;
+const O_FSYNC     = O_SYNC
```
Darvisim, line 85: @lerno this line was missing a semicolon hence why it failed to compile.
```suggestion
const O_FSYNC     = O_SYNC;
```

lerno: Yes, but so abandoned and not compiling doesn't evoke confidence.

lerno:
Abandoned PR, closing.

Issue #1297, opened 2024-08-03
Add support for Xtensa embedded llvm target


xfcasio:
request to add support for the Xtensa llvm embedded target (for esp development boards)

lerno:
This is currently blocked due to LLVM support of Xtensa is not completely merged yet. This might happen in LLVM 20

kassane:
> This is currently blocked due to LLVM support of Xtensa is not completely merged yet. This might happen in LLVM 20

**Q:** What is xtensa's relevance for most users?
**A:** esp32 (legacy)!

I have my doubts about that. Maybe LLVM 30+!!
Xtensa for LLVM is experimental (generic cpu-features only). Currently need espressif-fork for esp8266, esp32 & esp32-S[2|3].

upstream:
https://github.com/llvm/llvm-project/blob/master/llvm/lib/Target/Xtensa/Xtensa.td
espressif/llvm
https://github.com/espressif/llvm-project/blob/xtensa_release_18.1.2/llvm/lib/Target/Xtensa/Xtensa.td

## References
- https://github.com/espressif/llvm-project/issues/4
- https://github.com/ldc-developers/ldc/issues/4725 (similar question)

kassane:
My simple test (picsimlab/espressif-QEMU)

```bash
# static c3c compiler
c3c --version
C3 Compiler Version:       0.6.2 (prerelease)
Installed directory:       /home/kassane/.local/bin/
LLVM version:              18.1.2
LLVM default target:       x86_64-linux-musl
```
![image](https://github.com/user-attachments/assets/b7160ec9-fdb0-4e1a-aa9b-8c365f56d053)

lerno:
So what do you advice here?

kassane:
> So what do you advice here?

My proposal is to continue the development of `xtensa` support, focusing on esp32-devices (primary purpose for espressif users).

**Note:** New espressif devices, now replacing xtensa to risc-v32 (no SIMD/Vector) and xtensa (esp32/S[2|3] **legacy**) for higher performance.

Add `espressif/llvm` fork artifact in the CI (no impacting upstream artifacts). This will make upstream c3c accessible without forking (like, esp-rs).

lerno:
So like a special espressif c3c version?

kassane:
> So like a special espressif c3c version?

esp32 support only!
The other upstream targets will remain unchanged.

kassane:
> So like a special espressif c3c version?

esp32 support only!
[...]

PR #3446, opened 2026-08-16
feat: allow xtensa to compile without flags

Iskra-YT:
Enable Xtensa by default in C3C

Makes the existing `elf-xtensa` target available in standard C3C builds by removing the `XTENSA_ENABLE` requirement. Depends on c3lang/llvm-for-c3#5, which adds the Xtensa backend to the LLVM builds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,7 +195,6 @@ The compiler is currently verified to compile on Windows, MacOS, Linux, OpenBSD,
 *\* Inline asm is still a work in progress*<br>
 *\* OpenBSD 7.8 is the only tested version*<br>
 *\* OpenBSD has limited stacktrace, needs to be tested further*<br>
-*\* Xtensa support is enabled by compiling with `-DXTENSA_ENABLE`. The [espressif llvm fork](https://github.com/espressif/llvm-project) is recommended for best compatibility*
 
 More platforms will be supported in the future.
 </details>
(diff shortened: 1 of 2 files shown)


Iskra-YT:
Oh, I just noticed this can be related to #1297.

ManuLinares:
Have you done any testing? Does llvm build, have you tried to build LLVM?

I don't see the llvm Xtensa code getting any significant updates.

Iskra-YT:
I haven’t done a full build of LLVM for this PR yet. It depends on llvm-for-c3#5, which is still pending, so I haven’t been able to complete the full build/test validation yet. I’ll update the PR once that dependency is in place (but I will be offline for some weeks)

ManuLinares:
> I haven’t done a full build of LLVM for this PR yet. It depends on llvm-for-c3#5, which is still pending, so I haven’t been able to complete the full build/test validation yet. I’ll update the PR once that dependency is in place (but I will be offline for some weeks)

I know, could you though? clone llvm-for-c3, build and test this?

lerno:
@Iskra-YT this seems like a low effort PR at this point. What's the point if you didn't test it?

lerno:
Ok, I looked up the related issue. It doesn't seem like Xtensa is ready for prime time yet. I would suggest using the Espressif fork, and because of that it is more appropriate to have it as a flag.

PR #3450, opened 2026-08-18
Fix @str_pascalcase dropping digits (treated as word separators) #3287

### LINKED ISSUE #3287: `@str_*case` issues
1. Pascal and camel return the same values - camel case should _uncapitilise_ the first latter.
2. Pascal/camel case incorrectly lower case "inner" words, while snake case correctly inserts an underscore. e.g. `helloWorld` currently becomes `Helloworld` instead of `HelloWorld`.
3. Constant and snake case don't handle spaces.
4. Snake case incorrectly inserts an extra underscore if an uppercase letter is preceded by an underscore, e.g. `Hello_World` becomes `hello__world`.

```c3
const String[] NAMES = {
	"helloWorld",
	"helloworld",
	"Hello_World",
	"hello_world",
	"HELLO WORLD",
	"HELLO   WORLD",
};
$for var $i = 0; $i < NAMES.len; $i++:
	var $name = NAMES[$i];
	io::printfn("%s > p:%s c:%s s:%s cc:%s", $name,
		@str_pascalcase($name),
		@str_camelcase($name),
		@str_snakecase($name),
		@str_constantcase($name),
	);	
$endfor
```
Output:
```
helloWorld > p:Helloworld c:Helloworld s:hello_world cc:HELLO_WORLD
helloworld > p:Helloworld c:Helloworld s:helloworld cc:HELLOWORLD
Hello_World > p:HelloWorld c:HelloWorld s:hello__world cc:HELLO__WORLD
hello_world > p:HelloWorld c:HelloWorld s:hello_world cc:HELLO_WORLD
HELLO WORLD > p:HelloWorld c:HelloWorld s:hello world cc:HELLO WORLD
HELLO   WORLD > p:HelloWorld c:HelloWorld s:hello   world cc:HELLO   WORLD
```
 
---

This should probably be a separate issue, but just putting it here since it's related.
But a "human/nice" and "title" case could also be useful:
- **@str_humanise**: Strips all non-alphanumeric characters and replaces them with a space - `hello__world` becomes `hello world`
- **@str_titlecase**: Uppercase the first letter of each word - `hello world` becomes `Hello World`

youdie006:
### Summary
`@str_pascalcase` -- and therefore `@str_camelcase`, defined as `@str_uncapitalize($$str_pascalcase(...))` -- treated every non-alphabetic character as a word separator. Digits are non-alphabetic, so they were both treated as separators AND dropped from the output:

```
@str_snakecase("h2ello_world_2")  // "h2ello_world_2"  (digits kept)
@str_pascalcase("h2ello_world_2") // "HElloWorld"      (digits gone!)
```

Expected: `"H2elloWorld2"` -- digits preserved, matching how `$$str_snakecase` already passes digits through.

Fixes #3287 (the "numbers being separators" part; @lerno confirmed in-thread this is a bug).

### Root cause
In `src/compiler/sema_builtins.c` the `BUILTIN_STR_PASCALCASE` fold used `if (!isalpha(c)) { capitalize = true; continue; }`. A digit fails `isalpha`, so it flipped the word-boundary flag and was skipped instead of copied.

### Fix
- Use `isalnum` (not `isalpha`) for the separator test, so digits are no longer separators.
- Add a branch that copies a digit straight through *without* consuming the pending capitalization, so `"foo_2bar"` -> `"Foo2Bar"` and `"h2ello_world_2"` -> `"H2elloWorld2"`.

This mirrors the existing digit handling in `$$str_snakecase`. Output length can only shrink or stay equal, so the existing `malloc_string(len + 1)` buffer stays safe. ASCII-only is intentional ("intended for transforming ascii names").

### Testing
Added a case to `test/test_suite/builtins/string_builtins.c3t` (`"h2ello_world_2"`), asserting the folded IR: snakecase stays `"h2ello_world_2"` and pascalcase becomes `"H2elloWorld2"`. Verified red/green against a locally built c3c (LLVM 22): with the fix reverted the test fails (`test.ll did not contain @.str.9 ... "H2elloWorld2"`; buggy output is `"HElloWorld"`); with the fix, the builtins suite (37) and full test suite (1576) are green. Added a releasenotes.md entry.

---
This change was prepared with AI assistance and reviewed by me before submission.


lerno:
Please don't use AI to generate PRs, as per contribution guidelines. The verboseness of AI descriptions is a detriment to understanding rather than adding to it.

lerno:
Ouch I just discovered this one was a PR for *dev*

lerno:
I'll be moving this into master

lerno:
Merged into master separately.

PR #3452, opened 2026-08-18
Add `std::collections::Tree` implementation

archishmannag:
Added an implementation of a balanced Binary Search Tree structure as a complement of the HashSet/OrderedSet collection.

The API is kept mostly similar to HashSet, so they are interchangeable, provided the Type implements the comparison operators (is ordered). Although, the API of the TreeIterator is slightly different from that of HashSetIterator, in that it has `get` as a separate method than `next`.


lerno:
It should probably be called TreeSet then? Can you also add an entry in releasenotes?

archishmannag:
Yeah, I'll add an entry in the release notes. We can rename it to TreeSet, since that seems more fitting.

In case we rename the collection to TreeSet, what should be the name of the module? `tree` or `treeset`?

lerno:
It can be placed in the same module as `HashSet`, so `std::collections::set`

archishmannag:
Done. Added it under `std::collections::set` and updated the releasenotes.md

lib/std/collections/treeset.c3
```diff
@@ -0,0 +1,678 @@
+<*
+ @require Value::is_ordered : `Value must be ordered`
+*>
+module std::collections::set <Value>;
+import std::io @norecurse;
+
+const Allocator TREE_HEAP_ALLOCATOR = (Allocator)&dummy;
```
lerno, line 7: This is never used. Expected something like ONSTACK_TREE which would use it. Except it's not necessary, because you can use the one in HashSet

lerno:
Bugs found: 
1. inserting a single element, then removing it will cause a null pointer.
2. Using .next on an iterator with an empty set will cause a null pointer.

lib/std/collections/treeset.c3
archishmannag: Yeah, forgot to implement that 😅. Fixed it now.

lerno:
Fails because <Value> in HashSet and <Value> in TreeSet are unified. I changed name so it's fixed. But it still fails on the test I added. Please see if you can fix it.

lerno:
Thank you!

PR #3462, opened 2026-08-21
feat: add iOS to CI

Darvisim:
This PR gives the initial support for iOS.

The following changes were made:
- Two new targets have been added for iOS alongside `ios-aarch64`: `ios-aarch64-sim` and `ios-x64-sim`

- All 3 iOS targets have been added to the `manifest.json` in both `clib.c3l` and `clib2.c3l`.

- Linker setup in `linker_setup_darwin` of `linker.c`, target detection in `target.c` and support in `darwin_support.c` have been added for iOS.

- Modified `ci_tests.sh` to allow running ci tests for iOS.

- A workflow has been added for iOS which does the following: build host's c3c binary, boot the simulators, run the CI tests, test physical device binary.


lerno:
Looking good. Can you unify the duplicated code between mac_support and ios_support? And make sure it uses tab.

ManuLinares:
ios_tests.sh should be merged to ci_tests.sh to reduce the "drift" and duplication.

I don't think we need the full test suite for "iOS x64-sim".

What's the difference between simulator : true and false?

This isn't marked as draft but I see you're still making changes, what are you working on?

Darvisim:
> Looking good. Can you unify the duplicated code between mac_support and ios_support? And make sure it uses tab.

Like put those in one file: darwin_support.c?
Sure, I'll also unify the linker setups of both ios and macos into linker_setup_darwin

Darvisim:
> ios_tests.sh should be merged to ci_tests.sh to reduce the "drift" and duplication.
>
Part of me thought it would be good to have it in a separate script because of all these xcrun simctl spawns. Sorry about that; I’ll see what I can do.

> I don't think we need the full test suite for "iOS x64-sim".
>
Sure, I’ll remove it from the workflow.

> What's the difference between simulator : true and false?
>
Out of the 3 targets, 2 are simulators so that’s why there’s `simulator: true` and `simulator: false`. `simulator: true` for both simulator targets `ios-aarch64-sim` and `ios-x64-sim`, `simulator: false` for the physical device target `ios-aarch64`.

> This isn't marked as draft but I see you're still making changes, what are you working on?
>
I've been tinkering with `ios_tests.sh` to get as close as possible to `ci_tests.sh` that I didn't pay attention to marking this as draft. I’ll mark this as draft and start working on `ci_tests.sh` to include both `ios-aarch64` and `ios-aarch64-sim`.

lerno:
What is the progress on the draft?

Darvisim:
> What is the progress on the draft?
>
- `mac_support.c` and `ios_support.c` are merged into `darwin_support.c`.
- `linker_setup_macos` and `linker_setup_ios` are also merged into `linker_setup_darwin`.
- iOS now uses `ci_tests.sh` which means we no longer need a seperate script like `ios_tests.sh`.
- The workflow is simplified from build matrix to a unified workflow.

lerno:
Thank you!

PR #3469, opened 2026-08-23
fix(bigint): shift left for values > 32 bits

ricardobeat:
Fixes an issue where BigInt.shl silently returns 0 when shifting by 32 bits or more (e.g. 1 << 32 gave 0). It should allow the result to grow up to `MAX_LEN` instead, as other methods already do.


lerno:
What should we do about 8192 bit shifts?

ricardobeat:
> What should we do about 8192 bit shifts?

In the context I found this (the ECMAScript test262 suite) the cap is incidentally at the same 8192 bits, seems practical enough - RSA tops out at 4096 bits.

To have unlimited arbitrary precision, you'd need to make a wider change to `math::bigint` to introduce dynamic allocation; those being much narrower use cases (math research, crypto, etc), they would probably be better served by custom, more optimized libraries? Not particularly experienced in this area.

lerno:
Yeah, dynamic allocation is a different thing. I don't know if there is crypto requiring over 8192 bits right now? But I was thinking 8192 will be a little funny, corrupting the data. It will be zero but be quite right since it's not normalized?

ricardobeat:
I've added an `assert` so it will panic on overflow instead, seems in line with the other methods.

lerno:
If this is a contract violation it should be in the contract rather than an assert.

lerno:
So could you add it to the contract perhaps?

lerno:
Reviewing this again, the assert is the only thing possible here it seems.

lerno:
Thank you!

PR #3472, opened 2026-08-24
Fix stale tid in thread::current on POSIX

ManuLinares:
so thread::current() returns the valid kernel TID inside worker threads

```c3
module main;
import std;

fn int worker(void* arg)
{
	io::printfn("Worker Kernel TID: %d", thread::current().tid);
	return 0;
}

fn void main()
{
	Thread t;
	t.create(&worker, null)!!;
	t.join();
}
```


lerno:
Releasenotes 🙏

ManuLinares:
> Releasenotes 🙏

🤟️

lerno:
Thank you!

PR #3478, opened 2026-08-25
Fix flat_kind using wrong function for recursion

vyachkoslav:
flat_kind used flat_type for recursion and produced errors when was used with types of kind TYPEDEF or CONSTDEF, because of different return type


lerno:
Thank you!

Issue #3282, opened 2026-05-29
csv.c3 should handle "foo,bar"


lerno:
Currently `csv.c3` is very barebones.

Some things it could do:
1. Handle quote character, selecting quote character
2. Configure behaviour on double quote, i.e.does `""` escape a quote?
3. Disable quote escaping

Maybe something more as well.

lerno:
Still, the csv parser should be implemented straightforward and easy. I've seen horrors.

sandervdbosch:
Is there a specific API that we would like for all the different csv, json, xml, etc parsers?
I reckon it would be nice for all the filetypes to have the same return the same data structure or at lease a structure that has the same methods to use the parsed data.

lerno:
Probably not right now. It's ok that they are different.

PR #3482, opened 2026-08-28
Improved CSV parser

sandervdbosch:
As discussed in issue #3282 this new version handles csv files according to RFC 4180.

I published this as a draft to get feedback on the implementation.

Things that are still TODO:
- [x] Add more contracts.
- [x] Ensure header contains no empty entries.
- [x] Recognise and don't include UFT-8 header bytes in first entry of the file.
- [ ] Correctly parse non UTF-8 encoded csv files.
- [X] Improve performance of the parser.
- [x] Add to releasenotes.


sandervdbosch:
Currently I do not know how to go about skipping the uft8 header bytes and working with codepages, so I am opening this for review.

lerno:
conv::detect_bom will help you detect type of UTF

lerno:
codepage.c3 has conversion utilities for different codepages.

sandervdbosch:
Considering I don't currently have an elegant way of including codepage into the parser I am leaving it here for now.
I added a module description to have an overview of what the module offers.

If possible I would like to know if there are any issues with my coding style, since coding is a hobby and I don't know of industry standards for code style.

I am also happy to report that the parser is only marginally slower then the previous implementation when parsing a 4GB file of scientific measurements according to my own testing.

lerno:
Thanks for your patience. I've finally had time to look at this more closely. I have a concern but it's not about formatting or coding style.

The first issue is API compatibility: this changes the public API in ways that would break existing code, which we're not allowed to do in a 0.8.x release.

There's also the ownership model: `read_row()` now returns an allocated `CsvRow*`, but `CsvRow.free()` doesn’t free that allocation, making leaks easy. I also encountered crashes with malformed and edge-case CSV input.

I worked on refactoring the implementation, but I think it would benefit from stepping back and starting with a smaller design that preserves the old API where possible. The implementation is now nearly five times the size of the original, and I believe the added functionality could be achieved more simply.

There is useful work here, particularly in exploring the required behavior and adding tests, but I don't think the current design is ready to merge.

sandervdbosch:
Thanks for the time investment. 

I feel like API compatibility could be easily achieved by skipping all the specific header handling. The new code reads and parses the header during initialization, which could be seen more as quality of life rather then something the parser should do. Instead we can leave the handling of the header to the user while possibly providing functions to help check the header for correctness.

The ownership model of `read_row()` is something I did to reduce the `CsvReader` struct size that has the header row embedded as discussed before. I now feel like that was a mistake for the ownership model as well as for API compatibility.

The `find_row()` and `parse_row_entries()` are specialized versions of `io::readline()` and `String.split()` respectively. This was necessary to ignore delimiters and newlines within String entries. This inflates the codebase of this module substantially compared to the old situation.

For now I would not mind refactoring the code to be API compatible, although I would not know to achieve the `find_row()` and `parse_row_entries()` functionality in a more simple way. One thing I can think of is making `find_row()` more general by having it take a `String newline` argument instead of checking CRLF, LF and CR line endings at the same time.

I would like to know what course of action you would recommend or prefer for the standard library for now.

PR #3484, opened 2026-08-28
feat: `c3 init` now supports hyphen in project name

### LINKED ISSUE #3481: c3c init: accept hyphenated project names
```bash
$ c3c init my-project
'my-project' is not a valid project name.
```

Currently we cannot have hyphen in project name. 
I believe it's because project name is used as module name.

But it is a very common practice to have `-` in folder names.

May be `c3 init my-project` can create a folder with `my-project` and generate the module name as `my_project` (replacing hypen with underscore).

neerajnangireddy:
closes #3481 .
- Added support for hyphen in project name

Only `check_name` function needed to be updated to treat `-` as a valid char in project name.
The `module_name` function already sanitizes the project name before generating the `main.c3` template.


lerno:
I've thought about it and ok, we'll do it.

Geo25rey:
@neerajnangireddy thank you for fixing this. I just ran into this problem as well

PR #3486, opened 2026-08-29
feat: Add x86 control registers to the inline asm register list

ssimb:
We add x86 control registers to the register list so we can use them with inline asm

`CR0` is 32bit on both x86 and x86_64, that's why I registered it seperately.

### DIFF
--- a/src/compiler/asm_target.c
+++ b/src/compiler/asm_target.c
@@ -880,6 +880,8 @@ static void init_asm_x86(PlatformTarget* target)
 
 	target->clobber_name_list = X86ClobberNames;
 	target->extra_clobbers = "~{flags},~{dirflag},~{fspr}";
+
+
 	if (target->arch == ARCH_TYPE_X86)
 	{
 		reg_register_list(target, x86_long_regs, 8, ASM_REG_INT, ARG_BITS_32, X86_RAX);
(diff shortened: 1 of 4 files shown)


lerno:
Please add a test and update the releasenotes too

src/compiler/asm_target.c
```diff
@@ -898,7 +901,10 @@ static void init_asm_x86(PlatformTarget* target)
 		reg_register_list(target, x86_xmm_regs, 16, ASM_REF_FVEC, ARG_BITS_128, X86_XMM0);
 		reg_register_list(target, x86_ymm_regs, 16, ASM_REF_FVEC, ARG_BITS_256, X86_XMM0);
 		reg_register_list(target, x86_zmm_regs, 16, ASM_REF_FVEC, ARG_BITS_512, X86_XMM0);
+		reg_register_list(target, x86_control_regs, 3, ASM_REG_INT, ARG_BITS_64, X86_CR2);
 	}
+
+	reg_register(target, x86_control_regs[3], ASM_REG_INT, ARG_BITS_32, X86_CR0);
```
lerno: Is CR0 32 bit on 64 bit processors?

ssimb: Hi, yeah I double checked and it turns out CR0 is indeed 64-bits, just that the top 32-bits are unusable and reserved. I'll fix this

ssimb:
Quick question, where do I put the test? I saw in the README it says that they should go in `test/unit`, however I see that the bulk of the asm tests seem to be in `test/test_suite`. Furthermore the CR* registers result in a General Protection fault if used in usermode, so it shouldn't be run.

lerno:
Use test/test_suite. Look at how the other asm tests are done. They basically will generate asm that LLVM then validates. And that's enough to show that they work.

lerno:
Thank you!

PR #3487, opened 2026-08-30
Cleans up several oversized functions across `std::hash` (SHA-256, Streebog, BLAKE3) and `std::compression` (ZIP, Deflate):

ManuLinares:
- Replaces repetitive macro expansions in Streebog and BLAKE3 with shared `@noinline` functions and pointer passing.
- Unifies aligned/unaligned SHA-256 transform paths to avoid duplicate 64-round bodies.
- Breaks up monolithic ZIP (`open`, `close`, `open_reader`) and Deflate (`emit_block`, `pkg_merge`) functions into private helpers.

Reduces frontend IR instruction counts without affecting runtime performance or behavior.

instruction_reduction_results

### DIFF
--- a/lib/std/compression/deflate.c3
+++ b/lib/std/compression/deflate.c3
@@ -518,43 +518,8 @@ fn void? Deflater.scan_matches(&self, bool final)
 	self.match_pos = (uint)pos;
 }
 
-fn void? Deflater.emit_block(&self, bool final)=> @pool()
+fn void _rle_encode_tree_lengths(char[] all_lens, List{Token}* tree_tokens, uint[] tree_freqs) @local
 {
-	if (self.token_count == 0 && !final) return;
-
-	self.lit_freqs[256] = 1;
-	self.tokens[self.token_count++] = { 256, 0 };
-
-	// Build dynamic Huffman trees from frequency tables
-	Huffman lit_huff, dist_huff;
-	lit_huff.build_from_freqs(self.lit_freqs[..], 15);
-
-	uint d_sum = 0;
-	foreach (f : self.dist_freqs) d_sum += f;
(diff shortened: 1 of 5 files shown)


ManuLinares:
### Hyperfine shows 5.9% faster builld times for the stdlib

```bash
hyperfine \
  --warmup 5 \
  --runs 100 \
  --prepare 'git checkout bloat >/dev/null 2>&1' \
  './build/c3c compile hello.c3 -O1 --strip-unused=no -g0' \
  --prepare 'git checkout master_upstream >/dev/null 2>&1' \
  './build/c3c compile hello.c3 -O1 --strip-unused=no -g0'
```

lib/std/compression/deflate.c3
```diff
@@ -598,6 +563,104 @@ fn void? Deflater.emit_block(&self, bool final)=> @pool()
 			i++;
 		}
 	}
+}
+
+fn void _encode_match_len(uint b_len, uint* code, int* extra_bits, uint* extra) @local
```
Darvisim: Is there any reason for the underscore before the function name? Just curious

ManuLinares: It's just a convention for locals/private functions. Mostly helpers or intermediaries.

lerno:
I removed some `@noinline`, because I feel it's better to allow the compiler to have control over that. Also, I've tried to move the local part to the end of the files for ease of reading.

ManuLinares:
> I removed some `@noinline`, because I feel it's better to allow the compiler to have control over that. 

👍️

Synthetic benchs consistently regressed O2/O3 by about 3–4%, while O1 stayed flat.
But I think in a real use of blake3, this result is meaningless.

PR #3491, opened 2026-08-31
compiler: Optimize slice equality, fix $$reverse and $$mod

ManuLinares:
Optimize slice equality for flat types by lowering comparisons to
memcmp to avoid emitting scalar loops.

Fix heap buffer corruption when reversing vectors larger than 128
elements due to an undersized allocation.

Fix $$mod on unsigned integers emitting signed division instead of
unsigned remainder.


lerno:
Great stuff. Releasenotes? 🙏

ManuLinares:
Added release notes

lerno:
Excellent!

PR #3493, opened 2026-09-01
Use tree merge for single-module LLVM linking

ManuLinares:
Instead of merging every module one-by-one into the first one (which gets slower and slower as it grows), merge them in pairs across multiple passes.

+20% on single-module -O0

### DIFF
--- a/scripts/tools/ci_tests.sh
+++ b/scripts/tools/ci_tests.sh
@@ -102,6 +102,7 @@ run_examples() {
     run_c3c compile examples/hash.c3
     run_c3c compile examples/http_server.c3
     run_c3c compile-only examples/levenshtein.c3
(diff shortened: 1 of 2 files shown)


lerno:
Add a test for --single-module=yes?

ManuLinares:
like here https://github.com/c3lang/c3c/blob/master/scripts/tools/ci_tests.sh#L104 we could add one.

```bash
    run_c3c compile examples/http_server.c3
    run_c3c compile-only examples/levenshtein.c3
    run_c3c compile-only --single-module=yes examples/levenshtein.c3  #this one
    run_c3c compile examples/load_world.c3
```

lerno:
Yes

lerno:
Thank you!

PR #3494, opened 2026-09-01
fix: drop support of unused Win32_WIN32_FIND_DATAW fields

velikoss:
The `dwFileType`, `dwCreatorType` and `wFinderFlags` fields only existed behind `#ifdef _MAC`, and current Windows SDKs do not declare them at all. Including them made this struct 604 bytes: harmless when only a pointer is passed to `FindFirstFileW`, but wrong for anything that embeds it or measures it.

10.0.26100.0/um/minwinbase.h:

```c
typedef struct _WIN32_FIND_DATAW {
    DWORD dwFileAttributes;
    FILETIME ftCreationTime;
    FILETIME ftLastAccessTime;
    FILETIME ftLastWriteTime;
    DWORD nFileSizeHigh;
    DWORD nFileSizeLow;
    DWORD dwReserved0;
    DWORD dwReserved1;
    _Field_z_ WCHAR  cFileName[ MAX_PATH ];
    _Field_z_ WCHAR  cAlternateFileName[ 14 ];
} WIN32_FIND_DATAW, *PWIN32_FIND_DATAW, *LPWIN32_FIND_DATAW;
```

### DIFF
--- a/lib/std/os/win32/files.c3
+++ b/lib/std/os/win32/files.c3
@@ -29,14 +29,8 @@ struct Win32_WIN32_FIND_DATAW
 	Win32_DWORD         nFileSizeLow;
 	Win32_DWORD         dwReserved0;
 	Win32_DWORD         dwReserved1;
-	Win32_WCHAR[260]    cFileName;
+	Win32_WCHAR[MAX_PATH] cFileName;
 	Win32_WCHAR[14]     cAlternateFileName;
-	<* Obsolete. Do not use. *>
-	Win32_DWORD         dwFileType;
-	<* Obsolete. Do not use *>
-	Win32_DWORD         dwCreatorType;
-	<* Obsolete. Do not use *>
-	Win32_WORD          wFinderFlags;
 }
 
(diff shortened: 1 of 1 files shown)


lerno:
I use the definition here: https://learn.microsoft.com/en-us/windows/win32/api/minwinbase/ns-minwinbase-win32_find_dataw

lerno:
Also, can you double check that MAX_PATH there isn't mixed up with some other MAX_PATH?

velikoss:
> I use the definition here: https://learn.microsoft.com/en-us/windows/win32/api/minwinbase/ns-minwinbase-win32_find_dataw

My ABI tests with libuv showed that lib expects struct without these types. Plus they are obsolete from 90s I think

> Also, can you double check that MAX_PATH there isn't mixed up with some other MAX_PATH?

Yes, I will look into that

velikoss:
Yes, thats the right MAX_PATH. However new versions support path greater than 260. Probably const there is to change later for some env/sys variable, i guess?
https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry

lerno:
Looks good then!

PR #3504, opened 2026-09-06
c_backend: Implement C backend

ManuLinares:
Add the `--cflags` option to forward flags to the underlying C compiler.

It's a biggie, it passed all the unit tests:

```
c3c compile-test -O1 --backend=c --cc clang test/unit/
c3c compile-test -O1 --backend=c --cc gcc test/unit/
```

haven't tested on msvc yet so it most probably break there, but I'll put it in draft.


ManuLinares:
mmm, I don't know if I should "namespace" the externs, so they don't conflict with each other on same name but different args. `testproject`

lerno:
Pure vaarg should be () not (...)
`error: ISO C requires a named parameter before '...' 2034 | typedef void (*__c3_fnptr_303303b30)(...);`

`static int? test = 32;` miscompiles

PR #3507, opened 2026-09-07
implement MSVC header fetching

### LINKED ISSUE #3401: `fetch-sdk windows` does not fetch C headers for cross-compilation
With the recent addition of #3355 C/C++ headers for cross compilation are present in `usr/include/`, this allows people to cross compile C/C++. However for Windows only MSVC libraries are present.

LowByteFox:
Fetches MSVC headers allowing C/C++ cross compilation, results in additional tree structure:
```
include
include/crt/*
include/sdk/ucrt/*
include/sdk/um/*
include/sdk/shared/*
```

Fixes #3401

image


LowByteFox:
Do not merge yet! I'm finding headers that need to keep their name case

LowByteFox:
Had to change this to draft because not all headers may be present yet, still looking into it.

LowByteFox:
As of now, the headers add additional ~50MB to msvc_sdk. I will change the PR to normal but will double check something isn't missing. Windows Aarch64 isn't there yet.

ManuLinares:
Amazing work, for my own documentation can you write a few examples using `clang`?
So maybe we could automate this for `c-sources` when targeting `windows-x64`

LowByteFox:
I'm currently here and there working on [korsbox](https://github.com/LowByteFox/korsbox). I have talked about it with @lerno and it may later become part of C3 itself. It'll need a lot more work to be done, but if you want example usage with `clang`, you can find it [here](https://github.com/LowByteFox/korsbox/blob/main/src/targets/windows-x64.c).

It's not perfect and I'll have to fine tune it further, but hopefully before C3 0.9. I'll see how it goes.

lerno:
What's the status right now?

LowByteFox:
Header-wise, should be fine. Library remaping-wise, unsure. Additional rempaings can be added later if needed.

lerno:
It would be nice if this was tested in the CI

ManuLinares:
> It would be nice if this was tested in the CI

Needs to be carefully considered. Fetching too much from Microsoft in Github (although the same company) would be bad (last year we did this). I would limit with an invalidated daily cache for example, maximum once a day.

We need to reduce to a few OSes, for example one Linux and one MacOS.
And besides the fetching + C3 cross-compilation, it would be nice to have a few C plus C3 examples we could manually cross-compile with clang (simple test). If not, I would suggest a dedicated runner.

lerno:
True. Ok, let's grab this and make any fixes needed afterwards if problems show up. Thanks for the work @LowByteFox

PR #3508, opened 2026-09-08
Add FlatMap and FlatSet data structures to stdlib

nyr24:
As alternative to default hashmap and hashset implementations.

### DIFF
--- a/lib/std/collections/flatmap.c3
+++ b/lib/std/collections/flatmap.c3
@@ -300,15 +300,15 @@ fn void? FlatMap.remove(&self, Key key) @maydiscard
 <*
  Iterates over each key-value pair in the map.
 
- @require self.is_initialized() : "Must be initialized set"
+ @require self.is_initialized() : "Must be initialized map"
 *>
 macro void FlatMap.@each_pair(&self; @body(Key key, Value value))
 {
 	if (@unlikely(self.is_empty())) return;
 
 	HashType[] hashes = hashes_array(self);
-	Key[] keys        = keys_array(self.data, self.capacity);
-	Value[] values    = values_array(self.data, self.capacity);
+	Key[] keys        = keys_array(self);
+	Value[] values    = values_array(self);
 
 	foreach (idx, hash : hashes)
 	{
@@ -322,14 +322,14 @@ macro void FlatMap.@each_pair(&self; @body(Key key, Value value))
 <*
  Iterates over each key in the map.
 
- @require self.is_initialized() : "Must be initialized set"
+ @require self.is_initialized() : "Must be initialized map"
 *>
 macro void FlatMap.@each_key(&self; @body(Key key))
 {
 	if (@unlikely(self.is_empty())) return;
 
 	HashType[] hashes = hashes_array(self);
-	Key[] keys        = keys_array(self.data, self.capacity);
+	Key[] keys        = keys_array(self);
 
 	foreach (idx, hash : hashes)
 	{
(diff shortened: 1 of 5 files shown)


lib/std/collections/flatset.c3
```diff
@@ -0,0 +1,534 @@
...
+fn FlatSet* FlatSet.init(
+	&self,
+	Allocator allocator,
+	sz capacity = DEFAULT_CAPACITY,
+	float load_factor = DEFAULT_LOAD_FACTOR
+)
+{
+	sz mem_req = calc_mem_req(math::next_power_of_2(capacity));
+	self.data = alloc::malloc(allocator, mem_req);
+	self.capacity = capacity;
```
lerno, line 52: Unlike FlatMap, doesn't make capacity power of 2

lib/std/collections/flatset.c3
```diff
@@ -0,0 +1,534 @@
...
+<*
+ Returns the union of two sets (A | B)
+
+ @param [&in] other : "The other set to union with"
+ @param [&inout] allocator : "Allocator for the new set"
+ @return "A new set containing the union of both sets"
+*>
+fn FlatSet FlatSet.set_union(&self, Allocator allocator, FlatSet* other)
+{
+	sz new_capacity = math::next_power_of_2(self.count + other.count);
```
lerno: Unused

lib/std/collections/flatmap.c3
```diff
@@ -0,0 +1,518 @@
...
+	foreach (&hash : hashes)
+	{
+		*hash = HASH_EMPTY;
+	}
+}
+
+fn bool crossed_threshold(FlatMap* self) @inline @private
+{
+	// Capacity is a power of 2; avoid float division on every insert.
+	return self.count >= (sz)(self.capacity * self.load_factor);
```
lerno: You can precalculate the threshold so it doesn't need to be recalculated every pass.

lerno:
Add a benchmark comparison with hashmap and hashset as well.

nyr24:
@lerno 
made fixes for the FlatSet (add_inner previously could cause infinite loops on big amounts of inputs)
and added simple benchmarks

lerno:
Benchmark files share the same module name so can't compile together.

lerno:
Some possible bugs:
- FlatMap.set_inner might corrupts the map. It inserts at the first slot with data[idx] < HASH_FIRST_VALID, which matches tombstones as well as empty slots?
- Tombstones seem to never be reclaimed
- Check get_idx, it might be incorrect.
- HashMap applies rehash() before masking, this one doesn't –> worse hash
- FlatMap.has uses get, not get_ref and will need optimization to eliminate overhead.
- I think set might trigger a grow before discovering the key already exists?
- FlatSet.contains/remove and add_inner work differently, is this correct?

nyr24:
@lerno 
- set_inner was updated to match add_inner from FlatSet, to check for duplicated values; answering the question - yes, it will match tombstones as well, do you see a possibility for corruption in updated implementation?
- if loop founds HASH_DELETED, and there's no duplicates after in the probe sequence, then it will insert value there and update the hash, can you expand your position a little bit here?
- fixed
- I'm not sure why call to calc_hash() can't be considered sufficient here, is it mandatory to do rehash()?
- fixed
- although it can be seen as problematic, I think its a good thing that even if there's no insertion will happen, map still checks its threshold for matching load_factor, and it keeps implementation simple, so this remains as a question
- these was uniformed

lerno:
Rehash will ensure that values with entropy in their topmost bits don't all hash to the same thing. So yes. Required.

The contract on set is different too. Compare with hashmap and hashset in what it returns.

PR #3513, opened 2026-09-13
fix: Remove `NO_DEFAULT_PATH` from CMakeLists.txt.

Ashvith10:
Removing `NO_DEFAULT_PATH` resolves build failure for C3 when using `-DC3_LINK_DYNAMIC`.


Ashvith10:
Context for the PR: https://codeberg.org/guix/guix/pulls/11188

ManuLinares:
I can’t assign this to myself, but I’ll take a look when I can.

Ashvith10:
This seems to work inside of the Fedora container? Both with and without `-DC3_LINK_DYNAMIC=ON`?

However, for `-DC3_LINK_DYNAMIC=ON`, it requires the `-DCMAKE_EXE_LINKER_FLAGS="-lzstd"` flag.

Ashvith10:
I've also tried Debian container and it builds fine, but `-DC3_LINK_DYNAMIC=ON` can't be built, I'm assuming this is because Debian only ships static objects.

ManuLinares:
It allows cross-version pollution when multiple LLVM versions are installed, and breaks static `libcurl` resolution.

Maybe I can help you fixing the guix package, seems you got on the right track here https://codeberg.org/guix/guix/pulls/11188#issuecomment-22731031 but then why this one gnu/packages/patches/c3c-remove-bundled-miniz.patch ?

You are removing the c3c/dependencies/miniz/zlib_compat.c when you are doing `          (delete-file-recursively "dependencies") `

For other linux distros [here is a list of normal dependencies](https://github.com/c3lang/c3c#compiling-on-linux)

@Ashvith10 are you on any chance in the discord?

Ashvith10:
> It allows cross-version pollution when multiple LLVM versions are installed, and breaks static `libcurl` resolution.

I am not well-versed with CMake, but can't we handle this by specifically forcing CMake to look for a particular version? Like this:

```cmake
find_package(LLVM 21 REQUIRED CONFIG)
find_package(LLD 21 REQUIRED CONFIG)
```

> Maybe I can help you fixing the guix package, seems you got on the right track here https://codeberg.org/guix/guix/pulls/11188#issuecomment-22731031 but then why this one gnu/packages/patches/c3c-remove-bundled-miniz.patch ?

Hi, one of the packaging guidelines of Guix includes [extracting bundled dependencies](https://guix.gnu.org/manual/devel/en/html_node/Extracting-Bundled-Dependencies.html). As of now, I'm on C3's version `0.7.10`, which does not have this issue thankfully, but as you've rightfully pointed, this will soon be an issue on future versions.

Is it viable to use zlib as a drop-down replacement on Guix in place of miniz? I understand that it will need some patches, but if it's much  easier to do so, then we can use that.

> @Ashvith10 are you on any chance in the discord?

I'm not on Discord :sweat_smile:.

Ashvith10:
Perhaps, we can have a chain of if-else for both LLVM and LLD:

```cmake
find_package(LLVM 21 REQUIRED CONFIG)
if(LLVM_FOUND)
  # Log that LLVM was found
else ()
  if (C3_FETCH_LLVM)
    # Fetch LLVM from the network
  else()
    # Error for LLVM not found, stop CMake
  endif ()
endif()
```

ManuLinares:
@Ashvith10 apply this patch to your PR https://codeberg.org/guix/guix/pulls/11188

[0001-gnu-Add-c3c.patch](https://github.com/user-attachments/files/32265510/0001-gnu-Add-c3c.patch)

ManuLinares:
@Ashvith10 let me know, so we can start maybe adding issues for up-streaming and discussing some of the fixes.

Ashvith10:
@ManuLinares I tested the patch locally. It was failing from a minor regex-related issue, which would also delete the matching string at [line 449](https://github.com/c3lang/c3c/blob/d64dd83450f61a844f6be8addc3a9ac40e1b94b4/scripts/tools/ci_tests.sh#L449). It can be resolved by using initial and terminal anchor (`^` and `$`):

```scm
(("^run_parallel dynlib run_dynlib_tests\n$") "")
;; Skip slow test runner during check phase
(("^run_unit_tests\n$") "")
```

I was able to build the new patch without needing this:

```scm
(("message\\(STATUS \"Looking for shared lld libraries in .*\"\\)")
  (string-append
    "list(APPEND LLVM_LIBRARY_DIRS ${C3_LLD_DIR})\n"
    "message(STATUS \"Looking for shared lld "
    "libraries in ${LLVM_LIBRARY_DIRS}\")"))
```

As long as you also have a modified copy of lld from the fork which has the config flag `-DBUILD_SHARED_LIBS=ON`: (for this, you may want to rebase the current c3c branch to https://codeberg.org/guix/guix/pulls/10590)

```scm
(("(liblld[A-Za-z]+)\\.so" _ name)
(string-append name ".so " name ".a"))
```

I've also noticed that `#$(cc-for-target)` was replaced with `gcc`. I've restored them since it is related to the target architecture, and on an x86_64 system, it points to "gcc".

I was also able to trim a bunch of other stuff that added repetitive lines to `CMakeLists.txt`.

For some reason, this project doesn't seem to build without zlib? I thought that miniz alone was enough?

Also, I was not able to figure out the issue with static library working fine with `test-bin`, but failing without them.

```console
(let ((bin (string-append (getcwd) "/test-bin")))
  (mkdir-p bin)
  (symlink (which "gcc") (string-append bin "/cc"))
  (setenv "PATH" (string-append bin ":" (getenv "PATH"))))
````

(Removed the last issue; that was caused by me mixing `llvm-21` and `lld-21` with `clang-toolchain-22`)

**Comment by ManuLinares** [comment]
@Ashvith10

Complete working [c3.scm.txt](https://github.com/user-attachments/files/32347729/c3.scm.txt)


1. **Restored LLD search in `CMakeLists.txt`**:
   Upstream Guix `lld-21` only provides static archives (`.a`), not shared (`.so`), and places them in `C3_LLD_DIR` rather than `LLVM_LIBRARY_DIRS`. Don't assume PR 10590 is being merged. Without it, appending `C3_LLD_DIR` and allowing `.a` fallback is mandatory.
2. **Removed `clang-toolchain` and added `zlib`**:
   `clang-toolchain` pulled in Clang/LLVM 22 and caused toolchain conflicts with `llvm-21`/`lld-21`.
3. **Restored `test-bin` symlink (`gcc -> cc`) in the check phase**:
   You removed `test-bin`, which broke tests that invoke `cc` directly.

---

If you prefer, here is the diff:
```diff
diff --git a/gnu/packages/c3.scm b/gnu/packages/c3.scm
index 6c62686..474f529 100644
--- a/gnu/packages/c3.scm
+++ b/gnu/packages/c3.scm
@@ -56,6 +56,13 @@
             (("\"\\$\\{CMAKE_SOURCE_DIR\\}/dependencies/miniz/\"\\)") "")
             (("(target_link_libraries\\(c3c PRIVATE miniz) (xz_embedded)" _ prefix suffix)
              (string-append prefix "::miniz " suffix " z"))
+            (("message\\(STATUS \"Looking for shared lld libraries in .*\"\\)")
+             (string-append
+              "list(APPEND LLVM_LIBRARY_DIRS ${C3_LLD_DIR})\n"
+              "message(STATUS \"Looking for shared lld "
+              "libraries in ${LLVM_LIBRARY_DIRS}\")"))
+            (("(liblld[A-Za-z]+)\\.so" _ name)
+             (string-append name ".so " name ".a"))
             ((" NO_DEFAULT_PATH") ""))
           (substitute* "src/compiler/compiler.c"
             (("(return cc = \")cc(\";)" _ prefix suffix)
@@ -84,10 +91,16 @@
           (replace 'check
             (lambda* (#:key tests? #:allow-other-keys)
               (when tests?
+                (let ((bin (string-append (getcwd) "/test-bin")))
+                  (mkdir-p bin)
+                  (symlink (which "gcc") (string-append bin "/cc"))
+                  (setenv "PATH" (string-append bin ":" (getenv "PATH"))))
                 (setenv "C3C_CC" #$(cc-for-target))
+                (setenv "SHOW_SUCCESS_LOGS" "false")
                 (setenv "SKIP_NETWORK_TESTS" "1")
                 (invoke "../source/scripts/tools/ci_tests.sh" "c3c")))))))
-    (inputs (list clang-toolchain curl libffi miniz lld-21 llvm-21))
+    (inputs (list curl libffi miniz zlib lld-21 llvm-21))
+    (propagated-inputs (list gcc-toolchain))
     (synopsis "Compiler for the C3 language")
     (description
      "C3 is a programming language that builds on the syntax and semantics
```

Ashvith10:
I've updated the package again. Last time, I was missing:
```scm
(("(return cc = \")cc(\";)" _ prefix suffix)
  (string-append prefix #$(cc-for-target) suffix))
```

With this, there is no need for:
```scm
(let ((bin (string-append (getcwd) "/test-bin")))
  (mkdir-p bin)
  (symlink (which "gcc") (string-append bin "/cc"))
  (setenv "PATH" (string-append bin ":" (getenv "PATH"))))
```

Only dynamic test seems to be failing at this point.

By the way:
```scm
(propagated-inputs (list gcc-toolchain))
```

should not be a part of the package. It is already provided by `cmake-build-system` as an input. Since `gcc-toolchain` is an optional dependency and not needed by vanilla C3 project, it should be provided though the config (ie, manifest or package input).

Something like:
**manifest.scm**
```scm
(use-modules
  (gnu packages c3)
  (gnu packages commencement))
(packages->manifest
  (list c3c gcc-toolchain))
```

ManuLinares:
I'm sure you can resolve those specific things (I saw you were discussing them with user `Dariqq`.


I have only one question, can we close this PR if https://github.com/c3lang/c3c/pull/3524 works?

ManuLinares:
@Ashvith10 let me know, no rush of course. Anything else I can do to help, ping me with an email.

PR #3515, opened 2026-09-14
Boot the iOS simulator completely with `wait_for_boot: true`

Darvisim:
We allow the simulator to finish booting completely with the `futureware/simulator-action` actions' option `wait_for_boot` set to true.


ManuLinares:
Thanks for taking the time to look into CI performance.

Unfortunately this shows no runtime improvement: I don't think polling defaults domains reduce process spawn overhead and adds an unconditional 60-second delay, while resolving booted has no measurable difference compared to an explicit UDID.

Darvisim:
> Thanks for taking the time to look into CI performance.
> 
> Unfortunately this shows no runtime improvement: I don't think polling defaults domains reduce process spawn overhead and adds an unconditional 60-second delay, while resolving booted has no measurable difference compared to an explicit UDID.

I understand. I've removed the step as well as the explicit `UDID` and kept the `wait_for_simulator: true` option under the simulator action so the simulator is completely booted.

lerno:
Is this showing speedup now?

Darvisim:
> Is this showing speedup now?

Well this doesn’t speedup, it just forces the simulator to boot completely. The action has an option for settling down background daemons but that option will be available when v6 releases, so that’ll speed up but the complete boot won’t.

looking at it now, i think it’s better to hold off until v6, i’m also thinking of adding this option around then, but for now not needed so i’ll close this.

PR #3523, opened 2026-09-17
Adding missing linux MSG_* flags to use with send/recv.

hchac:
Values from:
https://github.com/torvalds/linux/blob/master/include/linux/socket.h#L318-L348


ManuLinares:
Since you are here, can you recheck and complete them? https://github.com/torvalds/linux/blob/master/include/linux/socket.h#L318

hchac:
Just to keep this PR up to date with what we discussed on the discord:  I avoided adding any of the ones with "internal" in their corresponding comment, because I don't really know if they're usable. So I just wanted to be somewhat conservative with the ones I included.

When I double checked to see what my linux distribution includes in `/usr/include/bits/socket.h` (only place I saw for these definitions when using grep), I didn't see any definitions for those that are marked internal in the kernel's github `socket.h`.

I'll include anything extra you think is worth adding though, just let me know.

ManuLinares:
Thanks for doing the double checking

lerno:
Thank you

PR #3526, opened 2026-09-18
added rdmsr, wrmsr, shlq and variants of shlq

000nico:
adds support for the following x86/x86-64 instructions to the backend:

- **`rdmsr`** and **`wrmsr`**: Configured with appropriate register clobbers (`RAX` and `RDX`) to handle msr and r/w correctly
- **`shlq`**: added 64-bit shift-left instruction support along with condition code flag masks (`cc_flag_mask`) for proper flag tracking.


lerno:
Please add test cases and a note in the releasenotes as well.

lerno:
You can look at what this PR did: https://github.com/c3lang/c3c/pull/3486

srkkov:
Note that SHL (and related SAL/SAR/SHR) do not accept arbitrary 8-bit register operands, there is only an encoding to use the CL 8-bit register.

PR #3528, opened 2026-09-20
std::gfx: add Wayland wlr-layer-shell support

ManuLinares:
Implement zwlr_layer_shell_v1 in std::gfx::wl. Support layer surface creation, anchoring, exclusive zones, margins, and keyboard interactivity. Add interactive example in resources/examples/std_gfx/layer_shell.c3.

https://wayland.app/protocols/wlr-layer-shell-unstable-v1


lib/std/gfx/private/gfx_macos.c3
```diff
@@ -111,7 +111,7 @@ fn GfxMonitorMode[] monitor_get_modes(GfxMonitor* monitor)
 		modes.push({
 			.size = { (int)mode.getWidth(), (int)mode.getHeight() },
 			.refresh_rate = (float)mode.getRefreshRate(),
-			.rgb = (char[<3>])8,
+			.rgb = { 8, 8, 8 },
```
lerno: This compiler bug should now be fixed

lib/std/gfx/gfx.c3
```diff
@@ -2804,3 +2806,111 @@ fn void unload_vulkan()
 {
 	_unload_vulkan_native() @inline;
 }
+
+enum GfxWlLayer : char @feat(WAYLAND)
```
lerno: This looks very out of place. The rest of the gfx.c3 code is about cross platform features. So it's completely misplaced here. std::os::linux or a `std::gfx::linux` but here.`gfx.c3` is prime estate for the common api shared across all platforms.

lerno:
So: 

1. move the Wayland stuff to os::linux or ::gfx::linux
2. The split into _allocate_window and _create_window just in order to do some configuration differently feels wrong. See if you can instead allow passing an additional argument to _create_window to customize, and the default _create_window just passing the default for the normal case.
3. Remove the { 8, 8, 8 } fixes, since that should be fixed now.

lib/std/gfx/private/gfx_macos.c3
ManuLinares: 🤯️ you don't know how many times I encountered this and took it as if it was "the correct behavior". These kind of bugs are scary.

lerno: Handling of simd -> array conversions have been *hard*

ManuLinares:
Moved layer-shell out of `gfx.c3` into `std::gfx::wl`. Layer-shell is a Wayland compositor protocol, not a Linux API.

It felt wrong calling `linux::set_layer_keyboard_interactivity` instead of `wl::set_layer_keyboard_interactivity`, etc

I've also added an unrelated example `resources/examples/std_gfx/transparent_windows.c3`

PR #3530, opened 2026-09-23
std::gfx::wasm Fixes, emscripten structs, window sizing and events

ManuLinares:
Hardcoded `#canvas` element references broke applications using custom canvas targets. Inline JavaScript evaluation for keyboard events introduced runtime overhead and garbage collection pressure. Canvas resizing, maximization, and touch tracking did not account for CSS element scaling, causing incorrect input coordinates and missed resize updates.

Add an example to demonstrate cross-platform 3D rendering using `std::gfx` on native platforms and Emscripten WebGL2. The example showcases sphere mesh generation, QOI texture decoding, solar positioning, camera orbital controls, raycast picking, and custom bitmap text rendering.


lib/std/gfx/private/gfx_wasm.c3
```diff
@@ -618,109 +659,222 @@ fn int gfxJsInnerWidth() => emscripten_run_script_int("window.innerWidth || 0;")
...
+	int screenX;
+	int screenY;
+	int clientX;
+	int clientY;
+	int pageX;
+	int pageY;
+	bool isChanged;
+	bool onTarget;
+	int targetX;
+	int targetY;
```
lerno, line 833: CanvasX/Y is missing?

ManuLinares: https://github.com/emscripten-core/emscripten/blob/b6f4364e625d07c497db8ecb9cb29b8f97f35cee/system/include/emscripten/html5.h#L349

I didn't found them on any generated Js struct, so they are truly gone

ManuLinares:
```
  int targetY;
  // canvasX and canvasY are deprecated - there no longer exists a Module['canvas'] object, so canvasX/Y are no longer reported (register a listener on canvas directly to get canvas coordinates, or translate manually)
  int canvasX;
  int canvasY;
```

